PR #52, opened 2016-04-20

obedmr:
proposed features:

  - New CIAO config sample file at etc/ciao/ciao.conf
  - In order to start ciao-launcher you need to have that sample file
    at /etc/ciao/
    - We used viper as config file parser, it supports JSON, TOML, YAML, HCL and
      Java properties config files

### DIFF
--- a/etc/ciao/ciao.json
+++ b/etc/ciao/ciao.json
@@ -1,0 +1,13 @@
+{
+    "launcher": {
+	"server": "localhost",
+	"cacert": "/etc/pki/ciao/CAcert-server-localhost.pem",
+	"cert": "/etc/pki/ciao/cert-client-localhost.pem",
+	"compute-net":  "",
+	"mgmt-net":  "",
+	"hard-reset": false,
+	"disk-limit": true,
+	"mem-limit": true,
+	"simulation": false
+    }
+}
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/golang/glog"
 
+	"github.com/01org/ciao/config"
 	"github.com/01org/ciao/payloads"
 	"github.com/01org/ciao/ssntp"
 )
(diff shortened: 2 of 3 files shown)


markdryan:
Hi @obedmr .  Thanks for the PR.  I've got a few questions.

1. Will you be making changes to the other components as well?  I presume so.
2. Do we really need to use viper?  My concern here is that will pull in at least 5 new dependencies.  I'm having a hard enough time vendoring our existing dependencies so I'm a little reluctant to pull in some more.  What does Viper really give us here?  As far as I can tell the only thing that it really adds is the ability to merge config files.  Do we really need this?

If not we could just do

```
defaults := &struct{
  Server string `json:"server"`
  Cacert string `json:"cacert"`
  Cert string `json:"cert"`
  Compute-net string `json:"compute-net"`
  MgmtNet string `json:"mgmt-net"`
  Hardreset bool  `json:"hard-reset"`
  DiskLimit bool  `json:" disk-limit"`
  MemLimit bool  `json:"mem-limit "`
  Simulation bool `json:"simulation"`
} { 
  server: "localhost",
  CACert: "/etc/pki/ciao/CAcert-server-localhost.pem",
  Cert:"/etc/pki/ciao/cert-client-localhost.pem",
  DiskLimit: true,
  MemLimit: true,
}
  
data,_ = ioutil.ReadFile(" /etc/ciao/config.json")
json.Unmarshall(data,   &defaults)
}

```

@sameo , @kaccardi , @tpepper  How do you guys feel about pulling in viper and it's dependencies?

markdryan:
Actually, I think
```
data,_ := ioutil.ReadFile("/usr/share/default/ciao/config.json")
json.Unmarshall(data,   &defaults)
data,_ = ioutil.ReadFile("/etc/ciao/config.json")
json.Unmarshall(data,   &defaults)
```

Would merge the config files, should we need to do this.  Note I haven't test this.

obedmr:
@markdryan , thanks for your feedback. Answering your questions:
**1. Will you be making changes to the other components as well?** yes, I'll do it.
**2. Why viper? ** I found, it can support those different types of configuration files and yes, the capability of merging files.
**3.  Regarding your last comment. ** I'm seeing simpler, I'm testing and if it works, I'll re-submit the patch to review.

markdryan:
Also, I wonder whether you could add some logs to main() that outputs final values of the parameters that can be configured via the configuration files.  Maybe you could put this in a separate function called by main after flag.Parse.  With all the different ways to configure launcher it might not be obvious which settings were used for an individual run.  You'd need to look in 3 places, ps, and the two conf files, to figure this out.

Note this launcher should already be doing this.  The fact it isn't is an oversight.

sameo:
@obedmr @markdryan I would prefer if we don't add further dependencies through viper. Viper features look nice but we probably don't need them right now.

etc/ciao/ciao.json
```diff
@@ -0,0 +1,13 @@
+{
+    "launcher": {
+	"server": "localhost",
+	"cacert": "/etc/pki/ciao/CAcert-server-localhost.pem",
+	"cert": "/etc/pki/ciao/cert-client-localhost.pem",
```
sameo: I suppose this is a sample, but we should still make it more relevant and actually match what ciao-cert generates:

"cacert": "/etc/pki/ciao/cert-Scheduler-ciao-ctl.intel.com.pem"
"cert": "/etc/pki/ciao/cert-CNAgent-localhost.pem"

obedmr: @sameo  yes, that would a sample.

obedmr:
@sameo @markdryan I'm almost done with the no-more-dependencies approach. I did some testing and then I'll send the patch. It's working really cool. Thanks @markdryan for your recommendation.

obedmr:
Closing for creating a new pull request with the removal of viper.

PR #56, opened 2016-04-21

obedmr:
proposed features:

    - This module would contain all modules default structs
    - We'll have the following sections:
      - launcher
      - scheduler
      - controller
      - cnci-agent
    defaults that were set in config module

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,83 @@
+/*
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+*/
+
+package config
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+type CIAOConfig struct {
+	UsedFile  string
+	Launcher  LauncherConfig
+}
+
+type LauncherConfig struct {
+	Server      string `json:"server"`
+	CACert      string `json:"cacert"`
+	Cert        string `json:"cert"`
+	ComputeNet  string `json:"compute-net"`
+	MgmtNet     string `json:"mgmt-net"`
+	HardReset   bool   `json:"hard-reset"`
+	DiskLimit   bool   `json:"disk-limit"`
+	MemLimit    bool   `json:"mem-limit"`
+	Simulation  bool   `json:"simulation"`
+}
+
+func (launcher *LauncherConfig) initLauncher(){
+	launcher.Server     = "localhost"
+	launcher.CACert     = "/etc/pki/ciao/CAcert-server-localhost.pem"
+	launcher.Cert       = "/etc/pki/ciao/cert-client-localhost.pem"
+	launcher.ComputeNet = ""
+	launcher.MgmtNet    = ""
+ 	launcher.HardReset  = false
+	launcher.DiskLimit  = true
+	launcher.MemLimit = true
+	launcher.Simulation = false
+}
+
+func loadConfigFile(path, filename string, ciaoConf CIAOConfig)  CIAOConfig {
+	filePath := fmt.Sprintf("%s/%s", path, filename)
+
+	if _, err := os.Stat(filePath); err == nil {
+		file,err := ioutil.ReadFile(filePath)
+		if err != nil {
+			log.Fatalf("Error on file:  %v \n %v", filePath, err)
+		} else {
+			json.Unmarshal(file, &ciaoConf)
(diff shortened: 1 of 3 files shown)


obedmr:
Closing due typo in paths. I put my workstation's paths

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	DiskLimit   bool   `json:"disk-limit"`
+	MemLimit    bool   `json:"mem-limit"`
+	Simulation  bool   `json:"simulation"`
+}
+
+func (launcher *LauncherConfig) initLauncher(){
+	launcher.Server     = "localhost"
+	launcher.CACert     = "/etc/pki/ciao/CAcert-server-localhost.pem"
+	launcher.Cert       = "/etc/pki/ciao/cert-client-localhost.pem"
+	launcher.ComputeNet = ""
```
markdryan: The zero value for strings and bools are "" and false so you don't need to initialize ComputeNet, MgmtNet if you don't want to.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	launcher.MgmtNet    = ""
+ 	launcher.HardReset  = false
```
markdryan: Hard reset should not be in the conf file.  This something you run exceptionally when you want to ensure that all VMs and networks on a CN have been cleaned up.  Systemd is never going to launch launcher with this value set.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	launcher.DiskLimit  = true
+	launcher.MemLimit = true
+	launcher.Simulation = false
+}
+
+func loadConfigFile(path, filename string, ciaoConf CIAOConfig)  CIAOConfig {
+	filePath := fmt.Sprintf("%s/%s", path, filename)
```
markdryan: use path.Join, e.g.,

filePath := path.Join(path, filename)

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if _, err := os.Stat(filePath); err == nil {
```
markdryan: Call to Stat not needed.  ReadFile will fail if the file does not exist.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	launcher.Cert       = "/etc/pki/ciao/cert-client-localhost.pem"
+	launcher.ComputeNet = ""
+	launcher.MgmtNet    = ""
+ 	launcher.HardReset  = false
+	launcher.DiskLimit  = true
+	launcher.MemLimit = true
+	launcher.Simulation = false
+}
+
+func loadConfigFile(path, filename string, ciaoConf CIAOConfig)  CIAOConfig {
```
markdryan: ciaoConf should be passed by pointer.  Otherwise you copy the entire structure when calling and returning from this function.  This is probably about 200 bytes in total.  If you pass by pointer you can get rid of the return value.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	filePath := fmt.Sprintf("%s/%s", path, filename)
+
+	if _, err := os.Stat(filePath); err == nil {
+		file,err := ioutil.ReadFile(filePath)
+		if err != nil {
+			log.Fatalf("Error on file:  %v \n %v", filePath, err)
```
markdryan: It's idiomatic go to return if there is an error rather than use nested if then else statements.  The use of tab for indenting blocks sort of forces you to do this.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+		} else {
+			json.Unmarshal(file, &ciaoConf)
```
markdryan: We should check the return value of Unmarshall.

errcheck can help you make sure you have checked all errors.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+			ciaoConf.UsedFile = filePath
+		}
+	} 
+	return ciaoConf
+}
+
+func InitConfig() CIAOConfig {
+	config := CIAOConfig{}
```
markdryan: You could just use a composite literal here to initialse config rather than a function.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	config.Launcher.initLauncher()
+	configPaths := [...]string{
+		"/home/onmunoz/dev/go/dev/viper/config1",
```
markdryan: These paths don't look correct.

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+			log.Fatalf("Error on file:  %v \n %v", filePath, err)
+		} else {
+			json.Unmarshal(file, &ciaoConf)
+			ciaoConf.UsedFile = filePath
+		}
+	} 
+	return ciaoConf
+}
+
+func InitConfig() CIAOConfig {
```
markdryan: return a pointer to CIAConfig

config/config.go
```diff
@@ -0,0 +1,83 @@
...
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+type CIAOConfig struct {
+	UsedFile  string
+	Launcher  LauncherConfig
```
markdryan: You could embed LauncherConfig.  This would result in slightly shorted code in main.go, e.g.,

```
flag.StringVar(&serverURL, "server",  defaults.Server, "URL of SSNTP server")  
```
instead of
```
flag.StringVar(&serverURL, "server", defaults.Launcher.Server, "URL of SSNTP server")  
```

PR #57, opened 2016-04-21

obedmr:
proposed features:

    - This module would contain all modules default structs
    - We'll have the following sections:
      - launcher
      - scheduler
      - controller
      - cnci-agent
    defaults that were set in config module

### DIFF
--- a/networking/cnci_agent/scripts/README.md
+++ b/networking/cnci_agent/scripts/README.md
@@ -7,38 +7,34 @@ Helper scripts to provision and test CNCI Images
 ## CNCI Image Provisioning ##
 
 The CNCI Image creation scripts helps you create a CNCI Image from
-a clear linux cloud image. Clear cloud images can be obtained from
+a clear linux cloud image. Clear cloud images for the CNCI can be obtained from
 
-https://download.clearlinux.org/image/
+https://download.clearlinux.org/demos/ciao/
 
 The scripts are used to provision the image with the CNCI Agent and
 the certificates it needs to connect to the ciao-scheduler.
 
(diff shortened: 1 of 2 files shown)


markdryan:
@obedmr I stupidly added my comments in #56 instead of this PR.  Most still apply apart from the comment on the paths.  Could you check #56?

Also, I don't think it makes sense to put config.go in a separate package as it contains launcher specific code.  I'd move that file into launcher and change names of the variables so that they start with a lower case letter.  Obviously you can't do this for the structure that contains the Unmarshalled json.

obedmr:
@markdryan , /me applying changes. Just a question. I may move the launcher structs to launcher code, but what if we keep the InitConfg and LoadConfigFile functions at config.go? Don't you think that it may be a repeated code if we also add into the other projects like ciao-controller, ciao-scheduler, etc?

I could just create the launcher struct for example, and pass it to the config.go for reading and loading the variables there. And then reuse the config.go for the rest of components. How is it?

markdryan:
@obedmr  Right now InitConfig and LoadConfigFile are launcher specific as they take a CiaoConfig parameter.  If you can make them generic, then I agree it makes sense to leave them in a separate package.  LauncherConfig would need to move to launcher but I guess InitConfig could accept an interface{} parameter to which launcher would pass a *LauncherConfig.  Do we need to return usedFile?

obedmr:
@markdryan ok, gotcha, let me make it more generic. Regarding the usedFile was just to have the record of what was the last taken config file. But It is not necessary certain at the end. Because user will only put those values he or she wants to change, the rest values would be taken from defaults.

PR #60, opened 2016-04-22

obedmr:
proposed features:

  - "config" package for handling the CIAO config files parsing
      - This package is generic for all CIAO services' confs
    defaults that were set on each service

### DIFF
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/golang/glog"
 
+	"github.com/01org/ciao/config"
 	"github.com/01org/ciao/payloads"
 	"github.com/01org/ciao/ssntp"
 )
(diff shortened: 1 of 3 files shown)


etc/ciao/ciao.json
```diff
@@ -0,0 +1,9 @@
+{
+    "server": "localhost3",
+    "cacert": "/etc/pki/ciao/CAcert-server-localhost.pem",
+    "cert": "/etc/pki/ciao/cert-client-localhost.pem",
+    "compute-net":  "",
+    "mgmt-net":  "",
+    "disk-limit": true,
+    "mem-limit": true
+}
```
sameo: How do you plan to support controller, launcher (NN and CN) and scheduler from the same config file ? Each block should be namespaced.

etc/ciao/ciao.json
```diff
@@ -0,0 +1,9 @@
+{
+    "server": "localhost3",
```
sameo: Please use a more realistic (Even though fake) URL. e.g. ciao-scheduler.example.com to let people understand that this should be a real URL or an IP. No one is ever going to run a real cluster on a single machine.

etc/ciao/ciao.json
```diff
@@ -0,0 +1,9 @@
...
+    "cacert": "/etc/pki/ciao/CAcert-server-localhost.pem",
```
sameo: Real ciao-cert generated CA certificates are going to look like: CAcert-ciao-scheduler.example.com.pem (To match my previous comment). Please use that.

etc/ciao/ciao.json
```diff
@@ -0,0 +1,9 @@
...
+    "cert": "/etc/pki/ciao/cert-client-localhost.pem",
```
sameo: A ciao-cert generated private key name will contain the embedded role. In that case this would be: cert-CNAgent-localhost.pem. Please use that name as well.

config/config.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+	}
+	err = json.Unmarshal(file, &ciaoConf)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func InitConfig(config interface{}) error {
```
sameo: Please document this exported function.

ciao-launcher/main.go
```diff
@@ -538,6 +582,12 @@ func main() {
 		log.Fatalf("Unable to initialise logs: %v", err)
 	}
 
+	if configError != nil {
+		log.Fatalf("Failure in Config files: %v", configError)
```
sameo: Not sure we want this one to be fatal. If the right information was provided through the command line, a wrongly formatted configuration file should only be a warning and not prevent us from starting launcher.

config/config.go
```diff
@@ -0,0 +1,58 @@
+/*
```
sameo: I'd expect to see a couple of unit tests for that new package. We want to easily verify through go test that a correctly formatted config file will work and a malformed one will fail as expected.

obedmr:
@sameo thanks for the feedback, /me applying changes.

obedmr:
Configuration support architecture changed to #66

PR #63, opened 2016-04-23
ciao-launcher: Networking: better macvtap defaults

mcastelino:
- Set up better macvtap defaults in ciao-launcher with mq and vhost support
- Update the iperf to use 8 tcp connections to leverage mq
- Update the CNCI workload definition to have more RAM


ciao-launcher/qemu.go
```diff
@@ -289,57 +289,76 @@ func (q *qemu) deleteImage() error {
...
-	/*
-	   3, what do you mean 3.  Well, it turns out that files passed to child
-	   processes via cmd.ExtraFiles have different fds in the child and the
-	   parent.  In the child the fds are determined by the file's position
-	   in the ExtraFiles array + 3. Since we're only specifying a single
-	   file we end up with an child fd of 3.
-	*/
+		f, err := os.OpenFile(tapDev, os.O_RDWR, 0666)
+		if err != nil {
+			glog.Errorf("Failed to open tap device %s: %s", tapDev, err)
```
markdryan: Won't we leak fds here if we get an error, i.e, if we fail to open a file for the 3rd queue, but the first 2 succeed.

ciao-launcher/qemu.go
```diff
@@ -289,57 +289,76 @@ func (q *qemu) deleteImage() error {
...
+			return nil, nil, err
+		}
+		fds[q] = f
+		/*
+		   3, what do you mean 3.  Well, it turns out that files passed to child
+		   processes via cmd.ExtraFiles have different fds in the child and the
+		   parent.  In the child the fds are determined by the file's position
+		   in the ExtraFiles array + 3.
+		*/
+		fdParam = fmt.Sprintf("%s%s%d", fdParam, fdSeperator, q+3)
```
markdryan: Might be more efficient to use a bytes.Buffer here.

ciao-launcher/qemu.go
```diff
@@ -453,19 +472,29 @@ func (q *qemu) startVM(vnicName, ipAddress string) error {
...
+		if q.cfg.NetworkNode {
+			var err error
+			var macvtapParam []string
+			//TODO: @mcastelino get from scheduler/controller
+			numQueues := 4
+			macvtapParam, fds, err = computeMacvtapParam(vnicName, q.cfg.VnicMAC, numQueues)
+			if err != nil {
+				return err
+			}
+			for _, f := range fds {
```
markdryan: Probably more efficient to do
```
defer func(fds []*os.File) {
   for _, f := range fds {
     _ = f.Close()
   }
}(fds)
``` 

This will also keep errcheck happy.  errcheck will complain about defer f.Close().  I realise the code was already like this.  It was on my today list to fix.

Also, I don't see why you check to see if the file handle is nil?  I don't see how that could be the case.

ciao-launcher/qemu.go
mcastelino: Thanks I missed it. Fixed.

ciao-launcher/qemu.go
mcastelino: Created a new function to handle both the error and success cases consistently.

mcastelino:
Rebased to the tip.

markdryan:
PR looks fine to me now.  Feel free to merge.

PR #76, opened 2016-04-28
travis: Create SSNTP certs and run SSNTP tests

sameo:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: go
+sudo: required
 
 go:
     - 1.6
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,11 +1,23 @@
...
+sudo: required
 
 go:
     - 1.6
     - tip
 
 go_import_path: github.com/01org/ciao
 
+install:
+   - go install -v github.com/01org/ciao/ssntp/ciao-cert
```
markdryan: I think if we used before_script instead of install we wouldn't need to build ciao-cert explicitly.  travis is doing a go get ./... during install so it should build ciao-cert with everything else.  Just conjecture though.  We'd need to try to see.

.travis.yml
```diff
@@ -1,11 +1,23 @@
...
+   - sudo mkdir /etc/pki/ciao/
```
markdryan: Seems to be failing.  Maybe it wants a -p

sameo:
@markdryan Indeed, before_script allows us to get rid of the explicit ciao-cert build, thanks.

Travis still seems to be failing on go tip, where for some reasons GOROOT does not seem to be exported.

Issue #81, opened 2016-05-02
Should use parameterized SQL queries over fmt.Sprintf 


gcmurphy:
Failure to use parameterized SQL queries could lead to SQL injection:

 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1007
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1119
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1431
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1432
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1448
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1467
 * https://github.com/01org/ciao/blob/master/ciao-controller/internal/datastore/sqlite3db.go#L1532

Should use a prepared statement of similar instead of fmt.Sprintf here.

kaccardi:
Hi, thanks!  I have created a pull request with the changes to address this issue, I'd appreciate your review. 
#87

PR #87, opened 2016-05-02
ciao-controller: datastore: use prepared statements rather than Sprintf

kaccardi:
Address issue #81

Go uses prepared statements for you under the covers... unless you
override this by using Sprintf.  Since prepared statements are
more secure, only use Sprintf if you have a very good reason - which
I didn't.


gcmurphy:
👍 looks good!

PR #111, opened 2016-05-09
travis: run ciao-controller/internal/datastore tests

chavafg:
run ciao-controller/internal/datastore tests

Signed-off-by: Salvador Fuentes <salvador.fuentes@intel.com>


markdryan:
Looks good.  However, could you just update the commit message to include a description?

PR #133, opened 2016-05-18
Networking: Use Testify for all Networking Unit test cases

mcastelino:
Using testify for unit test cases reduces the complexity of the unit test cases.
This reduces the cyclomatic complexity significantly.

### DIFF
--- a/networking/libsnnet/bridge_test.go
+++ b/networking/libsnnet/bridge_test.go
@@ -17,9 +17,20 @@
 package libsnnet
 
 import (
-	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 )
+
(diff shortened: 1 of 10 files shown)


markdryan:
The question is, should we vendor testify?  Vendoring test dependencies is a little trickier due to the way go list works but it can be done.  I can enter an issue for this.

mcastelino:
@markdryan it may not be worth the effort to vendor test dependencies. Also when reading the documentation later I noticed that testify supports versioning. Let me check if the v1 suffices for us and use that.

```
* Latest version: go get github.com/stretchr/testify
* Specific version: go get gopkg.in/stretchr/testify.v1
```

Issue #66, opened 2016-04-27
Ciao minimal configuration


sameo:
## Ciao Minimal Configuration
One of Ciao major goal is to get as close to a zero config cluster as possible. This issue documents how we want to get there.

### Overview
We want the cluster configuration to be as centralized as possible and this central point of configuration is the Ciao scheduler.
All launchers running on the compute and networking nodes should be able to run with zero configuration input by default.
The controller may need additional configuration information for fetching its HTTPS credentials.

### Configuration data
The configuration data is a [YAML payload] (https://github.com/01org/ciao/blob/master/payloads/configure.go)  that is part of 2 SSNTP frames:

1. The [CONNECTED](https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L234) status frame: Whenever a SSNTP client successfully connects to the Scheduler it will receive a CONNECTED frame back which payload is the configuration one.
2. The [CONFIGURE] (https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L213) command frame: Whenever a SSNTP client wants to update part of the cluster configuration it is allowed to send a [CONFIGURE](https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L213) command to the Scheduler. It is up to the Scheduler to decide if the sender is allowed to update the cluster overall configuration or not. Typically we should only allow Controller to update this piece of information.

The configuration payload currently contains the following piece of information: 

* The identity service type (e.g. `Keystone`) and its URL 
* The image service type (e.g. `Glance`) and its URL
* Configuration data for the controller:
  * The compute API configuration data:
     * The compute API port
     * The compute API certificates
  * The identity configuration data:
    * The identity credentials
* Configuration data for the launchers:
   * The cluster management network
   * The cluster compute network
   * The disk limit flag
   * The memory limit flag

### Configuration updates
Both clients and servers can propagate cluster configuration updates by sending a [CONFIGURE](https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L213) command.
The Scheduler will be responsible for forwarding all valid configuration changes to all clients (Controller, Agents).

As the central configuration point, the Scheduler is also responsible for updating the stored configuration data. The storage back end details are implementation specific.

### Scheduler
The Scheduler is Ciao's main configuration point and as such is responsible for:
* Propagating the latest configuration data through the [CONNECTED](https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L234) status frames it sends back to connected clients.
[...]

PR #168, opened 2016-05-25
ciao-launcher: Add Config package support for  ciao-launcher

obedmr:
Add config package support for enabling Scheduler-provided cluster
configuration.

ciao-launcher start commands should look like:
```
$ ./ciao-launcher --cacert=/path/to/cacert --cert=/path/to/cert \
       --server=<server-ip> --network=<cn|nn>
```

The options that were removed are:
```
--compute-net
--mgmt-net
--disk-limit
--mem-limit
```

Implements part of issue #66

Above configurations are provided by Scheduler at launch stage.

### DIFF
--- a/ciao-launcher/instance_test.go
+++ b/ciao-launcher/instance_test.go
@@ -249,6 +249,10 @@ func (v *instanceTestState) deleteInstance(t *testing.T, ovsCh chan interface{},
 	}
 }
 
+func (v *instanceTestState) ClusterConfiguration() (payloads.Configure, error) {
+	return payloads.Configure{}, nil
+}
+
 func cleanupShutdownFail(t *testing.T, instance string, doneCh chan struct{}, ovsCh chan interface{}) {
 	_ = os.RemoveAll(path.Join(instancesDir, instance))
 	shutdownInstanceLoop(doneCh, ovsCh)
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -99,12 +99,8 @@ func init() {
 	flag.StringVar(&serverURL, "server", "", "URL of SSNTP server")
 	flag.StringVar(&serverCertPath, "cacert", "/etc/pki/ciao/CAcert-server-localhost.pem", "Client certificate")
 	flag.StringVar(&clientCertPath, "cert", "/etc/pki/ciao/cert-client-localhost.pem", "CA certificate")
-	flag.StringVar(&computeNet, "compute-net", "", "Compute Subnet")
-	flag.StringVar(&mgmtNet, "mgmt-net", "", "Management Subnet")
 	flag.Var(&networking, "network", "Can be none, cn (compute node) or nn (network node)")
 	flag.BoolVar(&hardReset, "hard-reset", false, "Kill and delete all instances, reset networking and exit")
-	flag.BoolVar(&diskLimit, "disk-limit", true, "Use disk usage limits")
-	flag.BoolVar(&memLimit, "mem-limit", true, "Use memory usage limits")
 	flag.BoolVar(&simulate, "simulation", false, "Launcher simulation")
 }
 
@@ -133,6 +129,7 @@ type serverConn interface {
 	Close()
 	isConnected() bool
 	setStatus(status bool)
+	ClusterConfiguration() (payloads.Configure, error)
 }
 
 type ssntpConn struct {
@@ -318,6 +315,34 @@ func processCommand(conn serverConn, cmd *cmdWrapper, ovsCh chan<- interface{}) 
 	}
 }
 
+func startNetwork(doneCh chan struct{}) error {
+	if networking.Enabled() {
+		ctx, cancelFunc := context.WithCancel(context.Background())
+		ch := initNetworking(ctx)
+		select {
+		case <-doneCh:
+			glog.Info("Received terminating signal.  Quitting")
+			cancelFunc()
+			return fmt.Errorf("Init network cancelled.")
+		case err := <-ch:
+			if err != nil {
+				glog.Errorf("Failed to init network: %v\n", err)
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func printClusterConfig() {
+	glog.Info("Cluster Configuration")
+	glog.Info("-----------------------")
+	glog.Infof("Compute Netowork:     %v", computeNet)
+	glog.Infof("Management Netowork:  %v", mgmtNet)
+	glog.Infof("Disk Limit:           %v", diskLimit)
+	glog.Infof("Memory Limit:         %v", memLimit)
+}
+
 func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
 
 	defer func() {
@@ -344,7 +369,7 @@ func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
 		cmdCh: make(chan *cmdWrapper),
 	}
 
-	ovsCh := startOverseer(&wg, client)
+	var ovsCh chan<- interface{}
 
 	dialCh := make(chan error)
 
@@ -369,6 +394,27 @@ DONE:
 			if err != nil {
 				break DONE
 			}
+			clusterConfig, err := client.conn.ClusterConfiguration()
+			if err != nil {
+				glog.Errorf("Unable to get Cluster Configuration %v", err)
+				client.conn.Close()
+				break DONE
+			}
+			computeNet = clusterConfig.Configure.Launcher.ComputeNetwork
+			mgmtNet = clusterConfig.Configure.Launcher.ManagementNetwork
+			diskLimit = clusterConfig.Configure.Launcher.DiskLimit
+			memLimit = clusterConfig.Configure.Launcher.MemoryLimit
+			printClusterConfig()
+
+			err = startNetwork(doneCh)
+			if err != nil {
+				glog.Errorf("Failed to start network: %v\n", err)
+				client.conn.Close()
+				break DONE
+			}
+			defer shutdownNetwork()
+
+			ovsCh = startOverseer(&wg, client)
 		case <-doneCh:
 			client.conn.Close()
 			if !dialing {
@@ -391,7 +437,9 @@ DONE:
 		}
 	}
 
-	close(ovsCh)
+	if ovsCh != nil {
+		close(ovsCh)
+	}
 	wg.Wait()
 	glog.Info("Overseer has closed down")
 }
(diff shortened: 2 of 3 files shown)


markdryan:
@obedmr Could you update the Readme.md (https://github.com/01org/ciao/tree/master/ciao-launcher) as the command line options have changed?

ciao-launcher/main.go
```diff
@@ -333,6 +331,22 @@ func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
 		dialCh <- err
 	}()
 
+	var clusterConfig payloads.Configure
+	var data []byte
+	for {
```
markdryan: I don't really like this loop.  It will use loads of CPU and might never exit.  It will also stop launcher from exiting cleanly when there's a SIGTERM.  There must be a better way to access the SSNTP provisioned cluster config.  Is this data guaranteed to be present once dial suceeds?  If so we just need to move this call down a little bit, 
```
		select {
		case err := <-dialCh:
			dialing = false
			if err != nil {
				break DONE
			}

                        < Could we put the call to client.ClusterConfiguration() here?>
```

ciao-launcher/main.go
```diff
@@ -97,12 +99,8 @@ func init() {
 	flag.StringVar(&serverURL, "server", "", "URL of SSNTP server")
 	flag.StringVar(&serverCertPath, "cacert", "/etc/pki/ciao/CAcert-server-localhost.pem", "Client certificate")
 	flag.StringVar(&clientCertPath, "cert", "/etc/pki/ciao/cert-client-localhost.pem", "CA certificate")
-	flag.StringVar(&computeNet, "compute-net", "", "Compute Subnet")
-	flag.StringVar(&mgmtNet, "mgmt-net", "", "Management Subnet")
 	flag.Var(&networking, "network", "Can be none, cn (compute node) or nn (network node)")
 	flag.BoolVar(&hardReset, "hard-reset", false, "Kill and delete all instances, reset networking and exit")
-	flag.BoolVar(&diskLimit, "disk-limit", true, "Use disk usage limits")
-	flag.BoolVar(&memLimit, "mem-limit", true, "Use memory usage limits")
```
markdryan: When you remove this line the defaults for these globals will change from true to false, i.e., you'll disable mem and disk checking by default, which is not what we want.

ciao-launcher/main.go
```diff
@@ -528,6 +542,14 @@ func startLauncher() int {
 	timeoutCh := make(chan struct{})
 	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
 
+	configCh := make(chan payloads.Configure)
+	go connectToServer(doneCh, statusCh, configCh)
+	clusterConfig := <-configCh
```
markdryan: We might as well move the networking initialisation code into connectToServer after Dial has returned.  This way we can get rid of the channel.

ciao-launcher/main.go
markdryan: Actually, I think it will have to work like this.  Otherwise I don't see how we can ensure that we have the correct configuration data before we start processing SSNTP commands.  Your original code obviously does ensure this, but not in quite the right way.

ciao-launcher/main.go
```diff
@@ -333,6 +331,22 @@ func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
...
 
+	var clusterConfig payloads.Configure
+	var data []byte
+	for {
+		data = client.ClusterConfiguration()
+		if data != nil {
+			break
+		}
+	}
+	err := yaml.Unmarshal(data, &clusterConfig)
```
sameo: I just pushed commit 7d9337799a7442190a0b635da0ea3aa16298382d that changes the client.ClusterConfiguration() call to return a payloads.Configure structure instead.

ciao-launcher/main.go
mrkz: @markdryan so then `disk-limit` and `mem-limit` should default to **true**?
We had a conversation with @sameo and agreed to have them set to false by default, but of course those can be changed if this is not correct.

markdryan: Yep I think they should be true by default.  Otherwise launcher will not generate snntp.FULL messages and a node could become overwhelmed.  Even if we were to switch the defaults to false, we should do this in a separate commit so it's obvious that a change has been made.

ciao-launcher/main.go
obedmr: hi @markdryan,  I'm still having trouble with this, as far I can see, it could be good if we only have network variables (computeNet and mgmtNet), but, in CluserConfiguration we also have the diskLimit and memLimit variables which are used by the overseer.go. So, if it is loaded in a go routine, not sure how to pass those variables to the overseer which is consuming the main.go context. Do you have a recommendation for this?,  I updated my commit by applying the rest of comments.

markdryan: @obedmr This is a good point.  There'll be a race condition if we don't also delay the start of the overseer until after Dial has completed and we've retrieved our ClusterConfiguration and initialised the 4 global variables.  I think this should work, although I haven't tried.

obedmr: @markdryan @sameo, I updated my commit, when you have a chance could you please take a look on it and comment.

markdryan: @obedmr I'll take a look this afternoon.

markdryan:
The unit tests no longer compile with this patch applied.

ClusterConfiguration needs to be added to instanceTestState in instance_test.go

obedmr:
thanks @markdryan, I will update unit tests and see how we can start overseer after dial is succeed

ciao-launcher/instance_test.go
```diff
@@ -250,6 +250,10 @@ func (v *instanceTestState) deleteInstance(t *testing.T, ovsCh chan interface{},
 	}
 }
 
+func (v *instanceTestState) ClusterConfiguration() (payloads.Configure, error) {
+	return payloads.Configure{},nil
+}
+
```
markdryan: This file needs to be gofmted.  The formatting is incorrect.

markdryan:
gofmt needs to be run on both the launcher files.

ciao-launcher/main.go
```diff
@@ -356,6 +353,28 @@ DONE:
 			if err != nil {
 				break DONE
 			}
+			clusterConfig,err := client.conn.ClusterConfiguration()
+			if err != nil {
+				glog.Errorf("Unable to get Cluster Configuration %v", err)
+				dialCh <- err
```
markdryan: Is this a copy and paste error?  Maybe it was in the code I sent you.  I think this will deadlock.  There's no one to read from dialCh.  We can just remove this line.

ciao-launcher/main.go
```diff
@@ -356,6 +353,28 @@ DONE:
...
+				ch := initNetworking(ctx)
+				err := <-ch
+				if err != nil {
+					glog.Errorf("Failed to init network: %v\n", err)
+					break DONE
+				}
+			}
+			defer shutdownNetwork()
+
+			ovsCh = startOverseer(&wg, client)
```
markdryan: connectToServer is now too complex

```
16 main connectToServer ciao-launcher/main.go:320:1
```

We'll need to split it up.

markdryan:
Make sure you add the copyright notice to all the new files.

markdryan:
@obedmr Apart from the 3 comments (gofmt, gocyclo and the deadlock) the launcher change looks fine.  I'm still a bit confused about whether the defaults for memlimit and disklimit will change.  This PR should not change the defaults.  I can't quite figure out what would happen if the current PR was merged.

markdryan:
One more comment.  What is the plan for the cnci-agent, which also has mgmt-net and compute net parameters?

tpepper:
On Fri 03 Jun at 08:24:25 -0700 notifications@github.com said:
> Make sure you add the copyright notice to all the new files.

See "template.go" in the toplevel directory.

-- 
Tim Pepper -- Linux OS Engineering
Intel Open Source Technology Center

ciao-launcher/main.go
obedmr: @markdryan, not sure what happened by I'm getting 14, is that ok, or should we still split it?
14 main connectToServer ciao-launcher/main.go:316:1

obedmr:
@markdryan @sameo, I'm just waiting for @mrkz pull request #149 to be updated and then I'll do more testing. For now, I'm taking a look on (memLimit, diskLimit) and cnci-agent things.

ciao-launcher/main.go
markdryan: 14 is fine but I'm getting 16.  Have you rebased your patch set against the latest version of master?  If you have and you are still getting 14, perhaps I merged your patch incorrectly to my local workspace.  In any case, if there really is a problem here, Travis CI will fail the build when you update your PR.

obedmr:
@markdryan, now ``connectToServer`` function gives 15 with gocyclo, I split the network initialization in another function,  regarding the memLimit and diskLimit, they're being initialized at https://github.com/01org/ciao/blob/master/configuration/configuration.go#L50-57

In case that they're not configured in yaml file, they will have ``true`` by default.

ciao-launcher/main.go
```diff
@@ -318,6 +315,16 @@ func processCommand(conn serverConn, cmd *cmdWrapper, ovsCh chan<- interface{})
 	}
 }
 
+func startNetwork() error {
+	if networking.Enabled() {
```
markdryan: This code is not quite right.  The original networking code also selected on the signal channel so that the code that inits the network could not block a launcher exit via CTRL-C or SIGTERM.  This code has been deleted.  Launcher will still be able to shutdown but it will no longer be able to shutdown cleanly.

You don't have the signal channel in the connectToServer go routine but you do have doneCh which is closed when we received a signal

```
func startNetwork(doneCh chan struct{}) error {
	  if networking.Enabled() {
                ctx, _ := context.WithCancel(context.Background())
		ch := initNetworking(ctx)
		select {
		case <-doneCh:
			glog.Info("Received terminating signal.  Quitting")
			cancelFunc()
			return fmt.Error("Shutting down")
		case err := <-ch:
			if err != nil {
				glog.Errorf("Failed to init network: %v\n", err)
				return err
			}
		}
}
```

markdryan:
@obedmr Looks okay but I'd like to give it a quick test.  Will do so first thing tomorrow.  Is there anything preventing us from merging this now?  If not and it works, I'll merge tomorrow morning.

obedmr:
@mrkz, cool. I will be testing this Today in a clean environment an will
let you know if there's something else.

ciao-launcher/main.go
```diff
@@ -369,6 +385,23 @@ DONE:
 			if err != nil {
 				break DONE
 			}
+			clusterConfig, err := client.conn.ClusterConfiguration()
+			if err != nil {
+				glog.Errorf("Unable to get Cluster Configuration %v", err)
```
markdryan: @obedmr  Sorry, about this but I just noticed that we need to call

client.conn.Close()

here to shutdown the connection to the SSNTP server.

ciao-launcher/main.go
```diff
@@ -369,6 +385,23 @@ DONE:
...
+				break DONE
+			}
+			computeNet = clusterConfig.Configure.Launcher.ComputeNetwork
+			mgmtNet = clusterConfig.Configure.Launcher.ManagementNetwork
+			diskLimit = clusterConfig.Configure.Launcher.DiskLimit
+			memLimit = clusterConfig.Configure.Launcher.MemoryLimit
+			err = startNetwork(doneCh)
+			if err != nil {
+				glog.Errorf("Failed to start network: %v\n", err)
```
markdryan: We need to call client.conn.Close() here as well.

markdryan: I think it would also be useful to add a log statement that outputs the values of all the configuration variables to the launcher logs.

ciao-launcher/main.go
obedmr: Done

obedmr:
@markdryan missing client.conn.Close() function calls were added and printClusterConfig function was added

PR #174, opened 2016-05-26
Add documentation for creation of compute service and endpoint

leoswaldo:
The compute service and its corresponding endpoint needs to be created
in order to gain compatibility with openstack tools API calls for
novaclient which is used by tools as OpenStack client (cli), Rally
Benchmarking and others


sameo:
@leoswaldo The Openstack compatibility layer is currently implemented in ciao-controller. Could you do that README.md change to the [ciao-controller README](https://github.com/01org/ciao/blob/master/ciao-controller/README.md) instead of the launcher one ?

leoswaldo:
@sameo I moved it to ciao-controller README

PR #176, opened 2016-05-27
ciao-launcher:  Add unit tests for the instanceLoop

markdryan:
These commits add some unit tests to the instanceLoop go routine.  The tests work by using channels and interfaces to fake the overseer and the connection to the SSNTP server, allowing us to fully test the instanceLoop.  A few small changes were needed to the launcher code to permit this.  The travis build file also needed to be changed to pre-create the /var/lib/ciao/instances directory.  Doing this in the travis files allows the launcher unit tests to run as non-root.

This commit should bring the test coverage of launcher up to 16.2%.


markdryan:
Rebased against master at 14:20 BST.  I also ran a quick test on BigNuc.  Was able to start and delete 100 instances.

PR #177, opened 2016-05-27
payloads: Fix misspelling

sameo:
Introduced when implementing the flavors/detail endpoint.


kaccardi:
I'm not sure why this is failing the ci.  I told github to update it to master, and then it started failing.

markdryan:
It's weird.  It looks like the ssntp unit tests crashed.  My guess is TestClientReconnect.  That's usually what it means when you see 'NOT RUN' in test-cases.  Of course, it could always be a bug in test-cases.  I'll update output test-cases to print stderr from go test if it panics.  Next time we see something like this we should get some more precise info.

My guess is that if the PR is regenerated the build will pass.

kaccardi:
well, you were right. It magically fixed.

PR #178, opened 2016-05-27
Topic/glance

albertomurillo:
This PR address the following issues
- Implements delete image
- Implements download image to file
- Implements create and upload image in the same call
- Remove hardcoded values from create_image
- Add missing functions to create_image

### DIFF
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -1074,6 +1074,6 @@ func main() {
 		if *imageID == "" {
 			fatalf("Missing required -image-id parameter")
 		}
-		downloadTenantImage(*identityUser, *identityPassword, id, *imageID)
+		downloadTenantImage(*identityUser, *identityPassword, id, *imageID, *imagePath)
 	}
 }
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -108,6 +108,7 @@ var (
 	createImage      = flag.Bool("create-image", false, "Create a new image")
 	uploadImage      = flag.Bool("upload-image", false, "Upload an image")
 	downloadImage    = flag.Bool("download-image", false, "Download an image")
+	deleteImage      = flag.Bool("delete-image", false, "Delete an image")
 	minDiskGigabytes = flag.Int("min-disk-size", 0, "Minimum amount of RAM in GB")
 	minRAMMegabytes  = flag.Int("min-ram-size", 0, "Minimum disk size in MB")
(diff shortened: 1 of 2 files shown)


ciao-cli/main.go
```diff
@@ -1031,7 +1038,12 @@ func main() {
 		if *imageName == "" {
 			fatalf("Missing required -image-name parameter")
 		}
-		createTenantImage(*identityUser, *identityPassword, id, *imageName)
+		if *imageVisibility != "private" && *imageVisibility != "public" {
```
sameo: Here you want to define e.g. `type visibility string`, implement `func (role *visibility) Set(value string) error` and then call
```
var imageVisibility visibility
flag Var(&imageVisibility, ""image-visibility", "Image visibility. Value can be public or private")
```
The `Set()` method will allow you to define what's acceptable for a visibility value.
The same logic should apply to container and disk formats.

ciao-cli/main.go
```diff
@@ -1031,7 +1038,12 @@ func main() {
...
+			fatalf("image-visibility can be either 'private' or 'public'")
+		}
+		createTenantImage(*identityUser, *identityPassword, id, *imageName,
+			*imagePath, *containerFormat, *diskFormat, *imageVisibility, *imageTags,
+			*minDiskGigabytes, *minRAMMegabytes, *protected)
```
sameo: At that point it would make sense to create a `type imageOptions struct` and pass a pointer to `createTenantImage()`.

ciao-cli/main.go
albertomurillo: "github.com/rackspace/gophercloud/openstack/imageservice/v2/images"

already defines an ImageVisibility type,
Shouldn't it be better to include it and add the functions Set(value string) error and String() string there ?

ciao-cli/main.go
albertomurillo: There is also a type CreateOpts struct in "github.com/rackspace/gophercloud/openstack/imageservice/v2/images"

so same question as above, to avoid code duplication should I pass a CreateOpts struct to createTenantImage ?

sameo: Yes, here it would make sense.

ciao-cli/main.go
sameo: I'd prefer to use our own for now. Once our initial PR on gophercloud is accepted we can add the Set() and String() method as a 2nd PR and move to ImageVisibility then.
I'd prefer not to add those 2 methods now as it may look awkward for the gophercloud reviewer.

Issue #183, opened 2016-05-27
data-race in ssntp unit tests


markdryan:
Reproduced inside a docker container (running on docker-machine on OSX).  I guess this is a little closer to the environment travis uses.  Here are the races

```
==================
WARNING: DATA RACE
Read by goroutine 55:
  github.com/01org/ciao/ssntp.(*Server).Stop()
      /root/go/src/github.com/01org/ciao/ssntp/server.go:297 +0xb0
  github.com/01org/ciao/ssntp.testConnectVerifyCertificate()
      /root/go/src/github.com/01org/ciao/ssntp/ssntp_test.go:887 +0x34d
  github.com/01org/ciao/ssntp.TestConnectVerifyClientCertificateNegative()
      /root/go/src/github.com/01org/ciao/ssntp/ssntp_test.go:927 +0x3e
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:473 +0xdc

Previous write by goroutine 57:
  github.com/01org/ciao/ssntp.(*Server).Serve()
      /root/go/src/github.com/01org/ciao/ssntp/server.go:265 +0xf5f

Goroutine 55 (running) created at:
  testing.RunTests()
      /usr/local/go/src/testing/testing.go:582 +0xae2
  testing.(*M).Run()
      /usr/local/go/src/testing/testing.go:515 +0x11d
  github.com/01org/ciao/ssntp.TestMain()
      /root/go/src/github.com/01org/ciao/ssntp/ssntp_test.go:2337 +0x2d5
  main.main()
      github.com/01org/ciao/ssntp/_test/_testmain.go:194 +0x20a

Goroutine 57 (running) created at:
  github.com/01org/ciao/ssntp.testConnectVerifyCertificate()
      /root/go/src/github.com/01org/ciao/ssntp/ssntp_test.go:879 +0x2b7
  github.com/01org/ciao/ssntp.TestConnectVerifyClientCertificateNegative()
      /root/go/src/github.com/01org/ciao/ssntp/ssntp_test.go:927 +0x3e
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:473 +0xdc
==================
==================
WARNING: DATA RACE
Read by goroutine 55:
  crypto/tls.(*listener).Close()
      <autogenerated>:36 +0x54
  github.com/01org/ciao/ssntp.(*Server).Stop()
      /root/go/src/github.com/01org/ciao/ssntp/server.go:298 +0x104
[...]

PR #186, opened 2016-05-30
sameo/issues/183

sameo:
This PR fixes issue #183 by 

1. Providing a ssntp.server.Serve() synchronization method so that we no longer randomly wait 500ms before assuming the SSNTP server is ready to accept new clients. In a properly implemented SSNTP server we will then no longer call Stop() before we know for sure Serve() is done (And listener is set).
2. Adding a server.listener mutex to protect Stop() against Serve() and vice versa.


markdryan:
Seems to solve the problem. I'm unable to reproduce the --race errors inside my docker container with these patches applied.

PR #204, opened 2016-06-03
ciao-controller: Use specific constructors for compute structures

### LINKED ISSUE #166: CIAO- CN api returning null instead of empty list
CN api returning null instead of empty list

`openstack --debug server list`  if a tenant has no  instances/vms it returns: 

`RESP BODY: {"total_servers":0,"servers":null} `

Which is not compatible with novaclient, since http://git.openstack.org/cgit/openstack/python-novaclient/tree/novaclient/base.py line 259  expects a iterable.

When a project does not have any instance, it should return` {"servers": [], ...}`, where servers is a empty list, so novaclient can iterate in that (empty list), otherwise it would cause a "**'NoneType' object is not iterable**"

sameo:
We need to allocate compute structure slices so that JSON
marshalling produces an empty slice and not a nil pointer
when they represent empty lists.

Fixes #166

### DIFF
--- a/ciao-controller/compute.go
+++ b/ciao-controller/compute.go
@@ -107,7 +107,7 @@ func pagerQueryParse(r *http.Request) (int, int, string) {
 }
 
 func (pager *serverPager) getInstances(filterType pagerFilterType, filter string, instances []*types.Instance, limit int, offset int) ([]byte, error) {
-	var servers payloads.ComputeServers
+	servers := payloads.NewComputeServers()
 
 	servers.TotalServers = len(instances)
 	pageLength := 0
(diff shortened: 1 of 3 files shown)


markdryan:
Would it help if the nil arrays didn't appear in the json at all, i.e., if the output was

```
{"total_servers":0}
```

instead of

```
{"total_servers":0,"servers":null}
```

If so you could simply append omitempty to the json tags, e.g.,

```
struct {
		TotalServers int   `json:"total_servers"`
		Servers      []int `json:"servers,omitempty"`
}
```

sameo:
@markdryan I think all openstack tools expect to get an empty array when those lists are empty, but most likely not the array does not appear at all.
I'll ask in the #166 thread.

sameo:
@markdryan Interested parties from #166 confirm an empty list would be better.

markdryan:
Sounds good.  What about ciao-cli?  Does this need to be updated as well?  If so will do this in a separate commit?

Also, I'm a bit confused by this PR.  It doesn't seem to change anything.  The PR contains a commit and a revert of the commit.

sameo:
@markdryan Sorry about that, I refreshed and pushed my branch...
No, ciao-cli does not need to be updated. It's only python and node.js clients that expect an iterable output.

markdryan:
Great.  So good for me.  I guess we just need an okay from @kaccardi since we touch controller code and then we can merge.

kaccardi:
looks good to me.

PR #205, opened 2016-06-06
ssntp: Add SSNTP Frame Origin field

sameo:
Origin tracks the first creator and sender of an SSNTP frame.
This allows for any frame receiver to track who the first sender is, even across SSNTP forwarding.

No significant throughput regressions were identified with this new field added to the SSNTP frames.

### DIFF
--- a/ssntp/frame.go
+++ b/ssntp/frame.go
@@ -62,10 +62,17 @@ type FrameTrace struct {
 
 // Frame represents an SSNTP frame structure.
 type Frame struct {
-	Major         uint8
-	Minor         uint8
-	Type          Type
-	Operand       uint8
+	Major   uint8
+	Minor   uint8
+	Type    Type
+	Operand uint8
+
+	// Origin is the frame first sender and creator UUID.
+	// When a SSNTP frame is forwarded by a server, the client
+	// then only sees a new frame coming but it can not tell
+	// who the frame creator and first sender is. This method
+	// allows to fetch such information from a frame.
+	Origin        uuid.UUID
 	PayloadLength uint32
 	Trace         *FrameTrace
 	Payload       []byte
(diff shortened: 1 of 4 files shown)


ssntp/frame.go
```diff
@@ -291,3 +292,14 @@ func (f Frame) DumpTrace() (*payloads.FrameTrace, error) {
...
 	return &s, nil
 }
+
+// OriginUUID is the frame first sender and creator UUID.
+// When a SSNTP frame is forwarded by a server, the client
+// then only sees a new frame coming but it can not tell
+// who the frame creator and first sender is. This method
+// allows to fetch such information from a frame.
+func (f Frame) OriginUUID() (origin uuid.UUID) {
+	copy(origin[:], f.Origin[:16])
```
markdryan: I'm trying to figure out whether the [:16] slice is safe here.  As far as I can tell we're slicing a slice here and not an array, so there's no compile time guarantee that the slice will work.  But what I don't know is whether it's actually possible to for a SSNTP client to craft a frame with an Origin UUID of  less than 16 bytes?

sameo: @markdryan so you're suggesting that I check if len(f.Origin) == 16 ?
It's in theory not possible for a SSNTP client to craft a frame with an Origin UUID of less than 16 bytes.

markdryan: Or just make f.Origin a uuid.UUID.  That way you'd be sure that the copy would always work and as a bonus you wouldn't need to change session.go.  It's probably more space efficient as well, but the semantics will be slightly different.  If the session.src from which the Origin was initialised changed, the change would no longer be reflected in Origin.

kaccardi:
Looks good to me.

PR #210, opened 2016-06-06
ssntp: Do not dereference nil PEM blocks

### LINKED ISSUE #208: Do not use nil PEM blocks when parsing certificate
First go-fuzz findings:

Could not decode PEM for /tmp/ssntp-fuzz-certs/CACert
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x4b09c7]

sameo:
First go-fuzz catch !

Fixes #208


ssntp/ssntp.go
```diff
@@ -936,7 +936,7 @@ func (config *Config) parseCertificateAuthority() ([]string, []string, error) {
 
 	certBlock, _ := pem.Decode(caPEM)
 	if certBlock == nil {
-		fmt.Printf("Could not decode PEM for %s\n", config.CAcert)
+		return nil, nil, fmt.Errorf("Could not decode PEM for %s\n", config.CAcert)
```
markdryan: Should we remove the \n now this is an error string?

PR #214, opened 2016-06-06
configuration: Add Init method for Configure payload default values

obedmr:
Based on requirements for CIAO Minimal Config Package, we need
to have some default initial values:
https://github.com/01org/ciao/issues/66

Below are the config package related (In-Progress) PRs:
config_package PR:  https://github.com/01org/ciao/pull/149
config_launcher PR: https://github.com/01org/ciao/pull/168

### DIFF
--- a/configuration/configuration_test.go
+++ b/configuration/configuration_test.go
@@ -149,7 +149,7 @@ func emptyPayload(p payloads.Configure) bool {
 }
 
 func fillPayload(conf *payloads.Configure) {
-	fillDefaults(conf)
+	conf.InitDefaults()
 	conf.Configure.Scheduler.ConfigStorageURI = storageURI
 	conf.Configure.Controller.HTTPSCACert = httpsCACert
 	conf.Configure.Controller.HTTPSKey = httpsKey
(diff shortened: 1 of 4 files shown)


ssntp/client.go
```diff
@@ -487,6 +487,7 @@ func (client *Client) UUID() string {
 // status frame or a CONFIGURE command one.
 func (client *Client) ClusterConfiguration() (payloads.Configure, error) {
 	var conf payloads.Configure
+	conf.Init()
```
sameo: I don't think the client should be doing this, I think the configuration payload creator and sender (i.e. the scheduler) should send configuration data with proper default values. This is done here:
https://github.com/01org/ciao/blob/master/configuration/configuration.go#L50

obedmr: Yeah, agree with that, I just talked with @mrkz, @sameo should we keep that fillDefaults function or use the one at payloads/configure.go ?

payloads/configure.go
```diff
@@ -102,3 +102,13 @@ type ConfigurePayload struct {
 type Configure struct {
 	Configure ConfigurePayload `yaml:"configure"`
 }
+
+// Init initialises instances of the Configure structure.
+func (conf *Configure) Init() {
```
sameo: Could we call that one `InitDefaults()` and use it instead of `fillDefaults()` in the configuration package ?

obedmr: @sameo sure, updating it

configuration/configuration_test.go
```diff
@@ -208,15 +208,6 @@ func saneDefaults(conf *payloads.Configure) bool {
 		conf.Configure.Launcher.MemoryLimit == true)
 }
 
-func TestFillDefaults(t *testing.T) {
```
sameo: @obedmr we can still keep that one, and rename it TestInitDefaults().
It would check that conf.InitDefaults() builds a sane default configuration structure.

obedmr: Sure, done

PR #219, opened 2016-06-07
Get rid of the go-spew dependency

sameo:
ciao-cli was the only user, and we can dump JSON payloads by ourselves.


markdryan:
I'm planning to add an unvendor command (there's an issue for this) but it's not done yet.  In the meantime you need to update ciao-vendor, as you have already done, and git rm the directory, e.g,

git rm vendor/github.com/davecgh/go-spew/spew

And then commit the results.

You can check that everything is good afterwards by doing a

cd $GOPATH/src/github.com/01org/ciao
go run ciao-vendor/ciao-vendor.go check

PR #223, opened 2016-06-07
ciao-controller: Add Config package support for ciao-controller

obedmr:
Add config package support for enabling Scheduler-provided cluster
configuration.

ciao-launcher start commands should look like:
```
  $ ./ciao-controller --cacert=/etc/pki/ciao/CAcert-example.com.pem \
        --cert=/etc/pki/ciao/cert-Controller-localhost.pem --url example.com
```
The options that were removed are:
```
    -computeport
    -httpscert
    -httpskey
    -identity
    -password
    -username
```
Implements part of issue #66

Above configurations are provided by Scheduler at launch stage.

### DIFF
--- a/ciao-controller/README.md
+++ b/ciao-controller/README.md
@@ -100,16 +100,8 @@ Usage of ciao-controller/ciao-controller:
     	CA certificate (default "/etc/pki/ciao/CAcert-server-localhost.pem")
   -cert string
     	Client certificate (default "/etc/pki/ciao/cert-client-localhost.pem")
-  -computeport int
-    	Openstack Compute API port (default 8774)
   -database_path string
     	path to persistent database (default "./ciao-controller.db")
-  -httpscert string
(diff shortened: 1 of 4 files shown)


kaccardi:
These changes seem like they will work as long as the information provided is correct. LGTM.

kaccardi:
i don't want to merge just yet since it will prevent us from configuring controller until the scheduler can pass the correct config info.

sameo:
@kaccardi All our test clusters have been populated with the right configuration files now, and we documented this in the configuration [README](https://github.com/01org/ciao/blob/master/configuration/README.md).
Tomorrow I will test @obedmr PR on a couple of clusters and will let you know if it works as expected.

sameo:
@kaccardi with ciao-cli I could launch 20 instances with this patch on 2 test clusters, which means:
* The compute port was correctly set
* HTTPS certs were properly used
* controller could correctly talk to the identity service

_If_ you're fine with the code, I think we can merge that PR.

kaccardi:
yes - I'm fine with the code.

Issue #167, opened 2016-05-25
ciao-cli: implement subcommands


albertomurillo:
ciao-cli help prints all the available options
without any clue of which are dependant of others.

With support for more openstack components being added
like glance,cinder and swift, the list will grow even 
bigger without any structure.

'ciao-cli' without arguments should show the available
commands. and global options

'ciao-cli command' should show the available subcommands related to
'command'

and so on until you get a clear idea of how to run the cli.

mcastelino:
maybe something along the lines of

ciao-cli --command=launch-instances --workload=... --instances=....

sameo:
* event
  * list
  * delete
* instance
  * add
  * list
  * delete
  * restart
  * stop
* image
  * add
  * download
  * list
  * upload
* node
  * list
* tenant
  * list
* workload
   * list

sameo:
Also, we need to keep the current behaviour working as is, i.e. we should have a legacy mode where all current calls should be supported. We want to switch to the new `ciao-cli [item] [command]` mode asap and thus we won't accept any new options our commands outside of that pattern.

sameo:
@mcastelino convinced me offline that it's probably not worth it to keep the backward compatibility requirement. Let's drop it then and see how many people complain about it.

sameo:
I'd like to see both `items` and `commands` being implemented as command interfaces:

```
type Command interface {
        Run(args []string) error
        Help() string
}
```
We'd have one file per item where the item and all commands for the item would be implemented as such interface.
And then we'd have a slice mapping command strings to interface implementations

```
items := []struct {
        name string
        item   Command
}{
       {
              name: "instance",
              item:   instanceItem,
       },
       {
              name: "event",
              item:    eventItem,
       },
}
```

PR #224, opened 2016-06-08
ciao-cli: Implement subcommands

albertomurillo:
ciao-cli instance [add list delete restart stop]
ciao-cli workload [list]
ciao-cli tenant [list]
ciao-cli event [list delete]

### DIFF
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -27,12 +27,65 @@ import (
 	"net/http"
 	"os"
 	"strconv"
-	"time"
+	"text/template"
 
 	"github.com/golang/glog"
-
-	"github.com/01org/ciao/payloads"
 )
+
+// Item serves to represent a group of related commands
+type command struct {
+	SubCommands map[string]subCommand
+}
+
+// This is not used but needed to comply with subCommand interface
+func (c *command) parseArgs(args []string) []string {
+	return args
+}
+
+func (c *command) run(args []string) error {
+	cmdName := args[0]
+	subCmdName := args[1]
+	subCmd := c.SubCommands[subCmdName]
+	if subCmd == nil {
+		c.usage(cmdName)
+	}
+	args = subCmd.parseArgs(args[2:])
+	return subCmd.run(args)
+}
+
+// usage prints the available commands in an item
+func (c *command) usage(name ...string) {
+	fmt.Fprintf(os.Stderr, `ciao-cli: Command-line interface for the Cloud Integrated Advanced Orchestrator (CIAO)
+
+Usage:
+
+	ciao-cli [options] `+name[0]+` sub-command [flags]
+`)
+
+	var t = template.Must(template.New("commandTemplate").Parse(commandTemplate))
+	t.Execute(os.Stderr, c)
+
+	fmt.Fprintf(os.Stderr, `
+Use "ciao-cli `+name[0]+` sub-command -help" for more information about that item.
+`)
+	os.Exit(2)
+}
+
+// subCommand is the interface that all cli commands should implement
+type subCommand interface {
+	usage(...string)
+	parseArgs([]string) []string
+	run([]string) error
+}
+
+var commands = map[string]subCommand{
+	"instance": instanceCommand,
+	"workload": workloadCommand,
+	"tenant":   tenantCommand,
+	"event":    eventCommand,
+	"node":     nodeCommand,
+	"trace":    traceCommand,
+}
 
 var scopedToken string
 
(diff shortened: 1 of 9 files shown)


albertomurillo:
Opening pull request for early review as this is not finished yet.

albertomurillo:
TODO: 

- Implement the help command
- Add short description for items and commands 
- Parse global variables like -username, -password, etc

ciao-cli/main.go
```diff
@@ -506,68 +401,6 @@ func dumpCNCIDetails(cnciID string) {
...
-	defer resp.Body.Close()
-
-	if resp.StatusCode != http.StatusAccepted {
-		fatalf("Instance deletion failed: %s", resp.Status)
-	}
-
-	fmt.Printf("Deleted instance: %s\n", instance)
-}
-
 func actionAllTenantInstance(tenant string, osAction string) {
```
sameo: You probably want to move all `*Instance()` routines to ciao-cli/instance.go.

sameo:
@albertomurillo Besides one or two nitpicks, it looks really good to me. You will have to fix the golint errors as well.
Here are the options that your code currently does not support:
```
        listQuotas       = flag.Bool("list-quotas", false, "List quotas status for a tenant")
	listResources    = flag.Bool("list-resources", false, "List consumed resources for a tenant for the past 15mn")
	listComputeNodes = flag.Bool("list-cns", false, "List all compute nodes")
	listCNCIs        = flag.Bool("list-cncis", false, "List all CNCIs")
	listLength       = flag.Int("list-length", 0, "Maximum number of items in the reponse")
	listLabels       = flag.Bool("list-labels", false, "List all trace labels")
	dumpCNCI         = flag.Bool("dump-cnci", false, "Dump a CNCI details")
	dumpToken        = flag.Bool("dump-token", false, "Dump keystone tokens")
	clusterStatus    = flag.Bool("cluster-status", false, "List all compute nodes")
	cnci             = flag.String("cnci", "", "CNCI UUID")
	dumpLabel        = flag.String("dump-label", "", "Dump all trace data for a given label")
```

ciao-cli/tenants.go
```diff
@@ -0,0 +1,78 @@
...
+	Flag flag.FlagSet
+	all  bool
+}
+
+func (cmd *tenantListCommand) Usage() {
+	fmt.Fprintf(os.Stderr, `usage: ciao-cli [options] tenant list
+
+List tenants for the current user
+
+The add flags are:
```
sameo: 'The list flags are:'

ciao-cli/main.go
```diff
@@ -27,13 +27,50 @@ import (
...
-	"time"
+	"text/template"
 
 	"github.com/golang/glog"
-
-	"github.com/01org/ciao/payloads"
 )
 
+// Item serves to represent a group of related commands
+type Item struct {
```
sameo: @albertomurillo Sorry for the slight architectural change, but I thought about this over the weekend and I think that we should replace Item with:

``` type command struct {
             subCommands map[string]Command
}
```
and implement a complete Command interface for it (Usage, ParseArgs, Run) for it.
Run() will basically be looking for the sub-command and call into the sub-command's ParseArgs() and then Run() routine for it.

ciao-cli/main.go
```diff
@@ -188,11 +193,11 @@ func sendHTTPRequestToken(method string, url string, values []queryValue, token
 	infof("Got HTTP response (status %s)\n", resp.Status)
 
 	if resp.StatusCode >= http.StatusBadRequest {
-		respBody, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			errorf("Could not read the HTTP response %s\n", err)
+		respBody, errBody := ioutil.ReadAll(resp.Body)
```
sameo: Why are we changing this here ?

albertomurillo: to fix linter error,

err is already defined in line 147
https://github.com/albertomurillo/ciao/blob/master/ciao-cli/main.go#L147

but maybe this change should go in a separate commit

ciao-cli/main.go
```diff
@@ -855,127 +295,96 @@ func checkCompulsoryOptions() {
...
-		dumpCNCIDetails(*cnci)
+	// Print usage if no arguments are given
+	args := flag.Args()
+	if len(args) < 1 {
+		usage()
 	}
 
-	if *dumpLabel != "" {
-		dumpTraceData(*dumpLabel)
+	// Find item in cmdline args
```
sameo: So our argument parsing needs to be a more complex than this.
We want to be able to support commands, sub-commands, and options (general and subcommand specific ones) in any particular order, except that a sub-command must come after a command. So those 2 are correct:
`ciao-cli -v=2 -identity=alberto node list -cnci`
`ciao-cli node list -cnci -v=2 -identity=alberto`

For us to handle those properly and not having to force users into following a fairly complex order, we need to do 3 things:

1) Implement a helper to separate options from commands:
```
func splitArgs(args []string) (options []string, commands []string) {
	for _, a := range args {
		if len(a) > 1 && a[0] == '-' {
			options = append(options, a)
		} else {
			commands = append(commands, a)
		}
	}

	return options, commands
}
```
2) Implement a helper to generate a new FlagSet, that will automatically include all the generic options:

```
func newFlagSet() (*flag.FlagSet) {
	f := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
	
	identityURL      = f.String("identity", "", "Keystone URL")
	identityUser     = f.String("username", "", "Openstack Service Username")
	identityPassword = f.String("password", "", "Openstack Service Password")
	controllerURL    = f.String("controller", "", "Controller URL")
	tenantID         = f.String("tenant-id", "", "Tenant UUID")
	tenantName       = f.String("tenant-name", "", "Tenant name")
        computePort      = f.Int("computeport", openstackComputePort, "Openstack Compute API port")

	return f
}
```

and main.go will contain:
```
var (
	identityURL      *string
	identityUser     *string
	identityPassword *string
	controllerURL    *string
	tenantID         *string
	tenantName       *string
        computePort       *int
)
```

3) from main.go we would
    a) call `options, cmds := splitArgs(os.Args)`
    b) `command := commands[cmds[0]]` -> that will give you the command implementation
    c) `command.ParseArgs()` where `ParseArgs()` for a command will:
         * Find the sub-command
         * create a new FlagSet: `newFlagSet()`
         * Add all sub-command specific options to it
         * Call Parse() on this new flagset.
     d) Call `getCiaoEnvVariables()` and `checkCompulsoryOptions`
     e) Call the 
```
	/* If we're missing the tenant name let's try to fetch one */
	if *tenantName == "" {
		*tenantName, *tenantID, err = getTenant(*identityUser, *identityPassword, *tenantID)
		if err != nil {
			fatalf(err.Error())
		}
		warningf("Unspecified scope, using (%s, %s)", *tenantName, *tenantID)
	}

	scopedToken, *tenantID, _, err = getScopedToken(*identityUser, *identityPassword, *tenantName)
	if err != nil {
		fatalf(err.Error())
	}
```
       block
     f) Call command.Run()

ciao-cli/main.go
sameo: @albertomurillo I see. A golint failure will prevent you from passing CI, so keep this change in please.

ciao-cli/main.go
sameo: @albertomurillo I'll take that back. We want to support the following format:
`ciao-cli [generic options] [command] [subcommand] [command specific options]`

So I think your original options handling implementation is correct.

ciao-cli/main.go
```diff
@@ -27,13 +27,66 @@ import (
...
+
+	fmt.Fprintf(os.Stderr, `
+Use "ciao-cli `+name[0]+` sub-command -help" for more information about that item.
+`)
+	os.Exit(2)
+}
+
+// Command is the interface that all cli commands should implement
+type Command interface {
+	Usage(...string)
```
sameo: @albertomurillo So we don't need to export this interface, it's going to be a ciao-cli internal one.
As a consequence, it should `usage()`, `parseArgs` and `run()`.

sameo:
Besides the Command interface method names, the code looks good to me now.
There's one missing change: README.md. It should reflect the new command line handling.

sameo:
It all looks good to me now, thanks.
@tpepper or @markdryan: it's ready to be merged.

tpepper:
There's nice description of why this patch is being done and the choice to not include backwards compatibility in https://github.com/01org/ciao/issues/167.  This information should be included in the commit message.

PR #226, opened 2016-06-09
Single VM CI Enhancements and ciao-cnci-agent cert based scheduler configuration

mcastelino:
- Added pointers to the documentation maintained in the ciao wiki
- The script now downloads the workload images
- Minor fixes for bugs introduced due to typos in the certificate generation logic
- Added support in the ciao-cnci-agent to obtain the ciao-scheduler FQDN from the SSNTP certs
- Added support for cluster configuration via scheduler

With this the script automates all of the ciao specific setup, configuration and tests.

However the VM itself has to be setup properly as mentioned in the documentation. As this setup will be specific to the distro and the VM environment, it is not currently automated.

This also closes the following issues https://github.com/01org/ciao/issues/236 https://github.com/01org/ciao/issues/233


testutil/singlevm/setup.sh
```diff
@@ -60,15 +61,29 @@ cp "$ciao_scripts"/run_scheduler.sh "$ciao_bin"
...
-#Generate the CNCI VM
+#Generate the CNCI VM and seed the image and populate the image cache
 cd "$ciao_bin"
 "$GOPATH"/src/github.com/01org/ciao/networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh -c "$ciao_bin" -i "$ciao_cnci_image" -d
 qemu-img convert -f raw -O qcow2 "$ciao_cnci_image" "$ciao_cnci_image".qcow
 sudo cp "$ciao_cnci_image".qcow /var/lib/ciao/images
 cd /var/lib/ciao/images
 sudo ln -sf "$ciao_cnci_image".qcow 4e16e743-265a-4bf2-9fd1-57ada0b28904
 
+#Pre-polulate the workload caches
```
markdryan: Is this one commented out on purpose?

mcastelino:
@markdryan The download was taking a significant amount of time slowing the tests down. I will add user input to enable the download and test of this workload if the user desires.

markdryan:
@mcastelino Should I just merge this now then or are you planning to update it?

mcastelino:
@markdryan I am done with the changes for this PR. Please do review the launcher changes that I made. All of these have been self tested with the single VM CI script.

markdryan:
@mcastelino I'm not sure about the launcher changes.  Even though launcher can take the address of the scheduler from the certificate, it's still going to keep the --server flag so that the certificate URL can be overridden if desired.  If you remove the ciao.iso code, then there's the possibility that a launcher instance and the CNCI agents that it launches could end up talking to different schedulers, if launcher was itself launched with --server.

sameo:
@markdryan Do you have some specific use cases where one would need to override the certificate URL ?

markdryan:
@sameo No, but I thought the plan was to keep --server.  If we're all happy that we can get rid of --server, than the ciao.iso can go as well.

sameo:
@markdryan I thought we had to keep it because I believed that CNCI had to read it from the iso. But yesterday we realized with @mcastelino that we actually don't.
I am fine with getting rid of it.

markdryan:
@mcastelino In that case could you remove the --server parameter, update the launcher readme, and get rid of the createCiaoIso function from qemu.go.  This should improve launcher's test coverage as well.  If you prefer we could remove --server in a separate commit, but createCiaoIso shoud be removed now.

markdryan:
A few comments

1. The scripts need to be updated to work with the new cluster config stuff.  Right now launcher doesn't start up as the mgt-net parameters no longer exist.  We also need to add the default conf file for scheduler.
2. It would be nice if we could split the population of /var/lib/ciao/images and the code that runs everything into a separate script.  Or we could keep it in the same script but modify the script so that we only retrieve the test images and the cnci image if they are not already present.  This would be useful if you want to use the same vm instance to do some debugging.

mcastelino:
@markdryan I realized I needed to update the scripts to support the new config format. Will do that in this PR itself. I will also add an option to download the images on user request even if they are present, and always do it if they are not present.
That way the user can decide if the local images should be re-used or discarded.

I will add both the changes to this PR.

mcastelino:
@sameo @markdryan On the --server parameter, I had the same question as @markdryan . Are there any situations were we may still want to pass that as a parameter vs getting it from the cert. In the CNCI agent I have not removed the ability to pass the parameter. Is there a valid use case for it.

Also if the user does pass it, will it even work without the certs having the same IP or FDQN present in the certificates?

markdryan:
We need to download OVMF.fd as well for the CNCI image

mcastelino:
@markdryan 
1. Added support to download the firmware
2. Added a workaround to delete CNCI's that were being kept alive
3. Added support to make the download of dependencies optional

I have not been able to reproduce the CNCI image creation failure.

PR #231, opened 2016-06-09
configuration: Compute and Mgmt subnets are not mandatory

sameo:
On single physical interfaces, the network will autoconfigure
itself.


markdryan:
I guess we should update the minimal configuration in the readme, but we can do this later.

PR #263, opened 2016-06-15
Networking: Bridge Mode Support and Improved auto configuration

mcastelino:
The commit brings about some significant changes to ciao networking
- The ciao-launcher and networking has been enhanced to support hard-reset when network configuration is not available. (Which is typical of hard-reset usage). Without this hard-reset will not work on multi-home systems
- Re-vendor networking dependencies to bring in support for macvtap bridge mode and other enhancements
- CNCI VNIC's now use macvtap bridge mode vs VEPA which allows them to communicate with other CNCI's on the same host more efficiently.
- Allows host to CNCI communication without the need to setup hairpin mode.
- Allow bridge's to be used for management and compute networks (this will open the possibility of running the CI on travis)

The single biggest improvement due to these changes is that the the ciao single vm setup can also be run on bare metal. This allows for a high performance CI environment and eliminates the need to run VMs.

Another benefit of this is that the ciao single VM CI can now be run on any virtualization software as we no longer need hairpin mode support on the virtual network port.


tpepper:
I'd like to get @markdryan's thoughts on https://github.com/01org/ciao/pull/263/commits/a2de1bd0a6b106909f242cad4ce7da573e136a75 to understand if we can fix the issue properly.  Otherwise the other patches in the set look reasonable to me.

mcastelino:
@markdryan just expanding on the issues I saw in vendoring, the files under netlink/nl (new and renamed files) were not getting picked up. I manually copied them over into vendor after re vendoring.

markdryan:
@mcastelino ,@tpepper, the vendoring tool seems to be working fine for me.  After I revendor github.com/vishvananda/netlink the contents of this package in the vendored folder and the upstream version on $GOPATH are identical, apart from the test files which we do not copy.

Perhaps, this is where the confusion comes from.  @mcastelino I noticed that you manually added one of the test files.  We don't vedor the test files. There's no point.  To get them to work we'd need to vendor all the test dependencies as well, and this would needlessly increase our dependencies.  You'll be able to do

go run ciao-vendor.go/ciao-vendor test package version to test a vendored dependency very soon.

mcastelino:
@markdryan I cannot seem to reproduce the issue anymore. I updated the vendored dependencies again.

PR #264, opened 2016-06-15
misspell: trivial typo fix

tpepper:
Travis misspell is flagging "emtpy".


markdryan:
It seems as though misspell has been updated which is why the failure magically appeared.

PR #265, opened 2016-06-15
release: Add ciao BAT tests

kaccardi:
Add an "\_release" directory to the ciao project. The "\_" prefix
tells the go compiler to ignore this directory. This directory
will contain code related to the ciao release process. This commit
adds a "bat" subdirectory under the _release directory with a single
python test script. This test script is documented with python
docstring in the code so that pydoc can be used to extract the
latest documentation. A README file is also included for
convenience, which contains the output of pydoc.

### DIFF
--- a/_release/bat/release_test.py
+++ b/_release/bat/release_test.py
@@ -1,0 +1,475 @@
+#
+# Copyright (c) 2016 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+""" Basic Acceptance Tests for the ciao project
+
+This module contains a set of tests that should be run on a ciao
+cluster prior to submitting a pull request. The tests can be
+run on a physical cluster, or on the ciao single VM setup.
+The output is a TAP (Test Anything Protocol) format file, report.tap.
+These tests utilize the python unittest framework.
+
+Prior to running the tests, the following environment variables
+must be set:
+    "CIAO_IDENTITY" - the URL and port number of your identity service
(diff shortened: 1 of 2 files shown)


_release/bat/release_test.py
```diff
@@ -0,0 +1,351 @@
...
+
+def ciao_admin_env():
+    ciao_env = os.environ.copy()
+    ciao_env["CIAO_USERNAME"] = ciao_env["CIAO_ADMIN_USERNAME"]
+    ciao_env["CIAO_PASSWORD"] = ciao_env["CIAO_ADMIN_PASSWORD"]
+    return ciao_env
+
+# implement a wait loop that waits for all instances to move to "running"
+# but timeout after so many tries.
+def wait_till_running(uuid):
```
mcastelino: @kaccardi We may need a longer delay when running on a virtual enviornment. I have seen boot times > 30s. Can we make the wait configurable

tpepper:
Commit message should describe _release/ is special by way of the leading '_' in order to have its contents ignored by golang tooling.  And a shortlog line of "release: first commit adding BAT" would be more descriptive.  The PR's subject line makes the PR look like much less than what it is.

README.md needed, to describe at a high level what coverage is intended from the BAT.

_release/bat/release_test.py
leoswaldo: @kaccardi @mcastelino  I don't know pretty well the structure for this, but we could add a utils.go and create a function a bit more parametrizable, by creating something like  "wait_for" function where you give the resource, count, and also the status to check.

kaccardi: @mcastelino yes - you are definitely right there. I've already had to change the number of wait loops since I submitted this PR. I'll make it configurable with a higher default value.

@leoswaldo I think thats a good idea. I'll make a change for that.

kaccardi: @leoswaldo I do think your idea is a good one, however, I think waiting until I actually wait for more than one state to implement it is what I'm going to do. I don't want to add anything that's not actually needed yet.

kaccardi:
I can squash all these into a single commit and resubmit if you want.

kaccardi:
this is ready for another review.

tpepper:
FYI there was a question pre-merge about python dependency management, both from a code stability and license compatibility perspective.  The one dep here is taprunner and it is BSD licensed, so no license issue with respect to the BAT script which is Apache.  On the code stability side, consensus is to not vendor the python dependencies at this time.  This is a space we'll need to watch if the BAT test becomes more complex and broad.

PR #271, opened 2016-06-16
Single VM Dev: Fix race condition in image download

mcastelino:
Don't break if/when reading the "latest" file and
https://download.clearlinux.org/image/ races against a publication
of changes to those two directories.

Switch to downloading from the releases location which should
be available


sameo:
Works for me.

Issue #209, opened 2016-06-06
CIAO - API: /v2.1/{tenant_id}/servers  returning wrong format


leoswaldo:
Sending a request with POST to    /v2.1/{tenant_id}/servers   as follows is returning a wrong response format than expected

REQUEST: 

```
REQ: curl -g -i 'https://cloud-controllere.example.com:8774/v2.1/a461044ec60c4498a81b956527ae4f48/servers' -X POST -H "Accept: application/json" -H "Content-Type: application/json" -H "User-Agent: python-novaclient" -H  X-Auth-Project-Id: c_rally_c6d57543_XVtxPVlS" -H "X-Auth-Token: {SHA1}7c71ad614acf1fe7742fbdcfc34c822b20fd5ee8" -d '{"server": {"min_count": 1, "flavorRef": "e35ed972-c46c-4aad-a1e7-ef103ae079a2", "name": "s_rally_c6d57543_n5kK3LJj", "imageRef": "2e7044cc-6db4-43f7-a10c-1307d163eb0c", "max_count": 1}}'
```

RESPONSE

```
RESP: [202] {'Date': 'Mon, 06 Jun 2016 15:45:21 GMT', 'Content-Length': '1041', 'Content-Type': 'application/json'}
[...]

PR #284, opened 2016-06-16
Add compatibility for Compute API /v2.1/{tenant_id}/servers POST

leoswaldo:
Add compatibility for CIAO Compute API for /v2.1/{tenant_id}/servers
for method POST, the calls and results should be in same format as
Nova to be fully compatible with OpenStack.

### DIFF
--- a/ciao-controller/compute.go
+++ b/ciao-controller/compute.go
@@ -883,9 +883,9 @@ func createServer(w http.ResponseWriter, r *http.Request, context *controller) {
 
 	nInstances := 1
 
-	if server.Server.MaxInstances != 0 {
+	if server.Server.MaxInstances > 0 {
 		nInstances = server.Server.MaxInstances
-	} else if server.Server.MinInstances != 0 {
+	} else if server.Server.MinInstances > 0 {
 		nInstances = server.Server.MinInstances
 	}
 
--- a/ciao-controller/compute.go
+++ b/ciao-controller/compute.go
@@ -482,6 +482,11 @@ func showServerDetails(w http.ResponseWriter, r *http.Request, context *controll
 		return
 	}
 
+	// OpenStack compatibility: expect active status not running for API Call
+	if server.Server.Status == "running" {
+		server.Server.Status = "active"
+	}
+
 	b, err := json.Marshal(server)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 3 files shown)


leoswaldo:
Fix for  https://github.com/01org/ciao/issues/209

ciao-controller/compute.go
```diff
@@ -906,7 +911,16 @@ func createServer(w http.ResponseWriter, r *http.Request, context *controller) {
 	}
 	servers.TotalServers = len(instances)
 
-	b, err := json.Marshal(servers)
+	// set machine ID for OpenStack compatibility
+	server.Server.ID = instances[0].ID
```
markdryan: Is it possible for len(instances) to be less than 1?  If so, this code will panic if instances is empty.

leoswaldo: If the code arrives to that line, it would mean at least one instance was created otherwise it would fall into 907 (before)

https://github.com/leoswaldo/ciao/blob/4aa603d12b1238600b0e3b6a27d70b864f13db32/ciao-controller/compute.go#L907

markdryan: If instances is an empty array 907 will not run.  The entire for loop will be skipped and then we'll panic.  Now it could be that context.startWorkload will never return an empty array in which case we're okay.  But if it can return an empty array and a nil error, we'd panic here.

leoswaldo: @markdryan right, startWorkload verify the number of instances is not 0, so that scenario is covered :)  https://github.com/01org/ciao/blob/master/ciao-controller/command.go#L127

markdryan: I'm not sure it is.  The code you link to checks to see whether nInstances is 0 and if it is it returns an error.  But if nInstances was -1, you would get (nil, nil) returned from startWorkload and you would get your panic.  And looking at the code, nInstances could be -1, as is initialised from server.Server.MaxInstances, which comes in via HTTP and so could be anything.

leoswaldo: I see, then the verification should be modified  in  https://github.com/01org/ciao/blob/master/ciao-controller/compute.go#L881  instead of "!= 0" we shoud have  "> 0" and it would prevent the failure

markdryan: Yes, I think that should do it.  I'd change the MinInstances check as well and maybe the check  https://github.com/01org/ciao/blob/master/ciao-controller/command.go#L127 as well.

leoswaldo: Verification committed for negative number of instances commited

markdryan: @leoswaldo Looks good.

kaccardi:
need to coordinate this change with @sameo next week since it will impact the release tests.

ciao-controller/compute.go
```diff
@@ -906,7 +911,16 @@ func createServer(w http.ResponseWriter, r *http.Request, context *controller) {
...
+
+	// builtServers is define to meet OpenStack compatibility on result format and keep CIAOs
+	builtServers := struct {
+		payloads.ComputeCreateServer
+		payloads.ComputeServers
+	}{payloads.ComputeCreateServer{Server: server.Server}, payloads.ComputeServers{TotalServers: servers.TotalServers, Servers: servers.Servers}}
```
sameo: @leoswaldo Could we please get a more readable indentation for this:

```
builtServers := struct {
		payloads.ComputeCreateServer
		payloads.ComputeServers
	}{
		payloads.ComputeCreateServer{
				Server: server.Server
		},
		payloads.ComputeServers{
				TotalServers: servers.TotalServers, 
				Servers: servers.Servers
		}
}
```

leoswaldo: @sameo suggestion applied :)

sameo:
Looks good to me, this is ready to be merged.

PR #294, opened 2016-06-20
Add proper return codes to /v2.1/{tenant}/servers/{server} GET Method…

### LINKED ISSUE #158: Invalid response to /v2.1/​{tenant_id}​/servers/​{server_id}
If a token is not valid, the compute api response with a 500.  The compute api should actually respond with a 401 if the token is not valid, and a 403 if the role is not valid.

leoswaldo:
… API Call

The function showServerDetails attending calls from /v2.1/{tenant}/servers/{server}
GET Method was returning 500 InternalServerError for all error cases, this
fix covers Unauthorized, NotFound and InternalServer return errors for each
corresponding case. Also returning a more meaningful error message.

Fixes #158

### DIFF
--- a/ciao-controller/compute.go
+++ b/ciao-controller/compute.go
@@ -452,6 +452,22 @@ func instanceToServer(context *controller, instance *types.Instance) (payloads.S
 	return server, nil
 }
 
+// returnErrorCode returns error codes for the http call
+func returnErrorCode(w http.ResponseWriter, httpError int, message string) {
+	var returnCode payloads.HTTPReturnErrorCode
+	returnCode.Error.Code = httpError
+	returnCode.Error.Name = http.StatusText(returnCode.Error.Code)
+	returnCode.Error.Message = message
+
+	b, err := json.Marshal(returnCode)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	http.Error(w, string(b), httpError)
+}
+
 func showServerDetails(w http.ResponseWriter, r *http.Request, context *controller) {
 	vars := mux.Vars(r)
 	tenant := vars["tenant"]
@@ -461,24 +477,24 @@ func showServerDetails(w http.ResponseWriter, r *http.Request, context *controll
 	dumpRequest(r)
 
 	if validateToken(context, r) == false {
-		http.Error(w, "Invalid token", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusUnauthorized, "Invalid token")
 		return
 	}
 
 	instance, err := context.ds.GetInstance(instanceID)
 	if err != nil {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusNotFound, "Instance could not be found")
 		return
 	}
 
 	if instance.TenantID != tenant {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusNotFound, "Instance does not belong to tenant")
 		return
 	}
 
 	server.Server, err = instanceToServer(context, instance)
 	if err != nil {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusNotFound, "Instance could not be found")
 		return
 	}
 
--- a/payloads/compute.go
+++ b/payloads/compute.go
@@ -451,3 +451,19 @@ func NewCiaoEvents() (events CiaoEvents) {
 	events.Events = []CiaoEvent{}
 	return
 }
+
+// HTTPErrorData represents the HTTP response body for
+// a compute API request error.
+type HTTPErrorData struct {
+	Code    int    `json:"code"`
+	Name    string `json:"name"`
+	Message string `json:"message"`
+}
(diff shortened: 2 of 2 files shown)


markdryan:
The commit message needs to be updated.

1. The first line should start with the name of the component affected by the commit, e.g., ciao-controller:
2. The lines are too long.  They should wrap at 72 characters.

The code change looks okay, although I wonder whether it would be nicer to use the json package to generate the json for you rather than embedding json docs as strings in the controller code.  Might be safer to do this going forward, if for example, a future error message was generated from one or more variables.

Of course the problem with using the json package is MarshalIndent returns an error so you'd need to deal with that.  The easiest way to do this would be to pre-create all the error strings on controller startup in an init function, panicking if you can an error from MarshallIndent.  I'm not requesting that you make this change however.  This is more a call for the controller maintainers.  I'm just raising the issue.

ciao-controller/compute.go
```diff
@@ -461,24 +461,24 @@ func showServerDetails(w http.ResponseWriter, r *http.Request, context *controll
...
 	if instance.TenantID != tenant {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		http.Error(w, "{\"itemNotFound\": {\"message\": \"Instance not in tenant.\", \"code\": 404}", http.StatusNotFound)
 		return
 	}
 
 	server.Server, err = instanceToServer(context, instance)
 	if err != nil {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		http.Error(w, "{\"itemNotFound\": {\"message\": \"Instance could not be found.\", \"code\": 404}", http.StatusNotFound)
```
sameo: @leoswaldo Please add compute API error payloads to payloads/compute.go and use them to generate JSO payload through marshaling.

ciao-controller/compute.go
```diff
@@ -452,6 +452,23 @@ func instanceToServer(context *controller, instance *types.Instance) (payloads.S
...
+	returnCode.Error.Message = message
+
+	b, err := json.Marshal(returnCode)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+	}
+
+	// convert []byte since htt.Error expects a string data type
+	bMessage := string(b[:])
+	http.Error(w, bMessage, httpError)
```
sameo: `http.Error(w, string(b[:]), httpError)` should be enough.

ciao-controller/compute.go
```diff
@@ -461,24 +478,24 @@ func showServerDetails(w http.ResponseWriter, r *http.Request, context *controll
...
 	instance, err := context.ds.GetInstance(instanceID)
 	if err != nil {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusNotFound, "Instance could not be found")
 		return
 	}
 
 	if instance.TenantID != tenant {
-		http.Error(w, "Instance not available", http.StatusInternalServerError)
+		returnErrorCode(w, http.StatusNotFound, "Instance not in tenant")
```
sameo: I would replace "Instance not in tenant" with "Instance does not belong to tenant".

payloads/compute.go
```diff
@@ -451,3 +451,19 @@ func NewCiaoEvents() (events CiaoEvents) {
 	events.Events = []CiaoEvent{}
 	return
 }
+
+// HTTPReturnCodeData represents the data of a call to be return
+// when explicit data of the call as error messages
```
sameo: Could we call that one HTTPErrorData ?
And the comment should be something along those lines:
```
// HTTPErrorData represents the HTTP response body for
// a compute API request error.
```

sameo:
It looks good to me now.
@tpepper , @markdryan: This one is ready to merge as far as I'm concerned.

ciao-controller/compute.go
```diff
@@ -452,6 +452,21 @@ func instanceToServer(context *controller, instance *types.Instance) (payloads.S
...
+// returnErrorCode returns error codes for the http call
+func returnErrorCode(w http.ResponseWriter, httpError int, message string) {
+	var returnCode payloads.HTTPReturnErrorCode
+	returnCode.Error.Code = httpError
+	returnCode.Error.Name = http.StatusText(returnCode.Error.Code)
+	returnCode.Error.Message = message
+
+	b, err := json.Marshal(returnCode)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
markdryan: Shouldn't there be a return here?  Otherwise we'll send two errors.  Not quite sure what would happen.

ciao-controller/compute.go
```diff
@@ -452,6 +452,21 @@ func instanceToServer(context *controller, instance *types.Instance) (payloads.S
...
+	}
+
+	http.Error(w, string(b[:]), httpError)
```
markdryan: No need to reslice b.  It's already a slice, e.g., string(b) should work fine.

ciao-controller/compute.go
sameo: Correct, good catch.

leoswaldo: @markdryan right , I hadn't seen that, added

markdryan:
Looks good to me.

PR #298, opened 2016-06-21
Unittests: use testutil in ciao-scheduler's testing

tpepper:
This adds to scheduler's test coverage by leveraging the example in the testutil client-server test.  Scheduler coverage goes up from something like 63 to 75% percent.

Additionally some refactoring of testutil/ is done in response to bugs noted when I started using it for scheduler test coverage.

### DIFF
--- a/ciao-controller/compute_test.go
+++ b/ciao-controller/compute_test.go
@@ -234,7 +234,7 @@ func TestDeleteServer(t *testing.T) {
 	time.Sleep(2 * time.Second)
 
 	c := client.AddCmdChan(ssntp.STATS)
-	go client.SendStats()
+	go client.SendStatsCmd()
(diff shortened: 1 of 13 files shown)


tpepper:
There is an intermittent failure in the TestStart case for scheduler which I need to track down.  Happens on travis, but not locally for me....debugging.

tpepper:
Race resolved, ready for merge.

tpepper:
Based on my debug earlier I was pretty sure I'd fixed the race and a first pass through CI seemed to confirm that.  But since the issue wasn't quite 100% reproducible in travis before and rarely reproduced locally, I've just pushed identical code through CI multiple times.  Previously it was failing very often, and now I can't reproduce.  Definitely looks to be fixed.

PR #305, opened 2016-06-22
Remove github.com/docker/distribution/uuid dependency

### LINKED ISSUE #222: Get rid of distribution/uuid
In the interest of reducing our dependencies, distribution/uuid has been identified in Issue #173 as being surplus to requirements.  We just need to implement our own UUID generator.  How hard can it be?

markdryan:
This PR replaces the uuid generator we are currently using with a custom ciao generator, allowing us to remove one dependency.  The old generator basically just generates random 16 byte sequences so we're only reinventing a very small wheel here.

This PR

- Adds a new uuid package to ciao, github.com/01org/ciao/ssntp/uuid complete with its own unit tests
- Updates the travis build to run those unit tests
- Replaces the old uuid generator import statements with imports for the new generator
- unvendors github.com/docker/distribution/uuid

The changes have been tested on the single machine VM setup.

Fixes #222


sameo:
The SSNTP changes look good to me.
The uuid package also looks fine to me as well. I was wondering why @markdryan decided to use the encoding package explictly for BE translation, and the reason is otherwise it would pretty much be a docker uuid package copy.

tpepper:
@markdryan could you add a negative test case for Parse()?

Otherwise lgtm!

markdryan:
@tpepper Sure.  Actually, I forgot to document the existing tests as well.  I'll do this and push a new patch.

PR #310, opened 2016-06-23
ciao-controller: set proper http error return codes

leoswaldo:
This pull requests is to set the proper http error return codes when something goes wrong in the controller.
Here we define the type errors 
- StatusNotFound
- StatusInternalServerError
- StatusUnauthorized
- StatusBadRequest
- StatusServiceUnavailable

A complete reference for net/http module can be found in https://golang.org/pkg/net/http/
This series of commits are split by entities handled by CIAO-CONTROLLER, same functionality but I prefer to keep them separated for each one :

- Servers
- Flavors
- Tenants
- Nodes
- Events
- CNCI
- Trace


tpepper:
Looks good to me

PR #312, opened 2016-06-23
ciao-cli: implement instance show -instance UUID

### LINKED ISSUE #309: Implement ciao-cli instance show [instance UUID]
We should be able to show an instance specific details, not always all of them.

albertomurillo:
Provides a way to list details about an specific
instance

Fixes #309

### DIFF
--- a/ciao-cli/instance.go
+++ b/ciao-cli/instance.go
@@ -40,6 +40,7 @@ var instanceCommand = &command{
 		"add":     new(instanceAddCommand),
 		"delete":  new(instanceDeleteCommand),
 		"list":    new(instanceListCommand),
+		"show":    new(instanceShowCommand),
 		"restart": new(instanceRestartCommand),
 		"stop":    new(instanceStopCommand),
 	},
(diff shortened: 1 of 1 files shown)


ciao-cli/instance.go
```diff
@@ -377,26 +378,75 @@ func (cmd *instanceListCommand) run(args []string) error {
...
 	return nil
 }
 
+type instanceShowCommand struct {
+	Flag     flag.FlagSet
+	instance string
+}
+
+func (cmd *instanceShowCommand) usage(...string) {
+	fmt.Fprintf(os.Stderr, `usage: ciao-cli [options] instance show [flags]
```
sameo: Shouldn't this be:
`ciao-cli [options] instance show [instance uuid]` ?

albertomurillo: well, this is the generic form.
below the flags are specified. but i can change it if you want.

The show flags are:

  -instance string
    	Instance UUID

sameo: I just find it confusing, because `[flags]` sounds like there could be option flags after `instance show` while the only possibility is a UUID.

sameo: @albertomurillo I see...flags are `-instance UUID`. Thanks for the clarification.

Issue #254, opened 2016-06-14
Smart default for certificate paths


sameo:
If SSNTP clients or servers start without providing a CA and private key paths, we should at least look for them in a default directory like e.g. /etc/pki/ciao/.
If there are 2 certificates there, one of them being a CA and the other one a private key we should try using them if they match.

obedmr:
@sameo I think it was addressed at https://github.com/01org/ciao/commit/b82f3cc716e861a0c05d7ddbff805775b3b59f2c, right?

sameo:
Closed with b82f3cc.

PR #316, opened 2016-06-23
ssntp: Add smart default certificates finding

obedmr:
This change adds value to our Zero Configuration approach in CIAO
project. It sets ``--cacert`` and ``--cert`` as optional parameters.

For example:
```
  $ ciao-launcher --network=<Role>
```
It will search for ``CAcert*`` and ``cert-<Role>*`` certificate files at
``/etc/pki/ciao/`` path, if it finds them, it will try to use them.

Fix to Issue #254 

### DIFF
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -95,8 +95,8 @@ var simulate bool
 var maxInstances = int(math.MaxInt32)
 
 func init() {
-	flag.StringVar(&serverCertPath, "cacert", "/etc/pki/ciao/CAcert-server-localhost.pem", "Client certificate")
-	flag.StringVar(&clientCertPath, "cert", "/etc/pki/ciao/cert-client-localhost.pem", "CA certificate")
+	flag.StringVar(&serverCertPath, "cacert", "", "Client certificate")
+	flag.StringVar(&clientCertPath, "cert", "", "CA certificate")
 	flag.Var(&networking, "network", "Can be none, cn (compute node) or nn (network node)")
 	flag.BoolVar(&hardReset, "hard-reset", false, "Kill and delete all instances, reset networking and exit")
 	flag.BoolVar(&simulate, "simulation", false, "Launcher simulation")
--- a/ssntp/ssntp.go
+++ b/ssntp/ssntp.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
@@ -710,6 +711,9 @@ const defaultServerCert = "/etc/pki/ciao/cert-Server-localhost.pem"
 const defaultClientCert = "/etc/pki/ciao/client.pem"
 const defaultSchedulerCert = "/etc/pki/ciao/cert-Scheduler-localhost.pem"
 
+// Default CIAO certs path
+const ciaoCertsPath = "etc/pki/ciao/"
+
 // RoleToDefaultCertName returns default certificate names for each SSNTP role
 func RoleToDefaultCertName(role Role) string {
 	switch role {
@@ -1151,13 +1155,75 @@ func (config *Config) port() uint32 {
 	return port
 }
 
-func (config *Config) setCerts() {
-	if config.CAcert == "" {
-		config.CAcert = DefaultCACert
+func loadCertificate(certPath string) (*x509.Certificate, error) {
+	certPEM, err := ioutil.ReadFile(certPath)
+	if err != nil {
+		return nil, err
+	}
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		return nil, fmt.Errorf("Failed to parse certificate PEM")
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse certificate: %v", err)
+	}
+	return cert, nil
+}
+
+func getDefaultCertificate() (cacert, cert string, err error) {
+	certs := []string{}
+
+	files, err := filepath.Glob(ciaoCertsPath)
+	if err != nil {
+		return "", "", err
 	}
 
-	if config.Cert == "" {
-		config.Cert = defaultClientCert
+certsLoop:
+	for _, file := range files {
+		if cacert == "" {
+			cert, err := loadCertificate(file)
+			if err != nil {
+				return "", "", err
+			}
+			if cert.IsCA == true {
+				cacert = file
+				continue certsLoop
+			}
+		}
+		certs = append(certs, file)
+	}
+
+	if len(certs) > 1 {
+		return "", "", fmt.Errorf("More than one cert files at: %s", ciaoCertsPath)
+	} else if len(certs) == 0 {
+		return "", "", fmt.Errorf("%s Certificates are not found", ciaoCertsPath)
+	}
+
+	certPEM, err := ioutil.ReadFile(cacert)
+	certPool := x509.NewCertPool()
+	certPool.AppendCertsFromPEM(certPEM)
+	vOpts := x509.VerifyOptions{Roots: certPool}
+
+	clientCert, err := loadCertificate(certs[0])
+	if err != nil {
+		return "", "", err
+	}
+	_, err = clientCert.Verify(vOpts)
+	if err != nil {
+		return "", "", err
+	}
+
+	return cacert, certs[0], nil
+}
+
+func (config *Config) setCerts() {
+	var err error
+	if config.CAcert == "" {
+		config.CAcert, config.Cert, err = getDefaultCertificate()
+		if err != nil {
+			log.Fatal(err)
+		}
 	}
 }
 


sameo:
@obedmr I actually think this should be part of the ssntp package. From `ssntp.go:setCerts()` for example. If the config certs are empty, then we try to find some under `/etc/pki/ciao` and use them if there is only one CA and one private key there.

ciao-launcher/main.go
```diff
@@ -341,6 +343,54 @@ func printClusterConfig() {
 	glog.Infof("Memory Limit:         %v", memLimit)
 }
 
+func discoverCertificate(certPath, certPrefix string) (string, error) {
+	ciaoCertsPath := "/etc/pki/ciao/"
+	fmt.Println("discoverCertificate")
```
mcastelino: did you mean to log here? we typically do not perform direct printing. you always use --alsologtostderr

mcastelino:
@obedmr it would be good to take role into account here where doing auto discovery.
@sameo it becomes a chicken and a egg sort of. certs have roles. if you auto discover certs then if the user has provided the role, it may make sense to pick the cert with the right role. It makes test automation easier if you want to have multiple certs in the /etc/pki/ciao and if the use provides roles use the cert with the role

ciao-launcher/main.go
```diff
@@ -341,6 +343,54 @@ func printClusterConfig() {
...
+	if _, err := os.Stat(certPath); os.IsNotExist(err) {
+		files, err := ioutil.ReadDir(ciaoCertsPath)
+		if err != nil {
+			return certPath, err
```
mcastelino: may make sense to return "" string here vs the certpath in case of error... I know user should always check error but sending back a valid string in the case of error may not make sense

sameo:
@mcastelino We removed the Role field from the SSNTP config structure, so there is no longer a way to provide that information.
The goal of this is really to be slightly smarter when we don't get certs with the SSNTP config. When that happens we try our luck and poke under /etc/pki/ciao/ to see if there's a (CA, key) pair. If there is we just try to use it. If not we fail.

ciao-launcher/main.go
sameo: Agreed, thanks @mcastelino.

mcastelino:
@sameo I had forgotten about that. In that case this functionality belongs in SSNTP as you said. Else it will get duplicated in all agents...

ciao-launcher/main.go
obedmr: @mcastelino  good catch there, I forgot to remove that line.

ciao-launcher/main.go
obedmr: @sameo @mcastelino it wouldn't return "", it will return the default one that is specified at: https://github.com/01org/ciao/blob/master/ciao-launcher/main.go#L98-99.

This error's return will happen when the ``/etc/pki/ciao`` dir doesn't exist.

obedmr:
@sameo, yeah, agree with that ssntp.go approach, it would be better and would apply for all clients (Launcher, Controller, etc), /me updating it.

mcastelino:
@obedmr maybe I am missing something. If the user explicitly specifies the certificate, will that take precendence over auto discovery. The user should be able to place the certificate anywhere in any directory and use the command line to point to it.

obedmr:
@mcastelino yes, it will work like that, if the user specifies the certificate (anywhere) it will take as option 1. I'm updating my patch with changes that @sameo  pointed. I will send it updated.

obedmr:
@sameo I will continue with the ssntp_test.go issue. There are more than one ``cert-*`` in the test environment, I'll need to add more detail in the prefix I'm searching for.

ssntp/ssntp.go
```diff
@@ -1151,14 +1152,41 @@ func (config *Config) port() uint32 {
 	return port
 }
 
+func discoverCertificate(certPrefix string) (string, error) {
+	ciaoCertsPath := "/etc/pki/ciao/"
```
sameo: @obedmr Could you please make that one a ssntp const, along with the rest of the default certificate names:
`const defaultCertPath = "/etc/pki/ciao/"`

ssntp/ssntp.go
```diff
@@ -1151,14 +1152,41 @@ func (config *Config) port() uint32 {
...
+	certPath := strings.Join([]string{ciaoCertsPath, certPrefix, "*"}, "")
+
+	files, err := filepath.Glob(certPath)
+	if err != nil {
+		fmt.Println(err)
```
markdryan: Should this be a log?

ssntp/ssntp.go
```diff
@@ -1151,14 +1152,41 @@ func (config *Config) port() uint32 {
 	return port
 }
 
+func discoverCertificate(certPrefix string) (string, error) {
+	ciaoCertsPath := "/etc/pki/ciao/"
+	certPath := strings.Join([]string{ciaoCertsPath, certPrefix, "*"}, "")
+
```
markdryan: certPath := path.Join(cioaCertsPath, certPrefix+"*")

might be better here.

ssntp/ssntp.go
```diff
@@ -1151,14 +1152,41 @@ func (config *Config) port() uint32 {
 	return port
 }
 
+func discoverCertificate(certPrefix string) (string, error) {
```
sameo: So here we should avoid relying on specific certificate prefixes and names. Right now they're consistent because they're all generated by ciao-cert, but in the future we can't make that a hard requirement.
So what I'd prefer to see is a `getDefaultCertificate() (string, string, error)` function that would return both certificate paths when found. It would essentially do what `discoverCertificate()` does plus the following:
1. Actually open and read each of the 2 PEM files found
2. Verify that 1 is a CA and the other is not.
If that is the case, we return the 2 paths found.
Eventually we will also do:
3. Verify the non CA cert with the CA one

After building an io reader from a certificate file and getting a byte slice from it (`certPEM`), you have to do the following for implementing step 2:
```
	block, _ := pem.Decode(certPEM)
	if block == nil {
		panic("failed to parse certificate PEM")
	}
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		panic("failed to parse certificate: " + err.Error())
	}
```

Then you have a `x509.Certificate` structure that represents a decoded PEM block. Testing if it's a CA or not is verifying: `cert.IsCA == true`.

ssntp/ssntp.go
```diff
@@ -1151,13 +1155,76 @@ func (config *Config) port() uint32 {
...
-	if config.CAcert == "" {
-		config.CAcert = DefaultCACert
+func loadCertificate(certPath string) (*x509.Certificate, error) {
+	certPEM, err := ioutil.ReadFile(certPath)
+	if err != nil {
+		return nil, err
+	}
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		panic("failed to parse certificate PEM")
```
markdryan: Why panic here?

ssntp/ssntp.go
```diff
@@ -1151,13 +1155,76 @@ func (config *Config) port() uint32 {
...
-func (config *Config) setCerts() {
-	if config.CAcert == "" {
-		config.CAcert = DefaultCACert
+func loadCertificate(certPath string) (*x509.Certificate, error) {
+	certPEM, err := ioutil.ReadFile(certPath)
+	if err != nil {
+		return nil, err
+	}
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
```
markdryan: Might as well capture and log the error.

ssntp/ssntp.go
```diff
@@ -1151,13 +1155,76 @@ func (config *Config) port() uint32 {
...
+	}
+	return cert, nil
+}
+
+func getDefaultCertificate() (cacert, cert string, err error) {
+	certs := []string{}
+
+	files, err := filepath.Glob(ciaoCertsPath)
+	if err != nil {
+		fmt.Println(err)
```
markdryan: Should this be a log?

ssntp/ssntp.go
```diff
@@ -1151,13 +1155,76 @@ func (config *Config) port() uint32 {
...
+				continue certsLoop
+			}
+		}
+		certs = append(certs, file)
+	}
+
+	if len(certs) > 1 {
+		return "", "", fmt.Errorf("More than one cert files at: %s", ciaoCertsPath)
+	} else if len(certs) == 1 {
+		certPEM, err := ioutil.ReadFile(cacert)
```
markdryan: The code in this block represents the success condition and so should be on the main path through the function, e.g.,

```
} else if len(certs) == 0 {
    return "", "", fmt.Errorf("%s Certificates are not found", ciaoCertsPath)
}

certPEM, err := ioutil.ReadFile(cacert)
...

ssntp/ssntp.go
obedmr: You're right. Added a err return for it.

sameo:
@obedmr Could you also please change the commit title ? It should start with `ssntp:`, and not `ciao-launcher:` anymore.

PR #322, opened 2016-06-28
ssntp: Add smart default certificates finding

obedmr:
This change adds value to our Zero Configuration approach in CIAO
project. It sets ``--cacert`` and ``--cert`` as optional parameters.

Examples:
- When starting ciao-launcher:
```bash
     # ciao-launcher
```
- When starting ciao-controller:
```bash
    # ciao-controller
```

That's all, no more args. SSNTP certificates are auto-discovered
from ``/etc/pki/ciao`` path. It's important to mention that, if we
have more than 1 SSNTP certificates, it will fail.

### DIFF
--- a/ssntp/ssntp.go
+++ b/ssntp/ssntp.go
@@ -27,6 +27,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
@@ -710,6 +711,9 @@ const defaultServerCert = "/etc/pki/ciao/cert-Server-localhost.pem"
 const defaultClientCert = "/etc/pki/ciao/client.pem"
 const defaultSchedulerCert = "/etc/pki/ciao/cert-Scheduler-localhost.pem"
 
+// Default CIAO certs path
+const ciaoCertsPath = "/etc/pki/ciao/*"
+
 // RoleToDefaultCertName returns default certificate names for each SSNTP role
 func RoleToDefaultCertName(role Role) string {
 	switch role {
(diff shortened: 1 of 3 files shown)


ssntp/ssntp.go
```diff
@@ -710,6 +711,9 @@ const defaultServerCert = "/etc/pki/ciao/cert-Server-localhost.pem"
 const defaultClientCert = "/etc/pki/ciao/client.pem"
 const defaultSchedulerCert = "/etc/pki/ciao/cert-Scheduler-localhost.pem"
 
+// Default CIAO certs path
+const ciaoCertsPath = "etc/pki/ciao/"
```
sameo: Why is this not an absolute path: `const ciaoCertsPath = "/etc/pki/ciao/" ?

obedmr: @sameo,  It was my bad, it is an absolute path ``/etc/pki/ciao`` , code updated.

obedmr:
@sameo what do you think about this code, https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L704-711

It's only being used by ``controller``
https://github.com/01org/ciao/blob/master/ssntp/ssntp.go#L704-711

Also, by the ``testutil``
https://github.com/01org/ciao/blob/master/testutil/agent.go#L90
https://github.com/01org/ciao/blob/master/testutil/controller.go#L67

For the `` controller`` case, we could follow the same approach for smart certs finding , it would mean that we need to set defaults to empty strings instead of those provided ones at :
https://github.com/01org/ciao/blob/master/ciao-controller/main.go#L38-39  

What do you think?

mcastelino:
@obedmr @sameo given that role is no longer a valid option. Better if we do not have it in the PR message.

obedmr:
@mcastelino yeah, good point, /me updating commit message.

obedmr:
@sameo I think, I can answer my own question, I'm seeing this other controller-only certificates 
https://github.com/01org/ciao/blob/master/ciao-controller/main.go#L114-115

The way it(smart-certs) is implemented may cause a conflict if they are in the same location of the CACert and cert.

sameo:
@obedmr You're right, if we have the HTTPS and the SSNTP certs under /etc/pki/ciao/, we will fail. That should not happen for compute nodes, but only for controller nodes.
So one easy solution is for the controller to continue using the cert and HTTP command line options. That's the easy one and that won't be broken with this PR.
But we can be smarter. What we're looking for are SSNTP certificates, and they happen to be very specific as they contain specific extended key usage attributes for describing the certificate role. So our smart search could really look for that (the go x509 package makes that very easy to implement) and would only fail when it finds more than 1 SSNTP cert.

obedmr:
@sameo, good point, let me work on that smarter feature for controller. You're right, that role property will save us. :)

obedmr:
@sameo, it's updated to auto discover controller and launcher SSNTP certs.

sameo:
Superseded by [https://github.com/01org/ciao/pull/332](PR 332)

PR #326, opened 2016-06-29
Fix instance status strings inconsistency

sameo:
- Align the STATS instance status strings with the compute API ones
- Only use the payloads defined constants for manipulating instance statuses
- Redefine ComputeStatusRunning to "active"

### DIFF
--- a/ciao-controller/internal/datastore/sqlite3db.go
+++ b/ciao-controller/internal/datastore/sqlite3db.go
@@ -1196,7 +1196,7 @@ func (ds *sqliteDB) getInstances() ([]*types.Instance, error) {
 	)
 	SELECT	instances.id,
 		instances.tenant_id,
-		IFNULL(latest.state, "pending") AS state,
+		IFNULL(latest.state, "` + payloads.ComputeStatusPending + `") AS state,
 		workload_id,
 		IFNULL(latest.ssh_ip, "Not Assigned") as ssh_ip,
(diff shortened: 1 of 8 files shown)


tpepper:
Thank you for doing this cleanup.  A big +2 for maintenance!

ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -1196,7 +1196,7 @@ func (ds *sqliteDB) getInstances() ([]*types.Instance, error) {
 	)
 	SELECT	instances.id,
 		instances.tenant_id,
-		IFNULL(latest.state, "pending") AS state,
+		IFNULL(latest.state, "` + payloads.ComputeStatusPending + `") AS state,
```
mcastelino: There are a couple of other references. Should they also be replaced.

total_running AS
(
	SELECT	instances.node_id AS node_id,
		count(instances.instance_id) AS total
	FROM instances
	WHERE state='running'
	GROUP BY node_id
),
total_pending AS
(
	SELECT	instances.node_id AS node_id,
		count(instances.instance_id) AS total
	FROM instances
	WHERE state='` + payloads.ComputeStatusPending + `'
	GROUP BY node_id
),
total_exited AS
(
	SELECT	instances.node_id,
		count(instances.instance_id) AS total
	FROM instances
	WHERE state='exited'
	GROUP BY node_id
)

PR #331, opened 2016-06-29
ciao-cli: Add image modification support via the modify subcommand

erick0z:
- Add the support of modifying the following image properties:
    - visibility
    - name
    - tags

*The visibility modification depends on this PR https://github.com/sameo/gophercloud/pull/3 in order to work.

### DIFF
--- a/ciao-cli/image.go
+++ b/ciao-cli/image.go
@@ -72,6 +72,7 @@ var imageCommand = &command{
 		"upload":   new(imageUploadCommand),
 		"download": new(imageDownloadCommand),
 		"delete":   new(imageDeleteCommand),
+		"modify":   new(imageModifyCommand),
 	},
 }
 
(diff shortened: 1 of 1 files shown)


ciao-cli/image.go
```diff
@@ -350,6 +351,87 @@ func uploadTenantImage(username, password, tenant, image, filename string) error
 	return res.Err
 }
 
+type imageModifyCommand struct {
+	Flag       flag.FlagSet
+	name       string
+	imageID    string
```
albertomurillo: Other commands such as "image upload" have an "image" argument/variable
you might want to change this for consistency

ciao-cli/image.go
```diff
@@ -350,6 +351,87 @@ func uploadTenantImage(username, password, tenant, image, filename string) error
...
+	visibility string
+	tags       string
+}
+
+func (cmd *imageModifyCommand) usage(...string) {
+	fmt.Fprintf(os.Stderr, `usage: ciao-cli [options] image set [flags] imageID
```
albertomurillo: Maybe remove imageID from the end since the help will print:

```
The set flags are:

  -image string
        Image UUID
```

ciao-cli/image.go
```diff
@@ -350,6 +351,87 @@ func uploadTenantImage(username, password, tenant, image, filename string) error
...
+
+Modify an image
+
+The modify flags are:
```
albertomurillo: You are calling the command 'set' on line 75 so it should be named set here too.
Or name it modify if it makes more sense

PR #337, opened 2016-07-04
ciao-launcher: Update the -network parameter

### LINKED ISSUE #162: Update the ciao-launcher's --network parameter
Since commit 8502826b3441e151af6f1bc0f8ef91580c187074 the only way for launcher to specify it's SSNTP role is via a certificate.  This means that the --network parameter needs to switch to be a boolean, indicating that networking is either enabled or disabled.  The default should probably switch to true rather than false.  The reason for the default being false is historical.

In addition to the code, we'll need to update the documentation and any scripts that run ciao-launcher.

markdryan:
-network is now a bool parameter and it defaults to true.  Previously, it
was an emumerated type that accepted none, cn, nn, or dual.  But these
additional values are no longer needed as SSNTP can deduce the role
from its certificates.  Another important note is that -network defaults
to true.  Previously, if the networking flag was unspecified, networking
was disabled.  This default behaviour was left over from the early days
of ciao when we were integrating networking.  Networking on by default
is now the sensible choice.

Fixes #162

### DIFF
--- a/ciao-launcher/README.md
+++ b/ciao-launcher/README.md
@@ -72,7 +72,7 @@ Currently, this is the only data source supported by launcher.
 
 ciao-launcher can be launched from the command line as follows
 ```
-sudo ciao-launcher -network [cn|nn|none]
+sudo ciao-launcher
 ```
 
 Currently, launcher needs to be run as root so that it can create network links and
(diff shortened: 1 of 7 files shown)


markdryan:
SingleVM works but haven't tested successfully on a cluster yet.  Best to wait until I do this before merging.

PR #338, opened 2016-07-05
release: add release notes generator to ciao tree

kaccardi:
ciao-release is a small program which will generate the release
notes for ciao releases. It uses the github api to obtain
the list of pull requests closed on the master branch since
the time of the last tag, up to the release candidate sha. The rc
sha must be passed in as a string argument. It also gets the list of issue
events since the last tag and reports any closed issues.
Finally, it gets the list of all commits since the last tag.
The lists are output into a plain text file that is called
"release.txt". This file is used to create an annotated tag
for the next release.

### DIFF
--- a/_release/ciao-release/main.go
+++ b/_release/ciao-release/main.go
@@ -1,0 +1,213 @@
+//
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/golang/glog"
+	"github.com/google/go-github/github"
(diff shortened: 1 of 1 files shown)


_release/ciao-release/main.go
```diff
@@ -0,0 +1,213 @@
...
+
+func main() {
+	flag.Parse()
+
+	f, err := os.Create("release.txt")
+	if err != nil {
+		glog.Error(err)
+		os.Exit(1)
+	}
+	defer f.Close()
```
markdryan: f will not be closed if os.Exit(1) is called, even though you've deferred it.

_release/ciao-release/main.go
```diff
@@ -0,0 +1,213 @@
...
+
+	if len(commits) > 0 {
+		fmt.Fprintln(f, "---Full Change Log---")
+	}
+
+	for _, c := range commits {
+		lines := strings.Split(*c.Commit.Message, "\n")
+		fmt.Fprintf(f, "\t%s\n", lines[0])
+	}
+}
```
markdryan: Does goreportcard descend into _ directories?  If so it might complain about this function.  I haven't tested it, but it looks like it might have a gocyclo store of over 15.

_release/ciao-release/main.go
kaccardi: I refactored to do this correctly, thanks.

_release/ciao-release/main.go
kaccardi: I decided to stop being lazy and refactor this to make the code easier to maintain. So, hopefully the next iteration is better - I did check that the gocyclo score was less than 15 for all functions.

markdryan:
Looks good to me.

PR #339, opened 2016-07-06
[Don't merge] ciao-launcher: Limit the number of parallel starts

### LINKED ISSUE #8: Limit number of parallel starts
Currently, there is no limit to the number of instances that ciao-launcher will start in parallel.  This places extreme load on the compute node when spawning large amounts of instances at once and can lead to various errors.  It may be better for launcher to introduce some sort of semaphore to limit the number of instances that can be launched in parallel to some function of the number of cores on the machine.  We could also return a special STATUS, e.g., throttle, to indicate that launcher is overloaded but not full.

When launching large amounts of instances, e.g., 10000, we often see some failures and timeouts in qemu and networking.  Reducing the load on the compute node may prevent these failures.

markdryan:
This commit limits the number of parallel starts to a function
of the number of CPUs present in the node.  There really isn't
much point in allowing 1000 instances to be started on the same
node at the same time.  Doing so won't increase the start times
much and will increase the likelihood of failure due to the
resource exhaustion caused by the heavy demands of instance
startup.

Fixes #8

### DIFF
--- a/ciao-launcher/instance.go
+++ b/ciao-launcher/instance.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"path"
+	"runtime"
 	"sync"
 	"time"
 
@@ -61,6 +62,12 @@ type insDeleteCmd struct {
 type insStopCmd struct{}
 type insMonitorCmd struct{}
 
+var startSemaphore chan struct{}
+
+func init() {
+	startSemaphore = make(chan struct{}, runtime.NumCPU()*2)
+}
+
 /*
 This functions asks the server loop to kill the instance.  An instance
 needs to request that the server loop kill it if Start fails completly.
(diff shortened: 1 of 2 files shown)


markdryan:
@tpepper I think *20 will be too large.   We might not hit this limit on our NUC clusters.  The reason I chose the *2 value as it roughly matches the size of the semaphore in networking.  It's not clear to me whether increasing this value will actually improve start up times.  Let me do some testing on a bigger cluster.  The current patch seemed to speed things up on a small NUC cluster.

markdryan:
Let's not merge this right now.  I'll do a bit more testing first.

markdryan:
So after some experimentation with two different clusters I see different results.  On one cluster limiting the amount of parallel starts to 8 or 12 improves start time by about 50%.  On a second cluster limiting the amount of parallel starts to such a small number seems to increase start time by about 15%.  I don't yet have a good explanation for why this is although I suspect it might be related to the storage used on the different clusters.  Right now I don't have a good formula for limiting parallel starts that will work well on all clusters.  We might need to do as @tpepper suggested and make this a cluster configurable item.  

I'll do some more profiling to get a better understanding of what is going on.  I've already done a CPU profile and there's no great difference between the profiles generated on both cluster.  Launcher is spending most of its time gathering statistics.  Next I'll try to get the blocking profiler to work to see whether that sheds any light.

Closing this PR for now.  Will re-open later when I have a fix that works on all clusters.

PR #348, opened 2016-07-12
ciao-launcher: Update statistics interval and fix profiling

markdryan:
This PR contains two commits that

1. Update the statistics reporting interval to 6 seconds, the value used in the demo branch.
2. Fixes the cpuprofiling in launcher by ensuring that pprof.StopCPUProfile is called.


kaccardi:
looks good to me.

PR #350, opened 2016-07-14
Unittests: race fixes

tpepper:
This PR is a partial response to issue 83.  That issue included 5 data race reports from go-race.  2 no longer reproduce with my earlier unit test cleaning.  2 are fixed here.  1 remains in discussion with @kaccardi 

The two races fixed here are fairly straight forward:  an anonymous function goroutine was sharing unprotected data with its parent in an actually safe fashion imho but easily resolved with a channel, and ssntp clients usage of waitgroups could theoretically race due to no exclusion of readers/writers.

### DIFF
--- a/ciao-controller/controller_test.go
+++ b/ciao-controller/controller_test.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"sync"
 	"testing"
 	"time"
 
@@ -732,14 +731,10 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
 	newTenant := uuid.Generate().String() // random ~= new tenant and thus triggers start of a CNCI
 
 	// trigger the START command flow, and await results
-	// NOTE: "instances" is shared with the go subroutine and we must
-	// insure consistency between parent and child processes
-	var instances []*types.Instance
+	instanceCh := make(chan []*types.Instance)
 
-	var wg sync.WaitGroup
-	wg.Add(1)
 	go func() {
-		instances, err = context.startWorkload(wls[0].ID, newTenant, 1, false, "")
+		instances, err := context.startWorkload(wls[0].ID, newTenant, 1, false, "")
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -747,7 +742,8 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
 		if len(instances) != 1 {
 			t.Fatalf("Wrong number of instances, expected 1, got %d", len(instances))
 		}
-		wg.Done()
+
+		instanceCh <- instances
 	}()
 
 	_, err = netClient.GetCmdChanResult(netClientCmdCh, ssntp.START)
@@ -794,7 +790,7 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
 		t.Fatalf("Did not get correct Instance ID, got %s, expected %s", result.InstanceUUID, tenantCNCI[0].InstanceID)
 	}
 
-	wg.Wait() // for "instances"
+	instances := <-instanceCh
 	if instances == nil {
 		t.Fatal("did not receive instance")
 	}
--- a/ssntp/client.go
+++ b/ssntp/client.go
@@ -130,7 +130,15 @@ func (client *Client) handleSSNTPServer() {
 				break
 			}
 
+			client.status.Lock()
+			if client.status.status == ssntpClosed {
+				client.status.Unlock()
+				return
+			}
+			//insure new frame doesn't race with client.Close()
 			client.frameWg.Add(1)
+			client.status.Unlock()
+
 			go client.processSSNTPFrame(&frame)
 		}
 


ciao-controller/controller_test.go
```diff
@@ -732,22 +731,21 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
...
-	// NOTE: "instances" is shared with the go subroutine and we must
-	// insure consistency between parent and child processes
-	var instances []*types.Instance
+	instanceCh := make(chan []*types.Instance, 1)
 
-	var wg sync.WaitGroup
-	wg.Add(1)
 	go func() {
-		instances, err = context.startWorkload(wls[0].ID, newTenant, 1, false, "")
+		var instances []*types.Instance
```
markdryan: I guess you don't need this line anymore.

ciao-controller/controller_test.go
```diff
@@ -732,22 +731,21 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
...
 		if err != nil {
 			t.Fatal(err)
 		}
 
 		if len(instances) != 1 {
 			t.Fatalf("Wrong number of instances, expected 1, got %d", len(instances))
 		}
-		wg.Done()
+
+		instanceCh <- instances
```
markdryan: Does the race detector identify instances as the problem?Could something else be racy instead, e.g., context.ds?  The TenantCNCI.Subnets slice maybe?    The original code seems to protect instances as you already mention, although this code is a bit nicer I think.

ssntp/client.go
```diff
@@ -77,6 +77,7 @@ type Client struct {
 	closed    chan struct{}
 
 	frameWg              sync.WaitGroup
+	frameWgMutex         sync.Mutex
```
markdryan: This one is odd as well.  You shouldn't need to protect a WaitGroup with a mutex.  By design they are expected to be called safely from different go routines.   Could you paste the race detector logs?

ciao-controller/controller_test.go
tpepper: Ah true. Fixed

ciao-controller/controller_test.go
tpepper: It does not explicitly say it was instances.  But it pointed at code lines that iterate across the instances.

ssntp/client.go
tpepper: === RUN   TestDeleteInstance                                                    
==================                                                              
WARNING: DATA RACE                                                              
Write by goroutine 116:                                                         
  internal/race.Write()                                                         
      /usr/src/golang-1.6/src/internal/race/race.go:41 +0x2e                    
  sync.(*WaitGroup).Wait()                                                      
      /usr/src/golang-1.6/src/sync/waitgroup.go:125 +0xf9                       
  github.com/01org/ciao/ssntp.(*Client).Close.func1()                           
      /usr/src/go/src/github.com/01org/ciao/ssntp/client.go:359 +0x3d           
                                                                                
Previous read by goroutine 33:                                                  
  internal/race.Read()                                                          
      /usr/src/golang-1.6/src/internal/race/race.go:37 +0x2e                    
  sync.(*WaitGroup).Add()                                                       
      /usr/src/golang-1.6/src/sync/waitgroup.go:67 +0xfa                        
  github.com/01org/ciao/ssntp.(*Client).handleSSNTPServer()                     
      /usr/src/go/src/github.com/01org/ciao/ssntp/client.go:133 +0x5c3          
                                                                                
Goroutine 116 (running) created at:                                             
  github.com/01org/ciao/ssntp.(*Client).Close()                                 
      /usr/src/go/src/github.com/01org/ciao/ssntp/client.go:361 +0x25d          
  github.com/01org/ciao/ciao-controller.TestDeleteInstance()                    
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:317 +0x50b
  testing.tRunner()                                                             
      /usr/src/golang-1.6/src/testing/testing.go:473 +0xdc                      
                                                                                
Goroutine 33 (finished) created at:                                             
  github.com/01org/ciao/ssntp.(*Client).Dial()                                  
      /usr/src/go/src/github.com/01org/ciao/ssntp/client.go:334 +0xd34          
  github.com/01org/ciao/testutil.NewSsntpTestClientConnection()                 
      /usr/src/go/src/github.com/01org/ciao/testutil/agent.go:95 +0xacb         
  github.com/01org/ciao/ciao-controller.testStartWorkload()                     
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:663 +0x1cb
  github.com/01org/ciao/ciao-controller.TestDeleteInstance()                    
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:296 +0x90
  testing.tRunner()                                                             
      /usr/src/golang-1.6/src/testing/testing.go:473 +0xdc                      
==================                                                              
--- PASS: TestDeleteInstance (1.10s)

tpepper: https://golang.org/pkg/sync/#WaitGroup talks about the "main" goroutine and I've only ever seen examples in that way where a parent creates the waitgroup and awaits children.

In the case of a https://golang.org/pkg/sync/#Cond waiting there is noted to be a lock involved internally, but not with a waitgroup.  Perhaps it is there and opaque, but this go-race logs implies it is not.

This is perhaps a usage where a channel is more appropriate actually instead of a waitgroup.  @sameo what do you think of changing frameWg to a channel?

tpepper: I might be misunderstanding the code, but a channel doesn't look like it would work here.  It's possible that an arbitrary number of client.frameWg.Add(1) calls are made, each of which will have a corresponding client.frameWg.Done().  But the one waiter in Close() doesn't explicitly know how many to waitgroup Add(1)'s happened and how many to await.  I recall now in our Golang class being told you shouldn't do Add(1) on wait groups as it's a sign of a design flaw.  Instead you want to know how many children you have do an Add(N) once, and then one wait.

I think this points at the actual issue here.

If Close() hits its client.frameWg.Wait() _just_ ahead of one of the arbitrary number of handleSSNTPServer() runs doing a client.frameWg.Add(1), then the Close() will wait for one less frame handling completion than it should and it's possible the client.frameRoutinesChannel is closed ahead of its writers.  If that's the issue, my introduction of a lock doesn't truly address that without changing the lock/unlock placement.

In this case I think ssntpStatus needs an ssntpClosing ahead of ssntpClosed.  We'd enter that state in the Close() and then client.frameWg.Wait().  In the handleSSNTPServer() we'd want to bail out if client.status==ssntpClosing rather than client.frameWg.Add(1) and go client.processSSNTPFrame(&frame).

Or..there is already the closed state checked and set in Close().  Perhaps it just needs checked in the non-error path of the read returning inside handleSSNTPServer().

I'll test a couple options here and refactor.

markdryan:
@tpepper It looks like the problem is that ssntp is calling wg.Add and wg.Wait concurrently in two different go routines which you shouldn't do.  This isn't explicitly documented but there's some code in the implementation of Add that checks for this and panics if detects such a thing.  Normally, this isn't a problem as Add and Wait are called from the same go routine but this is not the case here.  In the short term your mutex actually looks correct, although there must be a cleaner way of handling this inside ssntp.

ciao-controller/controller_test.go
```diff
@@ -732,22 +731,21 @@ func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClie
 	newTenant := uuid.Generate().String() // random ~= new tenant and thus triggers start of a CNCI
 
 	// trigger the START command flow, and await results
-	// NOTE: "instances" is shared with the go subroutine and we must
-	// insure consistency between parent and child processes
-	var instances []*types.Instance
+	instanceCh := make(chan []*types.Instance, 1)
```
tpepper: I've inadvertantly created a buffered channel here.

instanceCh := make(chan []*types.Instance, 1)
is not
instanceCh := make(chan []*types.Instance, 0)
which is
instanceCh := make(chan []*types.Instance)

markdryan:
But as you point out the locks might fix the race but might not lead to the correct behaviour.

tpepper:
v3 of patch addresses a tiny remaining window of race that @markdryan noted.

The general race here is when the frame handler gets a new frame and a Close() is happening.  The Close() needs to wait on all in-flight frames to complete so the frame channel is not close ahead of the handler writing it.  After v1 and v2 there was still a gap in the frame handler where it could have determined the client was not in ssntpClosed status, and release the status lock. The Close() function could then set ssntpClosed and call Wait().  The frame handler could then call Add().  At which point the Close() might continue to interleave such that it finishes with a close of the frame channel and later the frame handler writes the channel for a panic.

I move the frame handler's Add() to be within critical section of holding the status lock.  This insures no Add's happen after the transition to ssntpClosed and Wait() call.

tpepper:
Re: the one remaining race issue @kaccardi noted on IRC: "The instances map in datastore used to have mutexes protecting instances from races - we used to be go race clean...I'll have to look at the code in question - but something could have snuck in."

markdryan:
Commit  7cf32c8ac7913ae looks good to me.  As mentioned on IRC the ideal fix, at least in my mind, would be to have handleSSNTPServer call both wg.Add and wg.Wait.  It would need to communicate with the Close methods via channels.  But this would be a more disruptive change requiring changes to both functions.  In the meantime  7cf32c8ac7913ae fixes the race.

markdryan:
@tpepper  Could you post the race logs for the first commit?

tpepper:
=== RUN   TestStartWorkloadLaunchCNCI                                           
==================                                                              
WARNING: DATA RACE                                                              
Write by goroutine 44:                                                          
  github.com/01org/ciao/ciao-controller.testStartWorkloadLaunchCNCI.func1()     
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:742 +0x1a6
                                                                                
Previous write by goroutine 112:                                                
  github.com/01org/ciao/ciao-controller.testStartWorkloadLaunchCNCI()           
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:785 +0x1169
  github.com/01org/ciao/ciao-controller.TestStartWorkloadLaunchCNCI()           
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:246 +0x5b
  testing.tRunner()                                                             
      /usr/src/golang-1.6/src/testing/testing.go:473 +0xdc                      
                                                                                
Goroutine 44 (running) created at:                                              
  github.com/01org/ciao/ciao-controller.testStartWorkloadLaunchCNCI()           
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:751 +0x706
  github.com/01org/ciao/ciao-controller.TestStartWorkloadLaunchCNCI()           
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:246 +0x5b
  testing.tRunner()                                                             
      /usr/src/golang-1.6/src/testing/testing.go:473 +0xdc                      
                                                                                
Goroutine 112 (running) created at:                                             
  testing.RunTests()                                                            
      /usr/src/golang-1.6/src/testing/testing.go:582 +0xae2                     
  testing.(*M).Run()                                                            
      /usr/src/golang-1.6/src/testing/testing.go:515 +0x11d                     
  github.com/01org/ciao/ciao-controller.TestMain()                              
      /usr/src/go/src/github.com/01org/ciao/ciao-controller/controller_test.go:865 +0x7cb
  main.main()                                                                   
      github.com/01org/ciao/ciao-controller/_test/_testmain.go:134 +0x20a       
==================                                                              
--- PASS: TestStartWorkloadLaunchCNCI (0.09s)

tpepper:
I should have noted to repro either of these:

cd ciao-controller
go test -v -race -cpu=4 -run TestStartWorkloadLaunchCNCI
go test -v -race -cpu=4 -run TestDeleteInstance

I've found if you don't specify -cpu >= 2 and do specify -run to run just a specific test case, go race doesn't find things.

markdryan:
@tpepper To me it looks like the race is with err and not instances.

tpepper:
@markdryan interesting point.  I assumed instances, because I previous to an earlier cleanup was truly seeing a data race in instances (it was used in the parent thread prior to the child thread writing it and was nil upon usage).  So the patch implicitly fixes a race on err also by using := instead of =, resulting in the child having its own err?

I think the channel fix anyway is beneficial in that it is more idiomatic?

I guess in my commit message I don't explicitly say what variable in the apparent closure could have been experiencing a data race. Either or both!  :)

markdryan:
@tpepper Yep, I think your new commit is better than the old code and fixes the issue.  I just wanted to understand where the race was coming from to make sure it wasn't in the real ciao code.

markdryan:
So for what it's worth, this PR is good for me.  I guess we still need @sameo approval.

sameo:
@tpepper @markdryan The SSNTP chunk looks good to me.

PR #362, opened 2016-07-19
ciao-launcher: Add qemu_test.go

markdryan:
This PR contains 3 commits that provide some unit testing for the code that manages qemu instances.  This code in itself is quite hard to test as we don't really want to launch qemu instances or create ISO or qcow images in our unit tests.  The tests then validate that that the go routines that manage qemu instances are working correctly, that ciao-launcher parses the result of qemu-img correctly and that it generates command line parameters for qemu as expected.

### DIFF
--- a/ciao-launcher/qemu.go
+++ b/ciao-launcher/qemu.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


markdryan:
Tested on the single machine VM.

kaccardi:
Looks good to me.

PR #363, opened 2016-07-19
ciao-launcher: Enable tracing

markdryan:
This commit adds tracing support to ciao-launcher.  When built with the
profile tag a new flag is added called --trace.  This flag can be used
to provide a file into which trace data can be written.  The data can
be analysed at a later stage using go tool trace.

### DIFF
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -35,6 +35,7 @@ import (
 )
 
 var profileFN func() func()
+var traceFN func() func()
 
 type uiFlag string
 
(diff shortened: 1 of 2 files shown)


tpepper:
TraceFN() could come with a unit test?

markdryan:
The code doesn't affect our coverage statistics as it's only enabled when you specify the profile tags when building.  Even if there was a unit test it wouldn't currently be run in travis.

tpepper:
My meta-worry is we're adding misc diagnostics over time, they're not normally used, they're not well documented or tested.  Sounds a lot like technical debt to me.

But the patch is fine in itself.

PR #368, opened 2016-07-20
ciao-cli: Add tenant name evironment variable

obedmr:
When user is part of more than one project (tenant), it
allows to use environment variables for passing the proper
project name instead of using the ``tenant-name`` or ``tenant-id``
parameters in ``ciao-cli`` command. It would verify if the
``CIAO_TENANT_NAME`` environment variable exists, it would use it;
if not, everything would normally proceed.


sameo:
This looks good to me and can be merged.

PR #390, opened 2016-07-21
ssntp: Add volumes attaching and detaching commands

sameo:
The AttachVolume and DetachVolume commands will be
typically sent by ciao-controller down to ciao-launcher
in order to respectively attach and detach storage
volumes to running instances.

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -34,6 +34,11 @@ import (
 	"github.com/golang/glog"
 )
 
+// custom errors
+var (
+	ErrNoTenant = errors.New("Tenant not found")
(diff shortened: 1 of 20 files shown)


markdryan:
Shouldn't this be submitted against kaccardi/topic/storage?

markdryan:
Also, I'm just wondering why AttachVolume and DetachVolume are in camel case.  Most of the other SSNTP commands are in uppercase, e.g., START, STOP, CONFIGURE.  Is it simply because AttachVolume is more natural Go name and we don't want to change the older command  names to avoid breaking things?

sameo:
@markdryan I will send the PR against @kaccardi branch, that makes sense.

PR #391, opened 2016-07-21
ssntp: Add volumes attaching and detaching commands

sameo:
The AttachVolume and DetachVolume commands will be
typically sent by ciao-controller down to ciao-launcher
in order to respectively attach and detach storage
volumes to running instances.


markdryan:
I think we're going to need AttachVolumeFailure and DetachVolumeFailure as well so launcher can report errors.

PR #394, opened 2016-07-25
Define a directory for CIAO packaging for major distributions.

erick0z:
- Add Debian packaging files for CIAO
- Dependencies already exists in Debian official repositories.

A directory called packaging were added. This directory should contain all the necessary files in order to construct a proper CIAO package for each major distribution(Fedora, Debian?). This PR add the files for Debian distribution.


packaging/deb/debian/control
```diff
@@ -0,0 +1,88 @@
...
+               golang-github-docker-engine-api-dev,
+               golang-github-docker-libnetwork-dev,
+               golang-github-vishvananda-netlink-dev,
+               golang-golang-x-net-dev,
+               golang-github-docker-docker-dev
+
+Standards-Version: 3.9.7
+Homepage: https://github.com/01org/ciao
+Vcs-Browser: https://anonscm.debian.org/cgit/pkg-go/packages/ciao.git
+Vcs-Git: https://anonscm.debian.org/git/pkg-go/packages/ciao.git
```
sameo: I think those 2 URLs are not correct.

packaging/deb/debian/copyright
```diff
@@ -0,0 +1,28 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: ciao
+Source: https://github.com/01org/ciao
+
+Files: *
+Copyright: 2016 01org
```
sameo: This should be an Intel copyright.

packaging/deb/debian/copyright
```diff
@@ -0,0 +1,28 @@
...
+License: Apache-2.0
+
+Files: debian/*
+Copyright: 2016
```
sameo: Same here.

sameo:
I'd prefer to see that one under the already existing `data/` directory.

tpepper:
While I get that "data/" is a go specific catch all for "stuff golang shouldn't try to compile" and would be the golang idiomatic location, I think "packaging" is a more descriptive top level for consumers of the project.  I wouldn't naturally go into a directory named "data/" in order to find deb or rpm specs, docker composition stuff or ansible scripts when an admin or distro builder is trying to figure out how to ship this ciao thing to their end users.

tpepper:
Turns out "data/" is _not_ special.  Rather "testdata/" is special in the language.  And these are definitely not test data files.

'.' and "_" prefixed file and directory names are not processed by the golang tools.

In that case, I suggest calling the directory "_DeploymentAndDirectoryPackaging/" and creating subdirectories "debian", "rpm", "ansible", "docker", etc.

obedmr:
@tpepper, agree with the subdirectories part.
But that "_DeploymentAndDirectoryPackaging" looks too long. Could we use something shorter? maybe "_deployment", "_packaging". ``data`` looks more like configuration files dir (I saw the systemd bytes).

tpepper:
I can't type.  I meant "_DeploymentAndDistroPackaging/"

That's descriptive and not horribly long, imho.  Looks ok in the github UI to me:
https://github.com/tpepper/ciao/tree/scratch

_DeploymentAndDistroPackaging/deb/debian/control
```diff
@@ -0,0 +1,86 @@
...
+
+Standards-Version: 3.9.7
+Homepage: https://github.com/01org/ciao
+XS-Go-Import-Path: github.com/01org/ciao
+Package: ciao
+Architecture: amd64
+Built-Using: ${misc:Built-Using}
+Depends: ${shlibs:Depends},
+         ${misc:Depends}
+Description: Ciao - Cloud Integrated Advanced Orchestrator
```
tpepper: Is it normal in debian to have this verbose of a description?  With this much detail, it will be more difficult to maintain and keep in sync with the project.  At least Go Report Card should not be the first line of the description.

erick0z: It is autogenerated and extracted from the github project description. It can be changed.

sameo:
@tpepper Does that look fine to you now ?
@erick0z I think we can remove the `[RFC]` prefix now, can't we ?

tpepper:
Yep lgtm

erick0z:
@sameo, @tpepper, @obedmr. Done, thanks!

PR #398, opened 2016-07-25
ciao-cli: Add Glance support

sameo:
With this PR, we can create, delete, modify, upload, download, list and show Glance images.

The PR also includes the gophercloud vendoring upgrade.


markdryan:
Looks good to me.  Builds fine and ciao-vendor is happy.

PR #402, opened 2016-07-26
Topic/storage pull request

kaccardi:
This pull request adds a cluster client package for our image service. Components of the cluster should be able to access the image filesystem directly without contacting the image api service. The client package is just a wrapper around some operations so that we can change filesystem layout without having each component need to be modified.  ciao-controller will use this client to check for an image's existence prior to creating a start workload request. This pull request also creates an initial ceph driver and adds configuration for cephx authentication to ciao-controller.

### DIFF
--- a/ciao-controller/controller_test.go
+++ b/ciao-controller/controller_test.go
@@ -19,13 +19,16 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
+	"path/filepath"
 	"testing"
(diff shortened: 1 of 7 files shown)


albertomurillo:
What about using the ceph golang bindings rather than executing the cli?

https://github.com/ceph/go-ceph

kaccardi:
the go-ceph package is a wrapper around the librdb C library and doesn't buy you anything over just using rbd directly.

PR #412, opened 2016-07-27
testutil: call flag.Parse in TestMain

kaccardi:
Because this test code is using ssntp, which uses glog, you
must call flag.Parse prior to using ssntp.


tpepper:
Is there a runtime panic without this?

markdryan:
@tpepper IIRC, I think your logs are just corrupted.  The logs are generated but you get an error message.

PR #419, opened 2016-07-29
travis: Fix build by restricting docker version.

tpepper:
This commit forces the docker version used in travis to be 1.11.2-0~trusty.
It seems that some of the network unit tests are not working with the
latest version of docker.


tpepper:
I've opened issue https://github.com/01org/ciao/issues/420 to track reverting this workaround once root caused and fixed.

PR #428, opened 2016-08-03
ciao-cnci-agent: minor fixups

tpepper:
Attempting to follow the README and run the script, I bumped into a couple
minor issues:

The image is downloaded by default, so describe how to download the image
and to where.  (Alternatively, could simplify the README by using the
"--download" option.)

Depending on OS specifics, the /dev/nbd0p* might not be created.  Running
'partprobe' is a fairly linux-OS-agnostic way to get them.  Given the
script currently only deals with raw format images, we could also simply do
a loopback mount and not deal with nbd and qemu-nbd, but I'm guessing
qemu-nbd is used to be better ready in the futue to handle qcow format
images in addition to raw?


mcastelino:
LGTM but Travis seemed to have failed for an unrelated reason on the tip

PR #436, opened 2016-08-05
testutil: add Shutdown() to clients and server

tpepper:
A Shutdown() helper is added to the testutil clients and server in order to
handle teardown of the SSNTP session and closing the testutil results
channels.

This should prevent channel writers that are blocked from causing
goroutine leaks across test cases.


tpepper:
Bah the locking and synchonization here is just totally fubar.  Will rerefix and submit new PR later.

PR #443, opened 2016-08-09
ciao-cli: code refactory of getTenant function

leoswaldo:
The getTenant function was evaluating single cases while it could have
a base to validate with a more structure scenarios as proposed below.

First validation: ensure the user has at least one project
Second validation: if the user has two or more project and not project
is specified, then print the projects the user has.
Third validation: iterate through all projects and see if the given
tentantID matches any project of the user.

If no condition matches, then it would return with the last instruction
in the function.

### DIFF
--- a/ciao-cli/identity.go
+++ b/ciao-cli/identity.go
@@ -252,36 +252,27 @@ func getTenant(username string, password string, tenantID string) (string, strin
 		return "", "", err
 	}
 
-	if len(projects) == 0 {
+	if len(projects) <= 0 {
 		return "", tenantID, fmt.Errorf("No tenant name for %s", username)
 	}
 
-	if len(projects) > 1 {
-		if tenantID == "" {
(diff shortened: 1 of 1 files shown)


tpepper:
lgtm

ciao-cli/identity.go
```diff
@@ -252,36 +252,27 @@ func getTenant(username string, password string, tenantID string) (string, strin
 		return "", "", err
 	}
 
-	if len(projects) == 0 {
+	if len(projects) <= 0 {
```
mcastelino: @leoswaldo When can this be <0? Is there a reason we changed this?

leoswaldo: no that I see, but I prefere to use <= 0  than just == 0, I mean it does not add extra complexity, and in the case in the future that function is modified and the  len(instances) becames a paramter that function would be a bit more ready, minimal prepation, not that big improvement  

This PR was to remove the code duplication as main purpose

Issue #334, opened 2016-07-01
Make bat tests work on travis with single machine setup


chavafg:
Run BAT with single machine setup in Travis 

Comments from IRC meeting: 

<mcastelino> also based on the work obedmr_ did... i.e. running dnsmasq on the macvtap, we should also be able to run single VM on Travis
<markusry> #action fuentess to work on integrating signle machine tests into travis
[11:08:48] <mcastelino> markusry: I see two steps
[11:08:55] <markusry> #action fuentess to open github issues for two tasks
[11:08:56] <mcastelino> integrate single machien with BAT
[11:09:04] <mcastelino> then single machine with BAT on Travis
[11:09:24] <markusry> sounds good.  I think that's what we've already captured with our actions
[11:09:47] <markusry> Let me make it explicit
[11:10:09] <obedmr_> fuentess: please let me know if need help on dnsmasq stuff
[11:10:17] <markusry> #info We need to integrate single machine with BAT and then get single machine working on travis

mcastelino:
Currently on Travis we lack the ability for the CNCI to connect to the underlay network to obtain an IPv4 Addresss.

Hence to support Single VM on Travis we will need to
1. Create a dummy (or some other hosting interface) to which the macvtap will attach itself to
2. Run a dnsmasq on the hosting interface so that the CNCI will get an IP address
3. The controller, scheduler and launcher should be able to talk to the CNCI
       - This means all network connections should go through hosting interface

mcastelino:
We will only be able to test launching of docker containers on Travis. Travis VM's do not expose VT-x. So the single VM script's VM launch tests should not be run on travis.

markdryan:
I could try modifying launcher to not pass --cpu host and --enable-kvm if VT-x is not available.  I'll give it a go to see if this works.

markdryan:
@kaccardi Can we close this one now?

PR #445, opened 2016-08-09
singlevm : Add networking automation (macvlan and DHCP Server)

obedmr:
This change adds the following features to Single VM:
- Auto macvlan interface creation with ``ip`` command
- DHCP Server setup with ``dnsmasq``
- When testing ssh connectivity, avoid fixed 7.2 openssh version.

This will help to accomplish #334

### DIFF
--- a/testutil/singlevm/configuration.yaml
+++ b/testutil/singlevm/configuration.yaml
@@ -9,6 +9,9 @@ configure:
   image_service:
     url: http://glance.example.com
   launcher:
+    compute_net: 192.168.12.1/24
+    mgmt_net: 192.168.12.1/24
     disk_limit: false
+    mem_limit: false
   identity_service:
     url: http://keystone.example.com
--- a/testutil/singlevm/configuration.yaml
+++ b/testutil/singlevm/configuration.yaml
@@ -9,8 +9,8 @@ configure:
   image_service:
     url: http://glance.example.com
   launcher:
-    compute_net: 192.168.12.1/24
-    mgmt_net: 192.168.12.1/24
+    compute_net: 198.51.100.1/24
+    mgmt_net: 198.51.100.1/24
     disk_limit: false
     mem_limit: false
   identity_service:
--- a/testutil/singlevm/setup.sh
+++ b/testutil/singlevm/setup.sh
@@ -207,6 +207,27 @@ sudo cp -f clear-"${LATEST}"-cloud.img /var/lib/ciao/images
 cd /var/lib/ciao/images
 sudo ln -sf clear-"${LATEST}"-cloud.img df3768da-31f5-4ba6-82f0-127a1a705169
 
+# Set macvlan interface
+if [ -x "$(command -v ip)" ]; then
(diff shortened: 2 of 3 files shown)


testutil/singlevm/configuration.yaml
```diff
@@ -9,6 +9,9 @@ configure:
   image_service:
     url: http://glance.example.com
   launcher:
+    compute_net: 192.168.12.1/24
```
mcastelino: Can we use the reserved IP subnet 198.51.100.0/24
This is a special network reserved for testing and will not show up in a real world network
See https://en.wikipedia.org/wiki/Reserved_IP_addresses

testutil/singlevm/setup.sh
```diff
@@ -207,6 +207,27 @@ sudo cp -f clear-"${LATEST}"-cloud.img /var/lib/ciao/images
 cd /var/lib/ciao/images
 sudo ln -sf clear-"${LATEST}"-cloud.img df3768da-31f5-4ba6-82f0-127a1a705169
 
+# Set macvlan interface
+if [ -x "$(command -v ip)" ]; then
```
mcastelino: you may want to check if the network is already setup and skip this step.
And user may run this script multiple times in the course of his/her workflow.

mcastelino: Oh.. I see you are doing a teardown later. Ignore this comment

testutil/singlevm/configuration.yaml
obedmr: Sure, /me applying the change

testutil/singlevm/setup.sh
```diff
@@ -207,6 +207,27 @@ sudo cp -f clear-"${LATEST}"-cloud.img /var/lib/ciao/images
...
+   sudo ip link set dev eth10 up
+   sudo ip -d link show eth10
+else
+    echo 'ip command is not supported'
+fi
+
+# Set DHCP server with dnsmasq
+sudo mkdir -p /var/lib/misc
+if [ -x "$(command -v ip)" ]; then
+    dnsmasq -C $ciao_scripts/dnsmasq.conf.macvlan0
```
mcastelino: Do we need to update our documentation to request the user to install dnsmasq?

mcastelino:
@obedmr LGTM with the minor IP subnet change comment and maybe a documentation update

testutil/singlevm/setup.sh
obedmr: yes, I will need to update docs for that.

mcastelino:
@obedmr one last point. We need to disable VM launch on Travis. We may want to add a parameter to disable VM launch so that we can use the exact same script on Travis.

obedmr:
@mcastelino, when running on travis, we could test the launch of Docker containers workloads, what do you think? or it should only test the rest and avoid VMs launching?

obedmr:
@mcastelino: BTW, subnet change is applied. thanks for the note.

PR #454, opened 2016-08-13
networking: Support both nil and empty subnet slices

### LINKED ISSUE #453: libsnnet fails to associate with networks
When the compute and management subnet slices len given to libsnnet are 0, libsnnet fails to associate as it only checks for nil slices.

sameo:
When not passing compute or management subnets, one could
either use nil or empty slices.

Fixes #453


mcastelino:
LTGM. Will merge once CI passes.

PR #458, opened 2016-08-15
ciao-controller: validate API docs with Travis

leoswaldo:
Automation of API Documentation.

CIAO-Controller has a API that exposes CIAO methods through HTTPS, documentation will be changing frequently due to the changes and new features in CIAO. 

Swagger will  allow us to automate the process of Docs generations. When we nave a new release the documentation will be updated, and every new PR needs to be validated it does not break the documentation.

This PR is to validate no issues with new integrations and the generation of docs for the API.

### DIFF
--- a/ciao-controller/compute.go
+++ b/ciao-controller/compute.go
@@ -13,6 +13,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 */
+// @SubApi Servers API [/v2.1/{tenant}/servers]
+// @SubApi Flavors API [/v2.1/{tenant}/flavors]
+// @SubApi Resources API [/v2.1/{tenant}/resources]
+// @SubApi Quotas API [/v2.1/{tenant}/quotas]
+// @SubApi Events API [/v2.1/{tenant}/events]
+// @SubApi Nodes API [/v2.1/nodes]
+// @SubApi Tenants API [/v2.1/tenants]
+// @SubApi CNCIs API [/v2.1/cncis]
+// @SubApi Traces API [/v2.1/traces]
 
 package main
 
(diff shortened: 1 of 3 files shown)


sameo:
This looks good to me.

ciao-controller/compute.go
```diff
@@ -1585,4 +1790,5 @@ func createComputeAPI(context *controller) {
 
 	service := fmt.Sprintf(":%d", computeAPIPort)
 	log.Fatal(http.ListenAndServeTLS(service, httpsCAcert, httpsKey, r))
+
```
sameo: Unneeded new line.

ciao-controller/compute.go
```diff
@@ -1544,7 +1749,7 @@ func createComputeAPI(context *controller) {
 	}).Methods("GET")
 
 	r.HandleFunc("/v2.1/flavors/{flavor}/servers/detail", func(w http.ResponseWriter, r *http.Request) {
-		listServerDetails(w, r, context)
+		listFlavorServerDetail(w, r, context)
```
sameo: Why are we changing this ?

ciao-controller/compute.go
```diff
@@ -1535,7 +1740,7 @@ func createComputeAPI(context *controller) {
 	}).Methods("GET")
 
 	r.HandleFunc("/v2.1/{tenant}/events", func(w http.ResponseWriter, r *http.Request) {
-		listEvents(w, r, context)
+		listTenantEvents(w, r, context)
```
sameo: Why are we changing this ?

ciao-controller/compute.go
leoswaldo: Swagger documents API calls 1 per function, and there were two API calling the same method, that means that one call would not be documented, for this reason, I had to create "secondary functions" that call the primary. This way we can document all API calls

ciao-controller/compute.go
leoswaldo: same case as previous one

ciao-controller/compute.go
```diff
@@ -1434,6 +1609,28 @@ func listEvents(w http.ResponseWriter, r *http.Request, context *controller) {
...
+// @Failure 500 {object} payloads.HTTPReturnErrorCode "The response contains the corresponding message and 50x corresponding code."
+// @Router /v2.1/{tenant}/events [get]
+// @Resource /v2.1/events
+// listTenantEvents is created with the only purpose of API documentation for method
+// /v2.1/{tenant}/events
+func listTenantEvents(w http.ResponseWriter, r *http.Request, context *controller) {
+	listEvents(w, r, context)
+}
+
+// @Title listTenantEvents
```
sameo: Title is wrong it seems.

ciao-controller/compute.go
sameo: Fair enough.

PR #459, opened 2016-08-15
bat: Fix tap package import and requirements

obedmr:
Update BAT testing with proper ``tap`` package import and requirements file was added to easy install with ``pip2``

Dependencies can be easily installed with:
```
     pip2 install -f requirements.txt
```


kaccardi:
Looks good to me.

PR #485, opened 2016-08-22
[DONOTMERGE] Add Ciao deployment via ansible

erick0z:
- The ansible playbooks were added from the clear-config-management project.
- The deployment of Mariadb and Keystone are containerized.
- [WIP] A new feature will be added in order to resolve dependencies based on each node role.


kaccardi:
would it make sense to have ciao deploy the keystone container itself when controller first boots rather than as part of the ansible scripts?

albertomurillo:
I think we are reducing the flexibility by integrating everything into ciao.

We were able to install keystone and reuse an existing database server but now mariadb is bundled in the keystone container.

Having keystone launched by ciao will make thing easier but also, will reduce the flexibility of having keystone in a separate (maybe HA) server.

tpepper:
If we optimize for simple case and continue with our self-hosting (ie: launching the CNCI as a system workload), I like the idea of ciao starting a keystone workload.  We could still have an option for people who want something more to disable this default simple mode.

kaccardi:
i agree with Tim - make it do the easiest possible thing by default, add a switch for people who want 'advanced' mode.

albertomurillo:
+ 1
That sounds good to me

kaccardi:
I think we need public IP support merged before we can do this - I think we'll need to assign a publicly accessible IP address to the keystone container.

_DeploymentAndDistroPackaging/ansible/README.md
```diff
@@ -0,0 +1,33 @@
+# CIAO Deployment with Ansible Playbooks
+
+* Work in Progress
+
```
ciao-github: Let's start cleaning this "Work-in-progress" comments

_DeploymentAndDistroPackaging/ansible/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Prework
+
+ - Deploy SSH key from the host running ansible, to all hosts in the cluster setup.
+ - Install Ansible and his dependencies.
+
```
obedmr: Some information that may be useful:
- How the ansible package is named at Ubuntu, ClearLinux, etc

_DeploymentAndDistroPackaging/ansible/README.md
```diff
@@ -0,0 +1,33 @@
...
+### Edit the hosts file according to your cluster setup
+```shell
+hosts.yml
+
+[controllers]
+controller.example.com
+
+[networks]
+network.example.com
```
obedmr: Is it the new standard we're going to follow with groups names? we could use ciao_network, ciao_controller, ciao_compute, etc

_DeploymentAndDistroPackaging/ansible/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+[computes]
+compute1.example.com
+compute2.example.com
+compute3.example.com
+```
+
+### Run the playbooks
+```shell
+ansible-playbook -i hosts ciao.yml --private-key=<ssh_key>
```
obedmr: I'd suggest to make it easier and use the Ansible Galaxy repositories that we already have at: https://galaxy.ansible.com/clearlinux/

leoswaldo:
@albertomurillo @erick0z  
Some recomendations for the docs:
-  Add a note saying something like "Required minimal configuration" where you ask to edit the hosts file
- Specify where the  hosts, ciao.yaml and the group_vars/all file is, just informational 
- Specify the user of the ssh key needs admin access

_DeploymentAndDistroPackaging/ansible/roles/ciao-network/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!) The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [2016] [Intel Corporation]
+
```
obedmr: We could remove the brackets "[ ]"

leoswaldo: you may also remove the text  from   "APPENDIX: How  ..."    to  "... third-party archives."

Issue #462, opened 2016-08-16
configuration: pass ceph cluster info


tpepper:
Modify the configuration payload topropagate the ceph cluster info, so that launcher can auto mount the /var/lib/ciao/images point if it is not already mounted.

markdryan:
@bryteise is going to work on this,.

markdryan:
The nice thing about this ticket is that you should be able to develop and test using only launcher and it's test tools.

Install launcher following the instructions [here](https://github.com/01org/ciao/blob/master/ciao-launcher/README.md).

For this ticket you need to work from the storage branch kaccardi/topic/storage.   So basically, you can do a 

go get -d github.com/01org/ciao/ciao-launcher
cd $GOPATH/src/github.com/01org/ciao/ciao-launcher
git checkout kaccardi/topic/storage
and then go build or go install

Once you've got it building, look further down the document for the section entitled 'Testing ciao-launcher in Isolation'.  Build the test tools and make sure you can send launcher commands via ciaolc.

Then it's time to code.  The problem we need to solve here is that we are hard coding two settings in [lines 403 and 404 of instance.go](https://github.com/01org/ciao/blob/kaccardi/topic/storage/ciao-launcher/instance.go#L403).  These values should come from the configuration data that launcher receives when it connects to the ssntp server.  Normally, this is scheduler, but to test and develop this feature we can use ciao-launcher-server as described in the readme above.

Launcher receives its configuration  information from the ssntp server in (main.go)[https://github.com/01org/ciao/blob/kaccardi/topic/storage/ciao-launcher/main.go#L380].  If you click on the link you'll see it currently retrieves the compute and management networks and disk and memory limits.  We need to add two new settings, key path and ceph id.

Off the top of my head, here are the steps

1. Add the new settings to the (configuration structure)[https://github.com/01org/ciao/blob/kaccardi/topic/storage/payloads/configure.go#L84] and make sure you update configure_test.go as well.
2. Add two new command line parameters to ciao-launcher-server for the keypath and the management id.  These will allow you to specify different values when testing.  The values you pass into the command line of ciao-launcher-server should make it to launcher.
3. Modify main.go in launcher to receive the new settings.  It's safe to store them in global variables in a similar way to the other settings.
4. Modify instance.go to use the global variables instead of the hardcoded values.
5. Update the test cases in configuration/congfiuration_test.go to include your new settings.
6. Update configuration/README.md

@obedmr, @mrkz can you think of anything I've missed here?

[...]

Issue #388, opened 2016-07-20
add discovery of storage cluster to ciao configuration


kaccardi:
Add discovery of storage cluster (storage url and cephx secrets in ssntp configuration frame)
Add mountpoint for cephfs based storage at /var/lib/ciao to contain images/logs/etc. and mount cephfs there.

sameo:
@kaccardi Could you please share what a ceph monitor URL would look like ?

Also, I think the cephfs mounting itself should be done by ciao-launcher, after receiving the configuration data. What do you say @markdryan ?

kaccardi:
@sameo is this complete then?

sameo:
Yes, this is complete.

PR #488, opened 2016-08-24
Storage cluster config handling

bryteise:
Add storage cluster configuration module that adds CephDriver fields for the storage url and cephx secrets to be passed from the ssntp configuration frame.

This address #462 and partially #388.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -58,8 +58,8 @@ var logDir = "/var/lib/ciao/logs/controller"
 
 var imagesPath = flag.String("images_path", "/var/lib/ciao/images", "path to ciao images")
 
-var keyringPath = flag.String("ceph_keyring", "/etc/ceph/ceph.client.ciao.keyring", "path to ceph client keyring")
-var cephID = flag.String("ceph_id", "ciao", "ceph client id")
+var keyringPath = flag.String("ceph_keyring", "", "path to ceph client keyring")
+var cephID = flag.String("ceph_id", "", "ceph client id")
 
 func init() {
 	flag.Parse()
@@ -86,14 +86,6 @@ func main() {
 
 	context := new(controller)
 	context.ds = new(datastore.Datastore)
-
-	context.BlockDriver = func() storage.BlockDriver {
-		driver := storage.CephDriver{
-			SecretPath: *keyringPath,
-			ID:         *cephID,
-		}
-		return driver
-	}()
 
 	context.image = image.Client{MountPoint: *imagesPath}
 
@@ -136,6 +128,12 @@ func main() {
 	identityURL = clusterConfig.Configure.IdentityService.URL
 	serviceUser = clusterConfig.Configure.Controller.IdentityUser
 	servicePassword = clusterConfig.Configure.Controller.IdentityPassword
+	if *keyringPath == "" {
+		*keyringPath = clusterConfig.Configure.Storage.SecretPath
+	}
+	if *cephID == "" {
+		*cephID = clusterConfig.Configure.Storage.CephID
+	}
 
 	if *singleMachine {
 		hostname, _ := os.Hostname()
@@ -161,6 +159,14 @@ func main() {
 		serviceUserName: serviceUser,
 		servicePassword: servicePassword,
 	}
+
+	context.BlockDriver = func() storage.BlockDriver {
+		driver := storage.CephDriver{
+			SecretPath: *keyringPath,
+			ID:         *cephID,
+		}
+		return driver
+	}()
 
 	context.id, err = newIdentityClient(idConfig)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


ciao-controller/main.go
```diff
@@ -89,8 +89,8 @@ func main() {
 
 	context.BlockDriver = func() storage.BlockDriver {
 		driver := storage.CephDriver{
-			SecretPath: *keyringPath,
-			ID:         *cephID,
+			SecretPath: keyringPath,
```
markdryan: Aren't we initialising this too early?  As far as I can tell, keyringPath and cephID will still contain their defaults at this point.  Changing the global variables after this point isn't going to do anything.

ciao-controller/main.go
```diff
@@ -136,6 +136,8 @@ func main() {
 	identityURL = clusterConfig.Configure.IdentityService.URL
 	serviceUser = clusterConfig.Configure.Controller.IdentityUser
 	servicePassword = clusterConfig.Configure.Controller.IdentityPassword
+	keyringPath = clusterConfig.Configure.Storage.SecretPath
```
markdryan: Maybe we should only perform the assignment if the SecretPath is "".  Otherwise there's no point in providing defaults, as they'll never be used.

ciao-launcher/tests/ciao-launcher-server/server.go
```diff
@@ -55,6 +57,8 @@ func init() {
 	flag.StringVar(&mgmtNet, "mgmt-net", "", "Management Subnet")
 	flag.BoolVar(&diskLimit, "disk-limit", true, "Use disk usage limits")
 	flag.BoolVar(&memLimit, "mem-limit", true, "Use memory usage limits")
+	flag.StringVar(&secretPath, "secret-path", "", "Key path for Ceph")
```
markdryan: We could use the same defaults that we use in controller here.

ciao-launcher/tests/ciao-launcher-server/server.go
```diff
@@ -491,6 +495,8 @@ func createConfigFile(confPath string) error {
 	conf.Configure.Launcher.MemoryLimit = memLimit
 	conf.Configure.Launcher.ComputeNetwork = []string{computeNet}
 	conf.Configure.Launcher.ManagementNetwork = []string{mgmtNet}
+	conf.Configure.Storage.SecretPath = secretPath
```
markdryan: And check to see if the values received from the command line are "" or not only performing the assignment if they are !="".

ciao-controller/main.go
bryteise: Ah indeed!

ciao-controller/main.go
bryteise: I was actually thinking of completely removing the defaults in a later patch since I had the two fields being required for the min config but I wanted to get feedback first.

ciao-launcher/tests/ciao-launcher-server/server.go
bryteise: So for the server I was thinking of making these required to be passed in but again looking for feedback on use of default values.

ciao-controller/main.go
sameo: Yes, those 2 fields should be required through validMinConf() but you want to allow for a grace period and give people time to update their cluster configuration, i.e.:

* `validMinConf()` should spit a warning for now if the storage config is not present for e.g. a couple of weeks. Then after that we make it a hard requirement.
* Send an email to the ciao mailing list about the upcoming requirement for the storage configuration entry.
* keyringPath should be set to the default value if we receive a "" string from the configuration server. And we should keep it like that for the long term.

bryteise: Okay and then in the case of the ciao-launcher I'll do the same with replacing the hardcoded value with a default value.

ciao-launcher/tests/ciao-launcher-server/server.go
bryteise: Actually if I add the default value in the flag, wouldn't I always want to be doing the assignment since it would contain either a value passed by the user or the default value?

markdryan: Yep you're correct. Ignore this comment.

bryteise:
updates from comments
- ciao-launcher-server defaults added
- fix setting cephdriver options before reading config
- don't add the storage options as required in validating the min config
- change how options are handled for the ciao-controller and ciao-launcher (no default options for either, if the option is specified have it override the configuration setting)

markdryan:
@bryteise Could you sign off your commits?

markdryan:
Looks good to me.  I've tested the launcher changes and they work fine.

@kaccardi  are you okay with the controller changes?

sameo:
I am ok with the configuration and payloads changes.

kaccardi:
looks fine to me.

PR #489, opened 2016-08-25
Prepare

tpepper:
@ikeydoherty started hacking up some code to make our documentation, deployment and ansible scripting easier.  The idea is that our components' binaries articulate their system dependencies, instead of the documentation and scripting.  Then at runtime the osprepare package's helper code detects the running OS and automatically installs the dependencies if they're not present.  This PR includes the ciao osprepare package, some cleanup to make our CI happy, and initial usage in controller, launcher and scheduler.

Still to do is adding plumbing to check that a particular version is present (eg: docker >= 1.12).  Also I wonder if we need to condition the call to osprepare.InstallDeps() in launcher to allow a network node to check for and install additional dependencies and am hoping @markdryan and @mcastelino can offer guidance there.

### DIFF
--- a/ciao-controller/deps.go
+++ b/ciao-controller/deps.go
@@ -1,0 +1,32 @@
+//
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package main
+
+import "github.com/01org/ciao/osprepare"
+
+var controllerDeps = osprepare.PackageRequirements{
+	/* sqlite3 is not strictly required, but useful for debug */
+	"clearlinux": {
+		{"/usr/bin/sqlite3", "cloud-control"},
+	},
+	"fedora": {
+		{"/usr/bin/sqlite3", "sqlite"},
+	},
+	"ubuntu": {
+		{"/usr/bin/sqlite3", "sqlite3"},
+	},
+}
--- a/ciao-controller/deps.go
+++ b/ciao-controller/deps.go
@@ -21,12 +21,12 @@ import "github.com/01org/ciao/osprepare"
 var controllerDeps = osprepare.PackageRequirements{
 	/* sqlite3 is not strictly required, but useful for debug */
 	"clearlinux": {
-		{"/usr/bin/sqlite3", "cloud-control"},
+		{BinaryName: "/usr/bin/sqlite3", PackageName: "cloud-control"},
 	},
 	"fedora": {
-		{"/usr/bin/sqlite3", "sqlite"},
+		{BinaryName: "/usr/bin/sqlite3", PackageName: "sqlite"},
 	},
 	"ubuntu": {
-		{"/usr/bin/sqlite3", "sqlite3"},
+		{BinaryName: "/usr/bin/sqlite3", PackageName: "sqlite3"},
 	},
 }
--- a/ciao-launcher/deps.go
+++ b/ciao-launcher/deps.go
@@ -25,21 +25,21 @@ import "github.com/01org/ciao/osprepare"
 // fuser for qemu instance pid
 
 var launcherClearLinuxCommonDeps = []osprepare.PackageRequirement{
-	{"/usr/bin/qemu-system-x86_64", "cloud-control"},
-	{"/usr/bin/xorriso", "cloud-control"},
-	{"/usr/sbin/fuser", "cloud-control"},
+	{BinaryName: "/usr/bin/qemu-system-x86_64", PackageName: "cloud-control"},
+	{BinaryName: "/usr/bin/xorriso", PackageName: "cloud-control"},
+	{BinaryName: "/usr/sbin/fuser", PackageName: "cloud-control"},
 }
 
 var launcherFedoraCommonDeps = []osprepare.PackageRequirement{
-	{"/usr/bin/qemu-system-x86_64", "qemu-system-x86"},
-	{"/usr/bin/xorriso", "xorriso"},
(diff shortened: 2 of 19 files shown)


ciao-launcher/deps.go
```diff
@@ -0,0 +1,41 @@
...
+	// xorriso for cloud init config drive
+
+	"clearlinux": {
+		{"/usr/bin/docker", "cloud-control"},
+		{"/usr/bin/qemu-system-x86_64", "cloud-control"},
+		{"/usr/bin/xorriso", "cloud-control"},
+	},
+	"fedora": {
+		{"/usr/bin/docker", "docker-engine"},
+		{"/usr/bin/qemu-system-x86_64", "qemu-system-x86"},
```
markdryan: launcher also uses fuser.

osprepare/distro.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (d *clearLinuxDistro) getID() string {
+	return "clearlinux"
+}
+
+// Correctly split and format the command, using sudo if appropriate, as a
+// common mechanism for the various package install functions.
+func sudoFormatCommand(command string, packages []string) bool {
+	toInstall := strings.Join(packages[0:], " ")
```
markdryan: Isn't packages[0:] == packages?

ikeydoherty: Ya but I had compiler warnings with that before

osprepare/distro.go
```diff
@@ -0,0 +1,268 @@
...
+
+	if syscall.Geteuid() == 0 {
+		executable = splits[0]
+		args = fmt.Sprintf(strings.Join(splits[1:], " "), toInstall)
+	} else {
+		executable = "sudo"
+		args = fmt.Sprintf(command, toInstall)
+	}
+
+	c := exec.Command(executable, strings.Split(args, " ")...)
```
markdryan: @tpepper What did we decide about logging? Do we want the outputs of these commands going to stdout,stderr or to glog, even if it's a local glog?

osprepare/distro.go
ikeydoherty: Note we can likely simplify this and remove the sudo treatment now, as long as euid == 0 is always maintained during component initialization.

osprepare/main.go
```diff
@@ -0,0 +1,118 @@
...
+			pkgsMissing = append(pkgsMissing, pkg.PackageName)
+		}
+		return pkgsMissing
+	}
+	return nil
+}
+
+// InstallDeps installs all the dependencies defined in a component
+// specific PackageRequirements in order to enable running the component
+func InstallDeps(reqs PackageRequirements) {
```
markdryan: So have we taken the decision that our components should try to run even if their dependencies cannot be installed?  In some ways this makes sense, if for example docker is not supported on a platform.  You could still launch VMs via qemu.  I just wanted to check this though.

markdryan: We need to add a context parameter here so we can cancel the InstallDeps function.  Otherwise launcher will not be able to clean up correctly when asked to exit by the OS.  I'm guessing InstallDeps is a long running function.

osprepare/distro.go
markdryan: Weird.  That shouldn't give you any warnings.  Note that if you want to reslice an entire slice or array you can just use [:] but you shouldn't need it here.

osprepare/main.go
mcastelino: @markdryan ideally we need to support a docker only (and maybe even or VM only mode of operation) for launcher. This will allow ciao to be used on virtual clusters (say in a public cloud provider).

markdryan:
I don't think the NN needs any dependencies that are not needed for the CN.  What it doesn't need though is docker, at least I don't think it does.  @mcastelino  can you confirm.  I wonder if there is some way we can tell os.prepare not to check for docker in some cases.

mcastelino:
@tpepper @markdryan The NN does as far as I remember and no special setup; and it does not need docker.

tpepper:
I'm thinking we'd do something like:

if role.IsNetAgent() {
    osprepare.InstallDeps(launcherNetNodeDeps)
} else {
    osprepare.InstallDeps(launcherComputeNodeDeps)
}

And declare two deps list variables instead of the current one, with the NN having no docker in the list, and perhaps adding some useful NN specific netmgmty stuff if needed over time.

tpepper:
@ikeydoherty and I think this is merge ready.  A next step would be to remove docker from the .travis.yml and see that osprepare gets it for us, but I'm not quite following the .travis.yml internals and could use @markdryan and @mcastelino help on this front.

ciao-controller/deps.go
```diff
@@ -0,0 +1,32 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package main
+
```
kaccardi: would it be possible to only include this with some kind of debug build tag? I hate including dependencies that people might not really use and don't entire trust that the future removal will happen.

tpepper: We can open a ticket to implement debug/developer package adds and note sqlite3 in controller is the test case for it (ie: is no longer added normally, but is added if the debug build tag is present).

tpepper: I just pushed a commit to make a very trivial implementation of debug/developer for sqlite3 in controller specifically.  We'd still want to make something more generic

kaccardi: this works for me.

osprepare/main.go
```diff
@@ -0,0 +1,123 @@
...
+var BootstrapRequirements = PackageRequirements{
+	"ubuntu": {
+		{"/usr/bin/cephfs", "ceph-fs-common"},
+		{"/usr/bin/ceph", "ceph-common"},
+	},
+	"clearlinux": {
+		{"/usr/bin/ceph", "storage-cluster"},
+	},
+}
+
```
kaccardi: are we missing fedora bootstrap requirements here?

tpepper:
As requested by @kaccardi:

To be clear osprepare is still WIP and doesn't cover all needs yet.  In particular there's work to do related to issue #462 and #388 and changes we'll need to make if logs come to be on /var/lib/ciao/logs across the cluster (ie: can't start logging until ceph's mounted which is after ceph is installed if need and the server has been contacted for the ceph server config endpoint info).

PR #494, opened 2016-08-26
Complete implementation of the OpenStack block api interface in ciao-controller

kaccardi:
Modify ciao-controller to implement the interface for the openstack DeleteVolume, ShowVolumeDetail and ListVolumeDetails endpoints.

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1406,6 +1406,8 @@ func (ds *Datastore) AddBlockDevice(device types.BlockData) error {
 	return nil
 }
 
+// DeleteBlockDevice will delete a volume from the datastore.
+// It also deletes it from the tenant's list of devices.
 func (ds *Datastore) DeleteBlockDevice(ID string) error {
 	// lock both tenants and devices maps
 	var err error
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -121,6 +121,7 @@ type persistentStore interface {
 	getAllBlockData() (map[string]types.BlockData, error)
 	createBlockData(data types.BlockData) error
 	updateBlockData(data types.BlockData) error
+	deleteBlockData(string) error
 	getTenantDevices(tenantID string) (map[string]types.BlockData, error)
 	createStorageAttachment(a types.StorageAttachment) error
 	getAllStorageAttachments() (map[string]types.StorageAttachment, error)
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1419,14 +1419,16 @@ func (ds *Datastore) DeleteBlockDevice(ID string) error {
 	if ok {
 		delete(ds.blockDevices, ID)
 		delete(ds.tenants[dev.TenantID].devices, ID)
-
-		go ds.db.deleteBlockData(ID)
-	} else {
-		err = ErrNoBlockData
 	}
 
 	ds.tenantsLock.Unlock()
 	ds.bdLock.Unlock()
+
+	if ok {
+		go ds.db.deleteBlockData(ID)
+	} else {
+		err = ErrNoBlockData
+	}
 
 	return err
 }
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1406,6 +1406,29 @@ func (ds *Datastore) AddBlockDevice(device types.BlockData) error {
 	return nil
 }
 
+func (ds *Datastore) DeleteBlockDevice(ID string) error {
+	// lock both tenants and devices maps
+	var err error
+
+	ds.bdLock.Lock()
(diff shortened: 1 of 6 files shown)


ciao-controller/openstack_volume.go
```diff
@@ -233,7 +233,36 @@ func (c *controller) ListVolumes(tenant string) ([]block.ListVolume, error) {
...
+	if err != nil {
+		return vols, err
+	}
+
+	for _, data := range devs {
+		var vol block.VolumeDetail
+
+		vol.ID = data.ID
+		vol.Size = data.Size
+		vol.OSVolTenantAttr = data.TenantID
```
tpepper: The prior patch for ShowVolumeDetails() has a tenant check, ie:

    if data.TenantID != tenant {

Should this ListVolumesDetail() function also do so for symmetry and safety?

Or should neither function do the check, with both functions trusting that a function higher in the stack has validated that the passed tenant is reasonable and trusting that GetBlockDevices(tenant) will not return data not for the tenant?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1406,6 +1406,29 @@ func (ds *Datastore) AddBlockDevice(device types.BlockData) error {
 	return nil
 }
 
+func (ds *Datastore) DeleteBlockDevice(ID string) error {
+	// lock both tenants and devices maps
+	var err error
+
+	ds.bdLock.Lock()
+	ds.tenantsLock.Lock()
+
```
tpepper: These locks aren't commented in ciao-controller/internal/datastore/datastore.go so I'm not 100% sure the design of what they're meant to protect, but I think you have introduced a race below...

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1406,6 +1406,29 @@ func (ds *Datastore) AddBlockDevice(device types.BlockData) error {
...
+	dev, ok := ds.blockDevices[ID]
+	if ok {
+		delete(ds.blockDevices, ID)
+		delete(ds.tenants[dev.TenantID].devices, ID)
+
+		go ds.db.deleteBlockData(ID)
```
tpepper: deleteBlockData() locks the sqliteDB so that can't have concurrent accesses.  But there's no relation between that and the bdLock (protecting blockDevices[]?) and tenantsLock (protecting tenants[]?).

The way it's coded here it feels like the deleteBlockData() call and its locking are meant to be hierarchically below the bdLock and tenantsLock, ie: happen while those are also held.  That would insure there's no racing.  But since it's a go routine, it's possible for the go ds.db.deleteBlockData(ID) to very minimally "start", and DeleteBlockDevice() run to completion and unlock tenantsLock and bdLock and return.

It may be then that other tenant and block device adds/removes happen.

And only later the deleteBlockData() happens, perhaps for a tenant or block device which is now removed.  I _think_ looking at sqlite3db.go that this is actually ok today.  But this is the type of things that leads to bugs later.

I suggest either adding a wait group or channel in the deleteBlockData() flow so this caller can await its completion, or moving the unlock calls upward so a future person looking at this code sees the db deletion is independent of any attempt to maintain coherency on tenants[] and blockdevices[], eg:

    	if ok {
    		delete(ds.blockDevices, ID)
    		delete(ds.tenants[dev.TenantID].devices, ID)
    		ds.tenantsLock.Unlock()
    		ds.bdLock.Unlock()
    
    		go ds.db.deleteBlockData(ID)
    		return err
    	}
    
       	ds.tenantsLock.Unlock()
       	ds.bdLock.Unlock()
    
    	return ErrNoBlockData

tpepper:
Other than the couple questions on consistency and the hypothetical race (which looks to be "ok" in practice), looks good to me.

ciao-controller/openstack_volume.go
kaccardi: this one we don't have to check the tenant ID on because the call to GetBlockDevices() uses the tenant as an argument. You do need to check on ListVolumeDetail() because the argument is simply the volume ID and the call to get the block device by volume ID doesn't look at the tenant.

ciao-controller/internal/datastore/datastore.go
kaccardi: it's ok to have this happen asynchronously. The reason it's ok is that each block device is only ever operated on if it exists in the cache. If it doesn't exist in the cache, it is treated as a failure. the sqlite db is really just used as long term persistent datastore. There *is* a possibility that if ciao-controller is terminated prior to the go routine which is doing the asynchronous update has completed, we'll leave behind a volume that was meant to be deleted that will then be repopulated into the cache with ciao-controller is restarted. but that would happen regardless of any additional measures I could take with wait groups or channels.

kaccardi: I'll move the go call as you suggested to outside the loop. that was actually a mistake anyway and an artifact of when I had this sqlite update happening synchronously.

ciao-controller/openstack_volume.go
tpepper: Ok I see now GetBlockDevice() versus GetBlockDevices().

PR #495, opened 2016-08-29
add unit test for block api os-detach action

kaccardi:
update the openstack api unit tests to include one for the os-detach action.


tpepper:
lgtm

PR #496, opened 2016-08-29
ciao-controller: datastore: use Fatal() for test errors

kaccardi:
In all cases, if a test fails a check we should terminate the test
instead of continuing. Replace use of Error() with Fatal() in unit
tests. Also do not use Fatalf if we are not using a format string.

### DIFF
--- a/ciao-controller/internal/datastore/datastore_test.go
+++ b/ciao-controller/internal/datastore/datastore_test.go
@@ -762,7 +762,7 @@ func TestGetBatchFrameStatistics(t *testing.T) {
 
 	_, err = ds.db.getBatchFrameStatistics("batch_frame_test")
 	if err != nil {
-		t.Fatalf(err.Error())
+		t.Fatal(err)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -345,33 +345,33 @@ func TestDeleteInstanceNetwork(t *testing.T) {
...
-		t.Error(err)
+		t.Fatal(err)
 	}
 
 	ip := net.ParseIP(instance.IPAddress)
 
 	ipBytes := ip.To4()
 	if ipBytes == nil {
-		t.Error(errors.New("Unable to convert ip to bytes"))
+		t.Fatal(errors.New("Unable to convert ip to bytes"))
```
tpepper: Any reason to have the errors.New() here?

ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -757,12 +757,12 @@ func TestGetBatchFrameStatistics(t *testing.T) {
...
 	err := ds.HandleTraceReport(trace)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 
 	_, err = ds.db.getBatchFrameStatistics("batch_frame_test")
 	if err != nil {
-		t.Errorf(err.Error())
+		t.Fatalf(err.Error())
```
tpepper: Should this just pass in "err" same as the others?

ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -797,45 +797,45 @@ func TestGetBatchFrameSummary(t *testing.T) {
...
 	err := ds.HandleTraceReport(trace)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 
 	_, err = ds.db.getBatchFrameSummary()
 	if err != nil {
-		t.Errorf(err.Error())
+		t.Fatalf(err.Error())
```
tpepper: same in multiple from here forward...why sometimes "err" and sometimes "err.Error()"?

tpepper:
In addition to going from Error() to Fatal(), can you make them consistent?  Ie: all either Fatal(err) or Fatal("string")?  Unless there's a reason for the mixing around of different arguments to Fatal() that was already there.

ciao-controller/internal/datastore/datastore_test.go
kaccardi: Fatalf takes a string format - Fatal does not. As to why - that is a result of different people writing the unit tests and copy and paste. making the usage consistent would be a different PR I believe. This PR is just to prevent the tests from continuing when they should halt.

PR #505, opened 2016-08-30
ciao-controller: confirm tenant in openstack volume interface

kaccardi:
Prior to doing any work in the openstack_volume interface, make
sure that we have confirmed the tenant. If this is a tenant we
don't know about, we need to add them to our tenant list and
start up a cnci for them. This is permissible because they have
passed authentication already.

### DIFF
--- a/ciao-controller/controller_test.go
+++ b/ciao-controller/controller_test.go
@@ -695,7 +695,12 @@ func TestDetachVolumeFailure(t *testing.T) {
 }
 
 func TestDetachVolumeByAttachment(t *testing.T) {
-	err := context.DetachVolume("invalidTenant", "invalidVolume", "attachmentID")
+	tenant, err := addTestTenant()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = context.DetachVolume(tenant.ID, "invalidVolume", "attachmentID")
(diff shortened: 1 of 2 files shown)


markdryan:
@kaccardi It looks like this one needs to be rebased.

ciao-controller/controller_test.go
```diff
@@ -695,7 +695,12 @@ func TestDetachVolumeFailure(t *testing.T) {
 }
 
 func TestDetachVolumeByAttachment(t *testing.T) {
-	err := context.DetachVolume("invalidTenant", "invalidVolume", "attachmentID")
+	tenant, err := addTestTenant()
```
tpepper: Can you add a negative (expected to fail) test case here?

kaccardi: This test is already expected to fail. Which part of the test are you wanting to see expanded?

tpepper: Seeing "invalidTenant" change to "addTestTenant()" I thought there needed to coverage of confirmTenant().  But you're not actually adding confirmTenant() but rather just some additional appropriate usage of that function which should already be validated for good and bad tenants by other distinct test cases.

PR #506, opened 2016-08-30
ciao-controller: datastore: update attachments correctly

kaccardi:
Do not delete attachments based on stats without first checking
the instance ID.  This fixes a bug which caused the attachment
list to get deleted whenever a node with no volumes attached sent
stats. Remember to delete attachments from the persistent datastore
when it's time to remove them.


markdryan:
@kaccardi Looks okay to me.  Any reason not to merge?

kaccardi:
@markdryan it's ready to merge.

PR #509, opened 2016-09-01
ciao-launcher: Fix docker dependency in ubuntu

### LINKED ISSUE #508: osprepare: installs wrong package in ubuntu
Compute node requiremets in ubuntu includes 'docker' which is a KDE package.
the container runtime package is called 'docker.io'

albertomurillo:
The 'docker' package is being installed as a dependency to
get docker insalled in Ubuntu, however, the package needed
is 'docker-engine'

From apt-cache search docker
  - docker - System tray for KDE3/GNOME2 docklet applications
  - docker-engine - Docker: the open-source application container engine

Fixes #508

### DIFF
--- a/ciao-launcher/deps.go
+++ b/ciao-launcher/deps.go
@@ -62,5 +62,5 @@ var launcherComputeNodeDeps = map[string][]osprepare.PackageRequirement{
 	"fedora": append(launcherFedoraCommonDeps,
 		osprepare.PackageRequirement{BinaryName: "/usr/bin/docker", PackageName: "docker-engine"}),
 	"ubuntu": append(launcherUbuntuCommonDeps,
-		osprepare.PackageRequirement{BinaryName: "/usr/bin/docker", PackageName: "docker"}),
+		osprepare.PackageRequirement{BinaryName: "/usr/bin/docker", PackageName: "docker-engine"}),
 }


tpepper:
Fixes issue #508

markdryan:
I'm a bit confused by all this.

I thought we were supposed to be pulling docker from apt.dockerproject.org so we get the latest version.  If we were to do this, the name of the docker package should be docker-engine and not docker-io.

There seems to be some code in osprepare for handling external repos but it seems incomplete and unused.  @tpepper  any idea what the status/plan here is?  Perhaps we should just merge this patch for now.  This would give us docker 1.10.3 on Ubuntu.

tpepper:
Yes osprepare does need polished to add ability to install non-distro shipped bits, such as docker.

tpepper:
Despite https://github.com/01org/ciao/pull/509#issuecomment-245117481 I do think the patch here is correct.  Osprepare is limited in its ability and needs to do:

if /usr/bin/docker is not present or not at version high enough, then if docker.io in Ubuntu official repo is high enough version then install it, else install docker-engine from apt.dockerproject.org.

markdryan:
Maybe we should remove docker entirely from the ubuntu install deps for the time being.  I seem to remember that we needed a version > 1.10.3.  Perhaps @mcastelino could confirm.  If this is the case, we could end up having to manually uninstall the docker installed by osprepare before we install the latest version by hand.

mcastelino:
@markdryan We need docker version 1.10.3 because prior builds have a bug in the network plugin framework. So the ciao docker network plugin will not work with prior releases.

tpepper:
Re: https://github.com/01org/ciao/pull/509#issuecomment-245548553 I think it will be a recurring issue and on each distro.  osprepare needs its minversion check added and for docker needs to know how to satisfy a minversion from apt.dockerproject.org if it's not satisfiable from the distro itself.

tpepper:
(which is also to say: osprepare does not function correctly/completely today...I'm working on patches)

albertomurillo:
docker.io in Ubuntu is Version: 1.11.2-0ubuntu5~16.04

So I think we can merge this patch until we get the extra code to setup the proper repository and get the proper docker-engine working.

It is better to have docker.io than a KDE applet

markdryan:
@albertomurillo Ubuntu 16.04 has docker 1.10.3.  @mcastelino has confirmed that this is good enough to run ciao.  The problem is though, what happens if we merge this patch and then we run launcher on a machine that already has docker 1.12 installed from the docker repos.  Will we end up with two versions of docker?  Will they interfere with each other?  Also, installing docker-io isn't much help for those on an earlier version of Ubuntu.  I think @tpepper had a suggestion in the bug to change the package name to docker-engine.  This will require the user to manually add the docker repos to the sources.list but from then on launcher and osprepre will do the right thing. I vote we do this.

albertomurillo:
It doesn't make too much sense to make osprepare install docker-engine while asking 
the user to manually add the repository.

The idea of osprepare is that it will install the required pacakges automatically.

If you put as a requirement that the repository has to be manually installed and in other 
hand we are also proposing ansible to make the deployment easier, then the logical thing 
to do is let ansible install the repository.

Currently, there is a clearlinux.docker role that installs docker repository and docker-engine
in Clearlinux, ubuntu and fedora.

I see two options from here:
1) Let ansible install docker repository, docker-engine and probably uninstall docker.io 
with the help of clearlinux.docker role and remove docker from osprepare.

2) Install docker-engine with this patch and make another patch to make osprepare setup
docker repository and probably uninstall docker.io.

I'm fine with either

markdryan:
Yes, actually, I was suggesting 2 above.  The requirement on users to manually add the docker sources would only be temporary.  The plan has always been for osprepare to install things from the docker repos directly.  The code's just not ready yet.

albertomurillo:
Im updating this PR to cover case #2 then.
And temporary enabling case #1 in ciao-deploy tool

PR #514, opened 2016-09-01
fix travis errors for ineffassign

kaccardi:
ineffassign has changed and caught some new errors. This PR addresses those errors so that travis will pass.


tpepper:
Great cleanups folks.  It's annoying that the tools break us, but it's good that they encourage an ever improving quality bar.

Issue #504, opened 2016-08-30
osprepare needs a logging abstraction


markdryan:
osprepare is currently logging everything to stdout which isn't very useful when debugging an installation on a remote node.  I realise that we have an outstanding issue that the drive to which logs files are written might not be available when osprepare runs, but OS prepare shouldn't know or care about this.  It should simply provide an abstract logging interface in the same way that the ssntp and qemu packages do.  This way it can log what ever it likes to this interface and let it's clients worry about where the logs actually get written.  See the qemu and launcher for details about how to do this.

https://github.com/01org/ciao/blob/master/qemu/qmp.go#L56
https://github.com/01org/ciao/blob/master/ciao-launcher/qemu.go#L47

bryteise:
Instead of adding another new logging abstraction, consolidate the launcher and qemu abstractions and introduce that abstraction into osprepare. Leave the other loggers alone for now.

markdryan:
Fixed in PR #515

PR #515, opened 2016-09-01
osprepare: Add logging abstraction

bryteise:
Instead of writing to stdout/stderr directly, adopt the use of a glog
using abstraction similar to the qemu component.

Signed-off-by: William Douglas <william.r.douglas@gmail.com>

Address #504

### DIFF
--- a/osprepare/distro.go
+++ b/osprepare/distro.go
@@ -174,7 +174,7 @@ type distro interface {
 	// InstallPackages should implement the installation
 	// of packages using distro specific methods for
 	// the given target list of items to install
-	InstallPackages(packages []string) bool
+	InstallPackages(packages []string, logger OSPLog) bool
 
 	// getID should return a string specifying
 	// the distribution ID (e.g: "clearlinux")
@@ -211,11 +211,11 @@ func (d *clearLinuxDistro) getID() string {
 
 // Correctly split and format the command, using sudo if appropriate, as a
 // common mechanism for the various package install functions.
-func sudoFormatCommand(command string, packages []string) bool {
-	toInstall := strings.Join(packages, " ")
-
+func sudoFormatCommand(command string, packages []string, logger OSPLog) bool {
 	var executable string
 	var args string
+
+	toInstall := strings.Join(packages, " ")
 	splits := strings.Split(command, " ")
 
 	if syscall.Geteuid() == 0 {
@@ -227,18 +227,31 @@ func sudoFormatCommand(command string, packages []string) bool {
 	}
 
 	c := exec.Command(executable, strings.Split(args, " ")...)
-	c.Stdout = os.Stdout
-	c.Stderr = os.Stderr
-
-	if err := c.Run(); err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to run command: %s", err)
+	read, err := c.StdoutPipe()
+	if err != nil {
+		logger.Warningf("Unable to create command output pipe: %s", err)
+	}
+	c.Stderr = c.Stdout
+	err = c.Start()
+	if err != nil {
+		logger.Errorf("Unable to run command: %s", err)
 		return false
 	}
+	scanner := bufio.NewScanner(read)
+	for scanner.Scan() {
+		logger.Infof(scanner.Text())
+	}
+	err = c.Wait()
+	if err != nil {
+		logger.Errorf("Error running command: %s", err)
+		return false
+	}
+
 	return true
 }
 
-func (d *clearLinuxDistro) InstallPackages(packages []string) bool {
-	return sudoFormatCommand("swupd bundle-add %s", packages)
+func (d *clearLinuxDistro) InstallPackages(packages []string, logger OSPLog) bool {
+	return sudoFormatCommand("swupd bundle-add %s", packages, logger)
 }
 
 // os-release *ubuntu*
@@ -250,8 +263,8 @@ func (d *ubuntuDistro) getID() string {
 	return "ubuntu"
 }
 
-func (d *ubuntuDistro) InstallPackages(packages []string) bool {
-	return sudoFormatCommand("apt-get --yes --force-yes install %s", packages)
+func (d *ubuntuDistro) InstallPackages(packages []string, logger OSPLog) bool {
+	return sudoFormatCommand("apt-get --yes --force-yes install %s", packages, logger)
 }
 
 // Fedora
@@ -263,6 +276,6 @@ func (d *fedoraDistro) getID() string {
 }
 
 // Use dnf to install on Fedora
-func (d *fedoraDistro) InstallPackages(packages []string) bool {
-	return sudoFormatCommand("dnf install -y %s", packages)
+func (d *fedoraDistro) InstallPackages(packages []string, logger OSPLog) bool {
+	return sudoFormatCommand("dnf install -y %s", packages, logger)
 }
--- a/osprepare/distro_test.go
+++ b/osprepare/distro_test.go
@@ -32,3 +32,54 @@ func TestGetDistro(t *testing.T) {
 		t.Fatal("Invalid ID for distro")
 	}
 }
+
+type ospTestLogger struct{}
+
+func (l ospTestLogger) V(level int32) bool {
+	return true
+}
+
+var info []string
+var warning []string
+var error []string
+
+func (l ospTestLogger) Infof(format string, v ...interface{}) {
+	info = append(info, format)
+}
+
+func (l ospTestLogger) Warningf(format string, v ...interface{}) {
+	warning = append(warning, format)
+}
+
+func (l ospTestLogger) Errorf(format string, v ...interface{}) {
+	error = append(error, format)
+}
+
+func TestSudoFormatCommandLogging(t *testing.T) {
+	info = []string{}
+	if getDistro() == nil {
+		t.Skip("Unsupported test distro")
+	}
+	l := ospTestLogger{}
+
+	sudoFormatCommand("echo -n foo\nbar", []string{}, l)
+
+	if info[0] != "foo" && info[1] != "bar%!(EXTRA string=)" {
+		t.Fatal("Incorrect log message received")
+	}
+}
+
+func TestSudoFormatCommandBadCommandReturn(t *testing.T) {
+	error = []string{}
+	if getDistro() == nil {
+		t.Skip("Unsupported test distro")
+	}
+	l := ospTestLogger{}
+
+	if sudoFormatCommand("false", []string{}, l) {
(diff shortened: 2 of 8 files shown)


osprepare/main.go
```diff
@@ -114,14 +179,17 @@ func InstallDeps(reqs PackageRequirements) {
...
 		}
 	}
 }
 
 // Bootstrap installs all the core dependencies required to bootstrap the core
 // configuration of all Ciao components
-func Bootstrap() {
-	InstallDeps(BootstrapRequirements)
+func Bootstrap(logger OSPLog) {
+	if logger == nil {
```
markdryan: This isn't needed as InstallDeps will default to the null logger.

osprepare/versions_test.go
```diff
@@ -33,7 +33,7 @@ func TestGetQemu(t *testing.T) {
 	if pathExists("/usr/bin/qemu-system-x86_64") == false {
 		t.Skip("Qemu not installed, cannot validate version get")
 	}
-	if vers := getQemuVersion(); vers == "" {
+	if vers := getQemuVersion(nil); vers == "" {
```
markdryan: Should this be ospNullLogger{} instead of nil?

markdryan:
osprepare only seems to log errors.  It doesn't actually log anything if everything works okay, which is a shame as you can't really tell by looking at the logs whether or not it's done anything.  Could you add an extra commit with some informational logging, such as, 

OS detected: Ubuntu
Missing packages detected: qemu, docker
Missing packages installed.

osprepare/distro.go
```diff
@@ -231,14 +231,14 @@ func sudoFormatCommand(command string, packages []string) bool {
 	c.Stderr = os.Stderr
```
markdryan: The actually output from apt-get and the other installers is still going to stdout/stderr which isn't really very useful except on a developer's machine.  I think we probably want to log this information, so could you also redirect this to the logs are well?

markdryan:
@bryteise I'm afraid this isn't working for me.

```
I0909 10:16:13.995117    3416 main.go:184] Missing packages detected: [docker]
W0909 10:16:14.383293    3416 distro.go:243] Problem reading command stdout: read |0: bad file descriptor
W0909 10:16:14.383318    3416 distro.go:249] Problem reading command stderr: read |0: bad file descriptor
I0909 10:16:14.383322    3416 main.go:188] Missing packages installed.
I0909 10:16:14.383396    3416 overseer.go:786] Starting Overseer
I0909 10:16:14.383403    3416 overseer.go:788] Allocated: Disk 0 Mem 0 CPUs 0
```

The problem is that you cannot use Run and StdoutPipe togther.  Run calls Wait and Wait closes the file descriptors created by Pipe.  You need to use Wait instead.

I was wondering whether it would be nicer to simply use a single pipe for both Stdout and Stderr.  I realise if you did this you wouldn't be able to distinguish between the two streams in the ciao logs using Info and Warning, but you the log would contain the information in the same order that a user would see it if he were to run the command from the command line so it might be easier to debug.  You could also read the single pipe using a Scanner, a line at a time.  This way the logs would get updated as the installation process runs and not when it's finished.  This might be useful it apt-get hangs, if for example you've forgotten to set up your proxies.

bryteise:
@markdryan So I'm not sure about the scanner approach. If the command being executed doesn't timeout then the code is stuck doing a dance where it has some internally configured timeout value, looping around trying to read bytes and doing book keeping when it hasn't seen bytes for longer than the timeout value.

Is that something you want the code to be doing?

Otherwise I sadly can't keep the stdout and stderr in the order they were sent afaik in go without using the https://golang.org/pkg/os/exec/#Cmd.CombinedOutput command which looks like it will cause the same issue Run does.

I can put them into the same log level but they'll be ordered in whatever order I write them to that log =(. Open to suggestions if there is another way to look at the problem.

markdryan:
I was thinking that we could call StdoutPipe and then set cmd.Stderr = cmd.Stdout
```
read, err := cmd.StdoutPipe()
if err != nil {
  return err
}
cmd.Stderr = cmd.Stdout
err := cmd.Start()
scanner := bufio.NewScanner(read)
for scanner.Line() {
   glog.Info(scanner.Text())
}
err := cmd.Wait()
```

The pipe will get closed when the process exits so this shouldn't block.  Note I haven't tried this and maybe it's a bit nasty as I had to look at the runtime code to figure out it would work.  What do you think?  The code is simpler as we don't have to spawn a go routine.

bryteise:
I'll give it a try, my memory is a little fuzzy on this but I want to say that without the go routine bits might have been dropped because I seem to remember wait not acting like it was flushing the pipe or at least the scanner was picking up the last bytes.

As an aside for the future, with tracing we probably could go back to splitting stderr and stdout because they can both have the same parent in the trace tree.

bryteise:
@markdryan So after building this and making some test cases the issue with the scanner is if the process hangs before printing a newline then the scanner won't print it out. If that's worth skipping the go routine I'll push that change.

markdryan:
@bryteise I see.  So the process hasn't quit but it has just hung for some reason, e.g., proxies not set?  But does the go routine version behave any differently?  It's calling ReadString('\n') which is presumably blocking until it sees a '\n'.

bryteise:
@markdryan Yes this is for the hung case. The reason that we get data is that the ReadString will return with an io.EOF error (the reason for that I do not understand) based on https://golang.org/pkg/bufio/#Reader.ReadString "If ReadString encounters an error before finding a delimiter, it returns the data read before the error and the error itself (often io.EOF)".

I have a little gist that demonstrates my behavior if you'd like to see my test: https://gist.github.com/bryteise/c4d21a96140c654a56e97417134c5257

markdryan:
@bryteise The reason V1 works is that it calls ReadString repeatedly in a busy loop.  On my machine ReadString was called 122882 times before it returns any data.  The reason is that he bytes.Buffer.Read returns EOF if the buffer is empty.  readOut also races with V1.

Also, I'm not sure we want to log partial lines.  Our logs might contain unwanted new lines if the execed process prints out some data without a new line, hangs for 1 second and then prints out some more data + a new line.  In this case the data would be split over two lines in ciao-launcher's logs as the glog logging functions always append a new line to the logged string.

bryteise:
@markdryan Ah so ReadString is just not blocking at all. Though since there is a last read after the Wait call is there a race issue with the output content since the readOut should have exited or do I need to have another lock to ensure that behavior?

The log messages being split over multiple lines definitely races. I'm not really sure if not logging is better though because a few of the package managers display progress meters and not knowing which package might have failed to install seems harder to deal with than reading output potentially split over multiple lines (as the only place this logging pattern occurs is with package installation commands).

bryteise:
So I see the CI tests passing on my branch but failing in the pull request (with 1.7 but not tip) and I'm not really sure why. Is this test giving a false negative?

bryteise:
Ah it was a false negative, and I stupidly didn't capture what test it was before I restarted the test thinking I'd have something saving the status in the build history which there was not X(.

osprepare/distro_test.go
```diff
@@ -32,3 +32,38 @@ func TestGetDistro(t *testing.T) {
...
+var info []string
+var warning []string
+var error []string
+
+func (l ospTestLogger) Infof(format string, v ...interface{}) {
+	info = append(info, format)
+}
+
+func (l ospTestLogger) Warningf(format string, v ...interface{}) {
+	info = append(warning, format)
```
markdryan: Should this be warning?

osprepare/distro_test.go
```diff
@@ -32,3 +32,38 @@ func TestGetDistro(t *testing.T) {
...
+}
+
+func (l ospTestLogger) Errorf(format string, v ...interface{}) {
+	info = append(error, format)
```
markdryan: Should this be

error = append(error, format)

markdryan:
We need to rebase as well.

markdryan:
@bryteise This looks good to me.  @kaccardi there's a small controller change.  Is this okay for you?

osprepare/main.go
```diff
@@ -97,31 +159,38 @@ func collectPackages(dist distro, reqs PackageRequirements) []string {
...
 	// Nothing requested to install
 	if reqs == nil {
 		return
 	}
 	if reqPkgs := collectPackages(distro, reqs); reqPkgs != nil {
-		if distro.InstallPackages(reqPkgs) == false {
-			fmt.Fprintf(os.Stderr, "Failed to install: %s\n", strings.Join(reqPkgs, ", "))
+		logger.Infof("Missing packages detected: %v", reqPkgs)
+		if distro.InstallPackages(reqPkgs, logger) == false {
+			logger.Errorf("Failed to install: %s", strings.Join(reqPkgs, ", "))
```
markdryan: Okay, sorry, I do have a few more comments.  We probably want a return here otherwise we'll get 

Failed to install: ...
Missing packages installed

Also, I don't think Install packages is detecting errors correctly.  When I got an error from apt, I saw the following logs

```
I0916 14:38:04.155526    5546 main.go:177] OS Detected: ubuntu
I0916 14:38:04.155844    5546 main.go:184] Missing packages detected: [ceph-fs-common]
I0916 14:38:04.162392    5546 distro.go:242] E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
I0916 14:38:04.162654    5546 distro.go:242] E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
W0916 14:38:04.163153    5546 distro.go:246] Error running command: exit status 100
I0916 14:38:04.163394    5546 main.go:188] Missing packages installed.
```

markdryan: In fairness the last example I posted is an existing bug in osprepare and not related to this PR.  So we can enter a separate bug for this.  All we need to do is to the fix, 

Failed to install: ...
Missing packages installed

kaccardi:
@markdryan yes, the controller change is ok with me.

osprepare/main.go
bryteise: @markdryan I'd prefer to just add a patch to this pull request as this change exposes the bug much more clearly.

markdryan: Great.  Sounds good to me.

tpepper:
FYI for a couple issues you've noted I've got patches coming.  @bryteise and I should compare our WIP's and coordinate changes on osprepare.

markdryan:
@tpepper This PR is okay for me now.  Would you like to wait or can we merge it now?

tpepper:
@markdryan feel free to merge and I'll rebase/fixup #578

markdryan:
Great.  Let's merge.  Thanks @bryteise.

PR #519, opened 2016-09-01
ciao-launcher: Support volume attach on start

markdryan:
This PR fixes some storage related issues in launcher.  The main change is provided by 70c4da3 which adds support to attaching volumes to launcher instances when they are started, i.e., when the instance is specified in the start workload.  Such instances can be live detached at a later stage if required.  Attaching the volumes on boot should be faster than live attaching as we do not need to rbd map the volume first.

The other commits rename a poorly named parameter in an interface that was inconsistently named among the types that implement that interface and also fix a gocyclo issue introduced by 70c4da3.

### DIFF
--- a/ciao-launcher/docker.go
+++ b/ciao-launcher/docker.go
@@ -243,7 +243,7 @@ func (d *docker) deleteImage() error {
 	return err
 }
 
-func (d *docker) startVM(vnicName, ipAddress string) error {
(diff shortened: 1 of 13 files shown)


markdryan:
I stupidly closed this PR by accident.  I've re-opened and rebased the commits against master.

markdryan:
@tpepper Could you check and merge this one?

PR #534, opened 2016-09-06
storage: implement ceph block delete

tpepper:
This is a trivial implementation of the missing delete function in our ceph
driver and addition of a simple test case.

### DIFF
--- a/ciao-storage/ceph.go
+++ b/ciao-storage/ceph.go
@@ -58,9 +58,10 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 }
 
 // DeleteBlockDevice will remove a rbd image from the ceph cluster.
-// Not implemented yet.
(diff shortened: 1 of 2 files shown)


ciao-storage/ceph.go
```diff
@@ -58,9 +58,10 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 }
 
 // DeleteBlockDevice will remove a rbd image from the ceph cluster.
-// Not implemented yet.
-func (d CephDriver) DeleteBlockDevice(string) error {
-	return nil
+func (d CephDriver) DeleteBlockDevice(volumeUUID string) error {
+	cmd := exec.Command("rbd", "--keyring", d.SecretPath, "--id", d.ID, "rm", volumeUUID)
+	_, err := cmd.CombinedOutput()
```
markdryan: If we don't care about the output it might be more efficient to simply call return cmd.Run()

PR #535, opened 2016-09-06
ciao-controller: Return an empty volumes list

albertomurillo:
OpenStack blockstorage v2 api state that the return value
for /v2/tenant/volumes/detail should be a list of volumes,
however, ciao was returning {"volumes":null}.

With this fix ciao now returns the correct response as
{"volumes": []}


markdryan:
golint isn't happy

```
/home/travis/gopath/src/github.com/01org/ciao/ciao-controller/openstack_volume.go:282:2: can probably use "var vols []block.VolumeDetail" instead

```

albertomurillo:
But "var vols []block.VolumeDetail" is what introduced the bug hum =/

PR #538, opened 2016-09-08
ciao-storage: Fix a couple of issues.

albertomurillo:
This PR will handle some issues im facing with ceph storage

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -67,7 +67,6 @@ var logDir = "/var/lib/ciao/logs/controller"
 
 var imagesPath = flag.String("images_path", "/var/lib/ciao/images", "path to ciao images")
 
-var keyringPath = flag.String("ceph_keyring", "", "path to ceph client keyring")
 var cephID = flag.String("ceph_id", "", "ceph client id")
 
 func init() {
@@ -137,9 +136,6 @@ func main() {
 	identityURL = clusterConfig.Configure.IdentityService.URL
 	serviceUser = clusterConfig.Configure.Controller.IdentityUser
 	servicePassword = clusterConfig.Configure.Controller.IdentityPassword
-	if *keyringPath == "" {
-		*keyringPath = clusterConfig.Configure.Storage.SecretPath
-	}
 	if *cephID == "" {
 		*cephID = clusterConfig.Configure.Storage.CephID
 	}
@@ -175,8 +171,7 @@ func main() {
 
 	ctl.BlockDriver = func() storage.BlockDriver {
 		driver := storage.CephDriver{
-			SecretPath: *keyringPath,
-			ID:         *cephID,
+			ID: *cephID,
 		}
 		return driver
 	}()
--- a/ciao-launcher/instance.go
+++ b/ciao-launcher/instance.go
@@ -409,8 +409,7 @@ func startInstance(instance string, cfg *vmConfig, wg *sync.WaitGroup, doneCh ch
 	ac *agentClient, ovsCh chan<- interface{}) chan<- interface{} {
 
 	storageDriver := storage.CephDriver{
-		SecretPath: secretPath,
-		ID:         cephID,
+		ID: cephID,
 	}
 
 	var vm virtualizer
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -66,7 +66,6 @@ var networking bool
 var hardReset bool
 var diskLimit bool
 var memLimit bool
-var secretPath string
 var cephID string
 var simulate bool
 var maxInstances = int(math.MaxInt32)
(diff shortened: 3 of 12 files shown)


kaccardi:
looks good to me once it passes CI.

markdryan:
@albertomurillo Could you rebase?  The patch seems to have some conflicts.

markdryan:
I'm afraid there's still some pieces missing here:

1. You need to pass the config path to the Ceph Driver launcher creates.   You do this for controller but not for launcher.  See here:

https://github.com/01org/ciao/blob/master/ciao-launcher/instance.go#L402

2. You need to tell qemu the location of the ceph configuration file.  Qemu assumes that it is in /etc/ceph/ceph.conf and if it's not there qemu will not be able to mount any ceph volumes on boot.  So you need to pass the config path to qemuV.startVM and then modify generateQEMULaunchParam.  See here for information about passing the config path to qemu.

http://docs.ceph.com/docs/jewel/rbd/qemu-rbd/

albertomurillo:
There are 3 variables to configure ceph access for the ciao nodes:

- ceph_id: CephX user to use to authenticate
ceph/qemu uses the 'admin' user by default unless other user is specified with 
'--id user' (for ceph cli) or 'id=user' (qemu cli) making ceph_id var useful to keep.

- config_path: path to ceph configuration file
ceph uses /etc/ceph/$cluster.conf file by default and both ceph/qemu have cmdline 
options to override the location of the file.

- secret_path: path to keyring file for `ceph_id` user
ceph uses /etc/ceph/$cluster.client.$user.keyring file by default.
another file can be specified for ceph cli with '--keyring file' option but there is
no equivalent for qemu cli.

Since we cannot use an alternate location for the keyring file in qemu, it means that at 
least one file assosiated with ceph HAS to be in /etc/ceph directory making pointless
to keep the ability to store ceph configuration file in another location.

I suggest to just keep the ceph_id option and remove secret_path, config_path from the cli
invocations and document which are the required ceph files that the user has to copy to
ciao nodes.

markdryan:
@albertomurillo You correctly identify that the launcher code is not passing the secret information to qemu.  But I think it is possible using password_secret.  But I don't yet know how to use this to specify the path.

markdryan:
Okay, so in 2.5.0 I think we need to do

```
  $QEMU -drive file=rbd:pool/image:id=myname:\
               key=QVFDVm41aE82SHpGQWhBQXEwTkN2OGp0SmNJY0UrSE9CbE1RMUE=:\
               auth_supported=cephx
```

i.e., just add the key parameter.  We'd need to open the file to extract the key.  In later versions of qemu there's a more secure way to do this, but 2.5.0 i what Ubuntu currently ships.

albertomurillo:
This is a very insecure way to do it since '$ ps -a' will show the 'secret' key

albertomurillo:
Now im questioning: is there a *need* to store those files in a non-standard location? whats the use case?

markdryan:
Indeed.  That's why they fixed it

https://patchwork.kernel.org/patch/8081781/

The new syntax being

```
  echo "QVFDVm41aE82SHpGQWhBQXEwTkN2OGp0SmNJY0UrSE9CbE1RMUE=" > poolkey.b64
  $QEMU -object secret,id=secret0,file=poolkey.b64,format=base64 \
        -drive driver=rbd,filename=rbd:pool/image:id=myname:\
               auth_supported=cephx,password-secret=secret0
```

Let me just double check the 2.5.0 sources to make sure the new format is not supported.

markdryan:
I checked the 2.5.0 source.  The more secure method of passing ceph secrets is not supported.

markdryan:
@kaccardi What do you think about this?  If we want to make the ceph secret configurable and we want to support Ubuntu 16.04 we'll need to pass the ceph password to qemu on the command line, which as @albertomurillo points out is not very secure.  The options I see are:

1. Don't make the password configurable right now.
2. Just live with the insecurity for the time being and fix it in a few months when Ubuntu is upgraded.  We'd need to enter a ticket for this of course.  Fedora 24 and 25 should be okay already.
3. Get launcher to detect the qemu version and use the secure method of passing the secret to qemu if it is supported.

Obviously, the insecurity would only exist if the secret is specified in the configuration data.

kaccardi:
@albertomurillo @markdryan I would go for option 1 - don't make the password configurable right now. This is going to require some changes to controller and ciao-storage to not use the keyring option.

albertomurillo:
@kaccardi should I cancel the commit to make the conf file configurable too ?

kaccardi:
@albertomurillo what is the status of this PR? Are you still wanting it reviewed and merged?

albertomurillo:
@kaccardi it is ready to be merged once tests pass

markdryan:
Tested this morning locally with launcher in isolation and I can still attach ceph volumes to docker containers.

PR #540, opened 2016-09-08
ciao-launcher: install bootstrap dependencies

albertomurillo:
osprepare.Bootstrap function was designed to install
dependencies required in all ciao cluster nodes but
was unused.

This change makes use of osprepare.Bootstrap() function
to install ceph cli on all nodes.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -143,6 +143,7 @@ func main() {
 		*cephID = clusterConfig.Configure.Storage.CephID
 	}
 
+	osprepare.Bootstrap()
 	osprepare.InstallDeps(controllerDeps)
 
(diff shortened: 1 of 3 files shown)


markdryan:
I'm wondering whether scheduler really has a dependency on ceph.  Do we just need this for ciao-controller and ciao-launcher.  @kaccardi, @tpepper any comments?

tpepper:
Scheduler doesn't have a dep on ceph today, but if/when we shift logs to ceph, scheduler node would also want ceph.

markdryan:
@kaccardi Just need your approval and then we can merge.

kaccardi:
@markdryan looks good to me.

PR #542, opened 2016-09-09
ciao-cli: Support specifying CA file in environment or command line

### LINKED ISSUE #180: Pass the needed CA to ciao-cli
Instead of expecting ciao-cli users to have the controller CA installed on their machine, we should allow them to pass the CA through e.g. a ciao-cli --ca-file option.

rbradford:
With this change the CA certificate file used for communicating with
keystone and the controller can be specified:

a.) either on the command line with -ca-file,
b.) or with an envionment variable, CIAO_CA_CERT_FILE,

thus avoiding the need to import the certificate into you local store.

This change required switching from openstack.AuthenticatedClient() to
the lower level API in order to override the transport options on the
HTTP client.

Fixes #180

### DIFF
--- a/ciao-cli/identity.go
+++ b/ciao-cli/identity.go
@@ -17,7 +17,9 @@
 package main
 
 import (
+	"crypto/tls"
 	"fmt"
+	"net/http"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/rackspace/gophercloud"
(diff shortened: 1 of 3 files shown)


ciao-cli/main.go
```diff
@@ -139,8 +141,11 @@ const (
 	ciaoPasswordEnv    = "CIAO_PASSWORD"
 	ciaoComputePortEnv = "CIAO_COMPUTEPORT"
 	ciaoTenantNameEnv  = "CIAO_TENANT_NAME"
+	ciaoCACertFileEnv  = "CIAO_CA_CERT_FILE"
```
sameo: @rbradford Could you please document this new env variable in the ciao-cli README.md file ?
The code itself looks good to me.

ciao-cli/identity.go
```diff
@@ -121,7 +123,19 @@ func getScopedToken(username string, password string, projectScope string) (stri
 		AllowReauth:      true,
 	}
 
-	provider, err := openstack.AuthenticatedClient(opt)
+	provider, err := openstack.NewClient(opt.IdentityEndpoint)
+	if err != nil {
+		return "", "", "", nil
```
markdryan: should we have an errorf statement here before we return?

ciao-cli/main.go
rbradford: @sameo Done (and for command line option too.) I also described using this instead of installing into the system keystore.

ciao-cli/identity.go
rbradford: Done.

sameo:
This change looks good to me.

PR #547, opened 2016-09-12
Remove support for Go 1.6

### LINKED ISSUE #541: Remove support for Go 1.6
We need to remove support for Go 1.6.  Docker no longer compiles with 1.6 and so we cannot update our vendored docker packages while we still support 1.6.  In addition, moving to 1.7 will allow us to start using the new ..Context functions in the standard library which are very useful for cancelling long running tasks.

We need to

1. Update the minimum requirements for Go in all of our documentation.
2. Remove the 1.6 build from travis.
3. Revert 15f0830d2aa67109333638288ef9b79360225a66
4. Replace all instances of golang.org/x/net/context with context in ciao's code.
5. Revendor golang.org/x/net/context.  Note we should be able to get rid of this entirely if we were to update our vendored docker libs.  But this will probably introduce other problems and so should be down separately.

markdryan:
This PR removes ciao support for Go 1.6.  It

* Disables the travis go 1.6 builds and removes some hacks in the travis.yml file that enable the travis 1.6 builds to work.
* Replaces all instances of the golang.org/x/net/context package in the ciao code with the new standard library context package.
* Fixes all the bugs detected by go vet.  go vet has some new checks for the context package.
* Revendored the golang.org/x/net/context package.  This is still vendored as it is still used by some of our dependencies.  We needed to update the source as the 1.7 version of this package has code which converts between the new context package and the old golang.org/x/net/context package.  Once we update our docker dependencies, we should be able to get rid of golang.org/x/net/context  entirely.

Fixes #541


markdryan:
Once this is merged we need to send out an email to the mailing list.

PR #550, opened 2016-09-12
qemu: Build qemu parameters from a configuration structure

sameo:
LaunchQemu() now takes a Config structure that contains more
descriptive fields than raw qemu parameters.

LaunchQemu is now simpler to call and more extensible as supporting more
qemu parameters would mean expanding Config instead of changing the API.

### DIFF
--- a/ciao-launcher/qemu.go
+++ b/ciao-launcher/qemu.go
@@ -361,7 +361,8 @@ func launchQemuWithNC(params []string, fds []*os.File, ipAddress string) (int, e
 		ncString := "socket,port=%d,host=%s,server,id=gnc0,server,nowait"
 		params[len(params)-1] = fmt.Sprintf(ncString, port, ipAddress)
 		var errStr string
-		errStr, err = qemu.LaunchQemu(context.Background(), params, fds, qmpGlogLogger{})
+
+		errStr, err = qemu.LaunchCustomQemu(context.Background(), "", params, fds, qmpGlogLogger{})
 		if err == nil {
 			glog.Info("============================================")
 			glog.Infof("Connect to vm with netcat %s %d", ipAddress, port)
@@ -378,7 +379,7 @@ func launchQemuWithNC(params []string, fds []*os.File, ipAddress string) (int, e
 
 	if port == 0 || (err != nil && tries == vcTries) {
 		glog.Warning("Failed to launch qemu due to chardev error.  Relaunching without virtual console")
-		_, err = qemu.LaunchQemu(context.Background(), params[:len(params)-4], fds, qmpGlogLogger{})
+		_, err = qemu.LaunchCustomQemu(context.Background(), "", params[:len(params)-4], fds, qmpGlogLogger{})
 	}
 
 	return port, err
@@ -397,7 +398,7 @@ func launchQemuWithSpice(params []string, fds []*os.File, ipAddress string) (int
 		}
 		params[len(params)-1] = fmt.Sprintf("port=%d,addr=%s,disable-ticketing", port, ipAddress)
 		var errStr string
-		errStr, err = qemu.LaunchQemu(context.Background(), params, fds, qmpGlogLogger{})
+		errStr, err = qemu.LaunchCustomQemu(context.Background(), "", params, fds, qmpGlogLogger{})
 		if err == nil {
 			glog.Info("============================================")
 			glog.Infof("Connect to vm with spicec -h %s -p %d", ipAddress, port)
@@ -416,7 +417,7 @@ func launchQemuWithSpice(params []string, fds []*os.File, ipAddress string) (int
 	if port == 0 || (err != nil && tries == vcTries) {
 		glog.Warning("Failed to launch qemu due to spice error.  Relaunching without virtual console")
 		params = append(params[:len(params)-2], "-display", "none", "-vga", "none")
-		_, err = qemu.LaunchQemu(context.Background(), params, fds, qmpGlogLogger{})
+		_, err = qemu.LaunchCustomQemu(context.Background(), "", params, fds, qmpGlogLogger{})
 	}
 
 	return port, err
@@ -519,7 +520,7 @@ func (q *qemuV) startVM(vnicName, ipAddress, cephID string) error {
 
 	if !launchWithUI.Enabled() {
 		params = append(params, "-display", "none", "-vga", "none")
-		_, err = qemu.LaunchQemu(context.Background(), params, fds, qmpGlogLogger{})
+		_, err = qemu.LaunchCustomQemu(context.Background(), "", params, fds, qmpGlogLogger{})
 	} else if launchWithUI.String() == "spice" {
 		var port int
 		port, err = launchQemuWithSpice(params, fds, ipAddress)
--- a/qemu/examples_test.go
+++ b/qemu/examples_test.go
@@ -38,10 +38,10 @@ func Example() {
 	// resources
 	params = append(params, "-m", "370", "-smp", "cpus=2")
 
-	// LaunchQemu should return as soon as the instance has launched as we
+	// LaunchCustomQemu should return as soon as the instance has launched as we
 	// are using the --daemonize flag.  It will set up a unix domain socket
 	// called /tmp/qmp-socket that we can use to manage the instance.
-	_, err := qemu.LaunchQemu(context.Background(), params, nil, nil)
+	_, err := qemu.LaunchCustomQemu(context.Background(), "", params, nil, nil)
 	if err != nil {
 		panic(err)
 	}
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -329,16 +329,66 @@ func (dev SerialDevice) Valid() bool {
 	return true
 }
 
+// RTCBaseType is the qemu RTC base time type.
+type RTCBaseType string
+
+// RTCClock is the qemu RTC clock type.
+type RTCClock string
+
+// RTCDriftFix is the qemu RTC drift fix type.
+type RTCDriftFix string
+
+const (
+	// UTC is the UTC base time for qemu RTC.
+	UTC RTCBaseType = "utc"
+
+	// LocalTime is the local base time for qemu RTC.
+	LocalTime = "localtime"
+)
+
+const (
+	// Host is for using the host clock as a reference.
+	Host RTCClock = "host"
+
+	// VM is for using the guest clock as a reference
+	VM = "vm"
+)
+
+const (
+	// Slew is the qemu RTC Drift fix mechanism.
+	Slew RTCDriftFix = "slew"
+
+	// NoDriftFix means we don't want/need to fix qemu's RTC drift.
+	NoDriftFix = "none"
+)
+
 // RTC represents a qemu Real Time Clock configuration.
 type RTC struct {
 	// Base is the RTC start time.
-	Base string
+	Base RTCBaseType
 
 	// Clock is the is the RTC clock driver.
-	Clock string
+	Clock RTCClock
 
 	// DriftFix is the drift fixing mechanism.
-	DriftFix string
+	DriftFix RTCDriftFix
+}
+
+// Valid returns true if the RTC structure is valid and complete.
+func (rtc RTC) Valid() bool {
+	if rtc.Clock != "" {
+		if rtc.Clock != Host && rtc.Clock != VM {
+			return false
+		}
+	}
+
+	if rtc.DriftFix != "" {
+		if rtc.DriftFix != Slew && rtc.DriftFix != NoDriftFix {
+			return false
+		}
+	}
+
+	return true
 }
 
 // QMPSocket represents a qemu QMP socket configuration.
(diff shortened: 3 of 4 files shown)


markdryan:
@sameo Code looks good to me.  However, I noticed that some of the commit messages don't have a description.  I'm not sure if that's in violation of our contributing guidelines or not as the merge commit will have a description.

qemu/qemu.go
```diff
@@ -513,6 +534,37 @@ func appendKnobs(params []string, config Config) []string {
 	return params
 }
 
+func appendNetDevices(params []string, config Config) []string {
```
mcastelino: @sameo longer term the networking setup for VM's will look closer to https://github.com/01org/ciao/blob/master/ciao-launcher/qemu.go#L278 rather than the current tap format.

We need that in order to enable both vhost and mq.

The current format used for tap interfaces is the older deprecated one.

I have a PR pending in netlink where I am adding support of mq tap. 

I can change the current codebase to move to the newer format without mq.
That may help abstract the interface better.

sameo:
@mcastelino let me know if commit 27bd743 and commit 5857eb9 address your concerns.

markdryan:
This is okay for me.  I checked launcher and it can still start VMs.

qemu/qemu.go
```diff
@@ -45,8 +45,8 @@ type Machine struct {
 
 // Device describes a device to be created by qemu.
 type Device struct {
```
mcastelino: Most if not all devices today support vhost and multi queue capability. So at the device level we should specify if vhost is desired and how many queues are needed. Typically the caller can set the queues to number of vCPU's in guest.

The one challenge would be to figure out how the fd's created for all of the different devices map into the final QEMU invocation, where there are multiple devices with fd's.

qemu/qemu.go
```diff
@@ -191,6 +191,13 @@ type NetDevice struct {
 
 	// Script is the tap interface configuration script.
 	Script string
+
+	// FDs represents the list of already existing file descriptors to be used.
+	// This is mostly useful for mq support.
+	FDs []int
```
mcastelino: We need to figure out how to fold in fd's from multiple devices into the qemu invocation.. @markdryan had observed some fd number movement as fd's moved from parent to child.

ciao-launcher/qemu.go
```diff
@@ -358,7 +358,8 @@ func launchQemuWithNC(params []string, fds []*os.File, ipAddress string) (int, e
 		ncString := "socket,port=%d,host=%s,server,id=gnc0,server,nowait"
 		params[len(params)-1] = fmt.Sprintf(ncString, port, ipAddress)
 		var errStr string
-		errStr, err = qemu.LaunchQemu(context.Background(), params, fds, qmpGlogLogger{})
+
+		errStr, err = qemu.LaunchCustomQemu(context.Background(), "", params, fds, qmpGlogLogger{})
```
mcastelino: @sameo should the fd's be held within their own structure. As there may be multiple devices with fd's. Also we may use other fd's for parent-child communication for example.

qemu/qemu.go
```diff
@@ -26,14 +26,616 @@ package qemu
...
+		}
+
+		params = append(params, "-qmp")
+		params = append(params, strings.Join(qmpParams, ""))
+	}
+
+	return params
+}
+
+func appendDevices(params []string, config Config) []string {
```
mcastelino: @sameo appending devices in some cases may need appending to the kernel command line also. So we will probably need the ability to pass out any additional kernel params generated by the device append.

Note: The comment applies to other devices too like char devices and fs devices

qemu/examples_test.go
```diff
@@ -41,7 +41,7 @@ func Example() {
 	// LaunchQemu should return as soon as the instance has launched as we
```
tpepper: comment no longer matches function name

qemu/qemu.go
```diff
@@ -34,6 +34,28 @@ import (
 	"context"
 )
 
+// Device describes a device to be created by qemu.
+type Device struct {
+	// Type is the qemu device type
+	Type string
```
tpepper: Could we more strongly type these data members beyond string?

qemu/qemu.go
```diff
@@ -127,6 +152,41 @@ func appendQMPSocket(params []string, config Config) []string {
...
 }
 
+func appendDevices(params []string, config Config) []string {
+	for _, d := range config.Devices {
+		if d.Type != "" {
+			var deviceParams []string
+
+			deviceParams = append(deviceParams, fmt.Sprintf("%s", d.Type))
+
+			if d.ID != "" {
```
tpepper: With a stronger typing, rather than:

if d.ID != "" {
}

it could be something like:

if d.ID.IsValid() {
}

where the validator function knows how to check for a zero value, as well as any other appropriate formatting and range correctness.

qemu/qemu_test.go
```diff
@@ -236,3 +236,25 @@ func TestAppendQMPSocket(t *testing.T) {
...
+
+var qemuString = "-name cc-qemu -cpu host -uuid 123456789"
+
+func TestAppendStrings(t *testing.T) {
+	var params []string
+
+	config := Config{
+		Path:     "qemu",
+		Name:     "cc-qemu",
+		UUID:     "123456789",
```
tpepper: While this unit test tests the append*() function family, it reports success on an invalid UUID, which feels like a bad precedent to me.

tpepper:
Overall, I'm a little worried that (almost) everything is a string and if the offered string is not "" then onto the qemu command line it flows.  That's not very robust in itself.

But also, having spent some time in a prior job looking at a bug in the qemu internals around command line parsing, I wonder if this potentially huge complex list of parameters should have some vetting once composed.  The particular issue I'd been looking at back then was that "-numa node[,mem=size][,cpus=cpu[-cpu]][,nodeid=node]" could be nonsensical relative to -smp and -mem arguments earlier or later on the command line and very weird stuff happened.

The right answer to all of my concerns may be that this happens at the top of our Ciao command stack up in controller when we implement a workload validator?  But lacking that, I want to voice my concern that we've got a gap.

mcastelino:
@tpepper . Most if not all of there qemu parameters are internally generated. The question is, if any of these are coming in from the payload/workload. I do not think so, but it would be worth checking, specially for storage. Even then I think we generate the UUID and the user does not get to choose.

But scrubbing input makes sense at the top (and even launcher as it may know specifics about docker and qemu invocations which may be hard to scrub at the top level)
1. QEMU parameters embedded workload/payload data (e.g. someone sneaking in -device etc into the workload data)
2. general scrubbing for malformed values (e.g. UUIDs, Numbers, ranges etc)

markdryan:
For ciao, parts of the qemu parameters will come from the payload or be derived from it, e.g., the ids of the ceph volumes, the amount of memory to allocate to the VM.  Launcher already does quite a bit of checking here.  For example, it checks to see that the UUIDs of instances and volumes are indeed UUIDs. but there may be some more checking we can do.  I can review this code to make sure everything is validated in launcher. 

Note I don't think you can do a qemu injection attack, for example by doing,

instance_uuid: d7d86208-b46c-4465-9018-fe14087d415f -device ...

even if launcher wasn't validating the UUIDs, this sort of thing would be picked up and rejected by exec.Command.  I believe I tested this late last year.

ciao-launcher/qemu.go
markdryan: I think this is okay.  We just need to document how it works in the qemu package.  The slice of fds passed to this array will be passed directly to Command.Extrafiles.  This means that the fds received inside qemu will be derived from the location of the parent fds within this slice.  As launcher has control over the array of fds and all the other parameters that it will pass to qemu, it can ensure that the two are in sync, i.e., that the fd 3 that it passes to qemu in a parameter is the first fd in the fd slice.  As long we document this we should be okay.

markdryan:
> You add LaunchCustomQemu() which is mostly the old LaunchQemu(), and simplify a new LaunchQemu(). But the new LaunchQemu() doesn't appear called in the patch series? Seems like you'd want to use the simplified one since you went through the effort of making it?

This is being done in two stages.  LaunchQemu in it's current form only satisfies the needs of CC.  It's not yet flexible enough to support launcher.  To avoid breaking launcher @sameo has modified laucher to use LaunchQemuCustom.  I've independently tested that this works.  After this patch has been merged, I will submit a second patch that modifies launcher to use LaunchQemu.  This will be quite a lot of work as lots of the code that in launcher's qemu.go file will need to be moved to the qemu package and refactored.

qemu/qemu_test.go
sameo: The unit test is about testing that the generated string is valid. Validating the actual input is an orthogonal and more general issue and also out of this patch scope.
I can use a valid UUID for this test, but the overall input validation strategy needs to be defined at a higher level and for the project as a whole, not as a specific component or even patchset level.
In that specific case, the qemu parameters are derived from user input but not used directly as is. Can we do qemu parameter injection ? Maybe, but this needs to be audited and then we can decide *where* we do input validation. Doing that at the end of the code path, which is where this code lives, is probably not the right thing to do.

tpepper: Yep, as per my overall comment on the patch, we need to decide the right place(s?) to do meaningful validation.

qemu/qemu.go
```diff
@@ -26,14 +26,876 @@ package qemu
...
+	// Type is the machine type to be used by qemu.
+	Type string
+
+	// Acceleration are the machine acceleration options to be used by qemu.
+	Acceleration string
+}
+
+// Device is the qemu device interface.
+type Device interface {
+	Valid() bool
```
mcastelino: @sameo can the  qemu parameter generation for Device also be an interface?

qemu/qemu.go
```diff
@@ -26,14 +26,876 @@ package qemu
...
+
+	return true
+}
+
+// NetDeviceType is a qemu networing device type.
+type NetDeviceType string
+
+const (
+	// TAP is a TAP networking device type.
+	TAP NetDeviceType = "tap"
```
mcastelino: @sameo need to add macvtap here

qemu/qemu.go
```diff
@@ -26,14 +26,876 @@ package qemu
...
+)
+
+// NetDevice represents a guest networking device
+type NetDevice struct {
+	// Type is the netdev type (e.g. tap).
+	Type NetDeviceType
+
+	// ID is the netdevice identifier.
+	ID string
+
```
mcastelino: @sameo we should add number of (desired) queues here. >1 implies user desires multi queue.
We can also add a boolean property vhost.

Both these will be generic properties of multi queue capable devices/vhost capable devices.

qemu/qemu.go
```diff
@@ -26,14 +26,876 @@ package qemu
...
+		netdevParams = append(netdevParams, ",vhost=on")
+	}
+
+	params = append(params, "-netdev")
+	params = append(params, strings.Join(netdevParams, ""))
+
+	return params, nil
+}
+
+func appendDrivers(params []string, config Config) []string {
```
mcastelino: @sameo may make sense to also have a kernel append function also for each device. hence they can have two append interfaces. qemuAppend and kernelAppend.

qemu/qemu.go
```diff
@@ -42,6 +42,9 @@ type Config struct {
 	// Ctx is not used at the moment.
 	Ctx context.Context
 
+	// Name is the qemu guest name
+	Name string
+
```
tpepper: Surprised go fmt didn't autofix this, but you've got trailing whitespace here on this line.

ciao-launcher/qemu.go
mcastelino: @markdryan so the device creation order (and hence insertion into the slice) should match the append order.

tpepper:
One tiny thing I saw for a white space oddity, but otherwise and given all the conversation things look reasonable to me.

qemu/qemu.go
```diff
@@ -26,14 +26,876 @@ package qemu
...
+	Virtio9P = "virtio-9p-pci"
+
+	// VirtioNet is the virt-io networking device driver.
+	VirtioNet = "virtio-net"
+
+	// VirtioSerial is the serial device driver.
+	VirtioSerial = "virtio-serial-pci"
+
+	// Console is the console device driver.
+	Console = "virtconsole"
```
mcastelino: We should probably switch to the newer variant of the same
   -drive file="$IMAGE",if=none,id="$DRIVE",aio=threads,format=qcow2 \
   -device virtio-blk,drive="$DRIVER",scsi=off,config-wce=off \

qemu/qemu.go
sameo: Where is the trailing whitespace, I do not see it.

qemu/qemu.go
sameo: Good point. I just pushed an additional commit that adds `QemuParams()` to the Device interface and allows us to get rid of the Driver structure.
Later we will add `KernelParams()` to the device interface as well.

qemu/qemu.go
sameo: Done.

qemu/qemu.go
sameo: Done.

qemu/qemu.go
sameo: @mcastelino There already is a VHost boolean. Also, is the number of queues != len (NetDevice.FDs) ?

ciao-launcher/qemu.go
sameo: @mcastelino I changed that so that each device can bring its own set of fds. Config.FDs is gone and each device QemuParams() call should append its fds to the overall fd slice.
See commit #c1a5e0177bb04392e737c7dc757b7a3f3423cead.

qemu/qemu.go
mcastelino: @sameo regarding the vhost. I meant having a top level boolean as most devices support vhost.
Also regarding the number of queues, what I was suggesting was that the number of queues is an input, the numbers of fds (or the fd's in the param outout) is the output. Similarly for vhost.

qemu/qemu.go
tpepper: Line 47 was showing a tab.  Either you found it or on some other edit/save goftmt removed it?  It's not there in what merged.

PR #551, opened 2016-09-12
ciao-image: Add delete image support

obedmr:
This commit adds support for deleting images. Now, you will be able to
delete images with the command:
```
  ciao-cli image delete -image 78306709-9720-417e-97c6-7b941cb5565b
```
If the operation was successful, you will get a 204 HTTP return code.
If not, it may return the proper error, for example a 404 error could
come when the image does not exist.

### DIFF
--- a/openstack/image/api.go
+++ b/openstack/image/api.go
@@ -16,6 +16,7 @@ package image
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -132,6 +133,17 @@ const (
 	ISO DiskFormat = "iso"
 )
 
+// ErrorImage defines all possible image handling errors
+type ErrorImage error
+
+var (
+	// ErrNoImage is returned when an image is not found.
+	ErrNoImage = errors.New("Image not found")
+
+	// ErrImageSaving is returned when an image is being uploaded.
+	ErrImageSaving = errors.New("Image being uploaded")
+)
+
 // CreateImageRequest contains information for a create image request.
 // http://developer.openstack.org/api-ref-image-v2.html#createImage-v2
 type CreateImageRequest struct {
@@ -181,9 +193,10 @@ type ListImagesResponse struct {
 	First  string            `json:"first"`
 }
 
-// UploadImageResponse contains the UUID of the image which content got uploaded
+// NoContentImageResponse contains the UUID of the image which content
+// got uploaded or deleted
 // http://developer.openstack.org/api-ref-image-v2.html#storeImageFile-v2
-type UploadImageResponse struct {
+type NoContentImageResponse struct {
 	ImageID string `json:"image_id"`
 }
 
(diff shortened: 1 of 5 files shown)


markdryan:
@obedmr Code looks okay to me.  Could you add a description to your commit message?

@sameo, @kaccardi is this one okay for you?

openstack/image/api.go
```diff
@@ -362,6 +378,19 @@ func uploadImage(context *Context, w http.ResponseWriter, r *http.Request) (APIR
...
+
+	defer r.Body.Close()
+
+	resp, err := context.DeleteImage(imageID)
+	if err != nil {
+		return errorResponse(err), err
+	}
+	return APIResponse{http.StatusNoContent, resp}, nil
+}
+
```
kaccardi: You are returning StatusNoContent - but you are returning a response. StatusNoContent should be used when there is no response, and the openstack spec says that /v2/images/{image_id} DELETE should return no response. http://developer.openstack.org/api-ref/image/v2/index.html?expanded=delete-an-image-detail#delete-an-image

openstack/image/api.go
```diff
@@ -201,9 +214,10 @@ type APIConfig struct {
 // information needed to implement the image endpoints.
 type Service interface {
 	CreateImage(CreateImageRequest) (DefaultResponse, error)
-	UploadImage(string, io.Reader) (UploadImageResponse, error)
+	UploadImage(string, io.Reader) (NoContentImageResponse, error)
 	ListImages() ([]DefaultResponse, error)
```
kaccardi: I know this isn't your bug - but I see that UploadImage also returns a response, when it should also return no response as indicated by the api spec.
http://developer.openstack.org/api-ref/image/v2/index.html?expanded=delete-an-image-detail,upload-binary-image-data-detail#upload-binary-image-data

openstack/image/api.go
obedmr: @kaccardi, you're right, good catch. /me updating it

openstack/image/api.go
obedmr: /me fixing it

obedmr:
@kaccardi, @markdryan @sameo, it's updated. When you have a chance could you take a look on it?

openstack/image/api.go
```diff
@@ -253,6 +267,8 @@ func (h APIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // on return values all the time.
 func errorResponse(err error) APIResponse {
 	switch err {
+	case ErrNoImage:
+		return APIResponse{http.StatusNotFound, nil}
 	default:
```
kaccardi: Shouldn't there be a response for ErrImageSaving? I would think this would be a 4xx response and not a 5xx.

obedmr: @kaccardi, this error is going to be used when you're using the Get method. If the image is not Found, it's getting a 404 error, which is the http.StatusNotFound 
https://golang.org/pkg/net/http/
```
  (StatusNotFound                     = 404 // RFC 7231, 6.5.4)
```

The idea of this errorResponse function (AFAIK) is to be the one for handling the different HTTP errors. For now those are the implemented ones, expect more to be added.

kaccardi: @obedmr ErrImageSaving is returned from UploadImage - which is a PUT method. This error is encountered if the state of the image is already in "Saving" when someone tries to do another upload. Correct?

obedmr: yes, this case should be handled differently.

obedmr:
@kaccardi, when you have a chance, could you take a look on it? I just added the missing test.

Issue #17, opened 2016-04-08
Check container instance memory use is correct when container has multiple processes


markdryan:
ciao-launcher returns the pss of the process associated with a docker instance in the stats command.  It's not really clear what happens if the container has multiple processes.  I'm guessing that launcher does not return the correct memory usage for such containers.  This needs to be checked and fixed if it is indeed a problem.

PR #553, opened 2016-09-13
ciao-launcher: Use docker ContainerStats api to get memory and CPU usage

rbradford:
When there were multiple processes running inside the container the
reported memory and CPU usage values were incorrect as they only
reflected the first process created in the container.

This change makes use of the Docker ContainerStats API to obtain the
memory and CPU usage for a container.

This was tested by comparing the results from "ciaolc istats" with
"docker stats".

Fixes: #17

### DIFF
--- a/ciao-launcher/docker.go
+++ b/ciao-launcher/docker.go
@@ -374,16 +374,34 @@ func (d *docker) stats() (disk, memory, cpu int) {
 	memory = -1
 	cpu = -1
 
-	if d.pid == 0 {
+	if d.pid == 0 || d.cfg == nil {
 		return
 	}
 
-	memory = computeProcessMemUsage(d.pid)
-	if d.cfg == nil {
+	cli, err := getDockerClient()
+	if err != nil {
+		glog.Errorf("Unable to get docker client: %v", err)
 		return
 	}
 
-	cpuTime := computeProcessCPUTime(d.pid)
+	resp, err := cli.ContainerStats(context.Background(), d.dockerID, false)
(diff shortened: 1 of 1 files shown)


ciao-launcher/docker.go
```diff
@@ -374,16 +374,33 @@ func (d *docker) stats() (disk, memory, cpu int) {
...
+
+	cli, err := getDockerClient()
+	if err != nil {
+		glog.Errorf("Unable to get docker client: %v", err)
 		return
 	}
 
-	memory = computeProcessMemUsage(d.pid)
-	if d.cfg == nil {
+	resp, err := cli.ContainerStats(context.Background(), d.dockerID, false)
```
markdryan: According to the documentation for this function we need to close the reader.

markdryan: See

https://github.com/01org/ciao/blob/master/vendor/github.com/docker/engine-api/client/container_stats.go#L11

markdryan:
Also I think we might be able to remove some code from docker.go now.  There's this pid member, which may not be needed any more.  I think it was only used to compute the container's stats.  There's also a pile of code in connected that initialises it.  Can you confirm pid is now unnecessary and, if so, remove it and the code that uses it?

ciao-launcher/docker.go
rbradford: Done with a defer.

rbradford:
@markdryan PID cleanup in follow-up patch on this PR.

PR #558, opened 2016-09-15
Networking: Enable vhost for instances

mcastelino:
- Enable vhost support for instances
- Unify the qemu invocation between macvtap and tap


markdryan:
Looks fine to me but I haven't tested on a full cluster.

PR #568, opened 2016-09-15
testing: add coverage for ciao-cert

tpepper:
Actually there's no coverage, based on inspection, but coveralls doesn't
notice because we don't try to run the nothing that's there.


tpepper:
This was mostly just a test and coveralls did not pass the test.  Coveralls appears blind to a package having no *_test.go files while also having *.go files.

Issue #572, opened 2016-09-16
dual node launchers install their dependencies twice


markdryan:
If launcher is run in dual mode, i.e., the launcher can launch both compute and cnci instances, osprepare tries to install the compute dependencies twice.  This doesn't create any real problems of course but it does slow down the start up time of dual node launchers.  We should probably fix this.

Issue #573, opened 2016-09-16
Rename osprepare main.go to something else to avoid confusion in the logs


markdryan:
The clients of osprepare use glog to perform their logging.  Glog prints the name of the file and the file number of the line of code on which the log statement appeared.  The problem is that it doesn't print any package information.  So when launcher calls osprepare functions I see something like this.

```
I0916 14:45:19.038576    5674 main.go:361] Ceph ID:              ciao
I0916 14:45:19.038900    5674 main.go:177] OS Detected: ubuntu
```

The main.go in the first line is launcher's.  The main.go in the second line is os-prepare's.  ciao-controller will suffer from the same problems.  Strictly speaking this isn't really os-prepares fault, but it's probably easier to make the problem go away by renaming main.go.  It's probably not a great name for a package that isn't an exe anyway.

Issue #500, opened 2016-08-30
osprepare prevents launcher from exitting cleanly


markdryan:
osprepare.distro.InstallPackages needs to take a context parameter as its first argument so it can be cancelled.  It initiates a long running process that blocks the calling thread until it finishes.  This prevents launcher and the other ciao components from shutting down cleanly when asked to exit.

tpepper:
I've got a patch which I think does the right thing, but need to test.

markdryan:
Does the patch actually quit the installer?  I ask as I tried this today but I forgot to set my proxies so  nothing got installed.  When I tried to quit, launcher quit after 1 second (in an unclean fashion) but the apt-get install process launched by osprepare was still running and still outputting stuff to my terminal, which made it very hard to use.

tpepper:
I'm using the context.WithCancel() as in your existing code, so if it's go's not actually quitting the sub-exec, then your existing calls are similarly persisting?

Issue #511, opened 2016-09-01
osprepare: Fail to install bundles in clearlinux behind a proxy


albertomurillo:
osprepare is ran by the launcher to detect if there is a missing pacakge (qemu, docker, xorriso, etc)
and install the required bundle (cloud-control).

This task fails when running behind a proxy.

This is not an issue in ubuntu and fedora because you can configure the proxies in /etc/apt/apt.conf or /etc/dnf/dnf.conf

tpepper:
Since osprepare has code for ClearLinux, Fedora and Ubuntu, and you say the latter two are ok, this sounds like it is a ClearLinux specific issue.  I thought ClearLinux has autoproxy, or a way to configure proxies manually for a system so system software can access a network?

I'm not sure how much osprepare itself can do about this.

albertomurillo:
sudoFormatCommand() is not honoring environment variables from /etc/profile
making osprepare use env vars rom /etc/profile can fix the issue in clearlinux.

tpepper:
I've got a patch which I think does the right thing, but need to test.

tpepper:
Upon further code inspection withing golang, we're pretty sure this patch is actually unneeded.  It makes explicit grabbing the current Env, but if none was given (the state before the patch), the golang runtime pulls in the current Env.

I can only guess that the env was actually wrong (eg: maybe sudo didn't pass the user proxied env through to the root env running the ciao binary?)...in which case this bug is invalid?  But I'm guessing and assuming a lot here.  @albertomurillo if you can confirm how you're running things in more detail, eg: where is the proxy env var specified, what's the ciao command line and sudo arg's, etc., we can figure out more concretely what's gone wrong.

albertomurillo:
I think this bug is invalid too.

The reason ciao-launcher is not getting the proxy env variables is because we are not starting ciao-launcher from the command line but from a systemd unit file.

I found that the admin can create /etc/systemd/system/ciao-compute.service.d/proxy.conf file
(docker style) to make ciao-launcher start with proxy environment values.

tpepper:
Based on the last two comments I'm closing this as invalid, worked around in systemd standard way.

PR #578, opened 2016-09-16
osprepare: bits of bug fixing and extending

tpepper:
This PR fixes issues #572, #573, and #500 with relatively simple tweeks.  Still to come is proper minimum version checking and fallback to non-distro sources of package dependencies.

Note: the diffstat is worse than it arguably is because one of those issues is a git mv of osprepare/main.go to osprepare/osprepare.go, and the non-issue specific second commit is a large gorename patch.


ciao-launcher/main.go
```diff
@@ -237,14 +236,36 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+		}
+
+		ch <- nil
+	}()
+
+	select {
+	case <-doneCh:
+		glog.Info("Received terminating signal.  Installation of launcher dependencies cancelled.")
+		cancelFunc()
+		return
```
rbradford: Spurious return?

ciao-launcher/main.go
```diff
@@ -237,14 +236,36 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+	case err := <-ch:
+		cancelFunc()
+		if err != nil {
+			glog.Errorf("Failed to install launcher dependencies: %v\n", err)
+			return
```
rbradford: Ditto: spurious return

ciao-launcher/main.go
tpepper: Simplified in amended patch.

ciao-launcher/main.go
tpepper: Simplified in amended patch.

ciao-launcher/main.go
```diff
@@ -237,15 +236,35 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+	select {
+	case <-doneCh:
+		glog.Info("Received terminating signal.  Installation of launcher dependencies cancelled.")
+	case err := <-ch:
+		cancelFunc()
+		if err != nil {
+			glog.Errorf("Failed to install launcher dependencies: %v\n", err)
+		}
 	}
+	cancelFunc()
```
rbradford: On the success and err path you end up calling cancelFunc() twice. Drop the one at line 262?

osprepare/distro.go
```diff
@@ -229,6 +229,7 @@ func sudoFormatCommand(command string, packages []string) bool {
 	c := exec.Command(executable, strings.Split(args, " ")...)
 	c.Stdout = os.Stdout
 	c.Stderr = os.Stderr
+	c.Env = os.Environ()
```
tpepper: Good question.  The original bug report didn't have enough details for me to understand how their system was configured.

The "If Env is nil, Run uses the current process's environment." comment and if nil return os.Environ() has been in os/exec/exec.go since 2011.

@albertomurillo can you confirm how you're running and failing in issue #511 and whether this patch resolves for you?  Based on @rbradford's archaeology, we're thinking this patch doesn't actually change anything.

ciao-launcher/main.go
```diff
@@ -241,17 +241,20 @@ func (client *agentClient) installLauncherDeps(doneCh chan struct{}) {
 
 	ch := make(chan error)
 	go func() {
-
 		osprepare.Bootstrap(ctx)
 
+		launcherDeps := make(osprepare.PackageRequirements)
```
markdryan: This line is a little confusing.  At first I didn't think it was valid go until I realised that PackageRequirements was a map.  It might be better to create a constructor function for this type.  This way launcher doesn't need to know what the underlying implementation is.  Having said that I can't find any best practices document suggesting that we shouldn't do this.  So feel free to ignore if you disagree.

ciao-controller/main.go
```diff
@@ -143,8 +144,8 @@ func main() {
 		*cephID = clusterConfig.Configure.Storage.CephID
 	}
 
-	osprepare.Bootstrap()
-	osprepare.InstallDeps(controllerDeps)
+	osprepare.Bootstrap(context.TODO())
+	osprepare.InstallDeps(context.TODO(), controllerDeps)
```
markdryan: nice usage of TODO!

ciao-launcher/main.go
markdryan: It's okay to call cancelFunc twice.  Often it's hard to avoid.

https://golang.org/pkg/context/#CancelFunc

ciao-launcher/main.go
```diff
@@ -237,15 +236,35 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+		ch <- nil
+	}()
+
+	select {
+	case <-doneCh:
+		glog.Info("Received terminating signal.  Installation of launcher dependencies cancelled.")
+	case err := <-ch:
+		cancelFunc()
+		if err != nil {
+			glog.Errorf("Failed to install launcher dependencies: %v\n", err)
```
markdryan: Right now we'll never execute this line of code.

ciao-launcher/main.go
tpepper: What the heck...I thought I deleted both.  My bad.  Re-amended...

ciao-launcher/main.go
```diff
@@ -237,15 +236,35 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+	ch := make(chan error)
+	go func() {
+
+		osprepare.Bootstrap(ctx)
+
+		role := client.conn.Role()
+		if role.IsNetAgent() {
+			osprepare.InstallDeps(ctx, launcherNetNodeDeps)
+		}
+		if role.IsAgent() {
```
markdryan: If we have a launcher running in dual mode and we cancel the context during the first call to InstallDeps, we'll then try to call InstallDeps again, which will parse a file to determine the distro.  The easiest thing to do would be to return an error from InstallDeps if there is a cancel or a timeout.  There's a special error code for this I think.  I bet CommandContext returns it.  The alternative would be to select on the  context.Done() channel here.

ciao-launcher/main.go
```diff
@@ -237,15 +236,35 @@ func (client *agentClient) ErrorNotify(err ssntp.Error, frame *ssntp.Frame) {
...
+			osprepare.InstallDeps(ctx, launcherComputeNodeDeps)
+		}
+
+		ch <- nil
+	}()
+
+	select {
+	case <-doneCh:
+		glog.Info("Received terminating signal.  Installation of launcher dependencies cancelled.")
```
markdryan: The way the code is written, there is no guarantee that the inner go routine we have spawned has exitted by the time we exit installLauncherDeps.  We can ensure this by doing


```
case <-doneCh:
  cancelFunc()
  <-ch
```
If we do this we don't need the call to cancelFunc at the bottom.

ciao-launcher/main.go
tpepper: see newer version

ciao-launcher/main.go
tpepper: see newer version

ciao-launcher/main.go
markdryan: It looks the same.

ciao-launcher/main.go
tpepper: Does that leak the cancel?  Or would I need to re-add it to the "case err := <-ch:".

And I'm unclear about how the cancelFunc would work relative to the read of the channel.  Will the cancelFunc() result in ch being closed, so the subsequent "<-ch" will return zero immediately if the deps install was cancelled or return nil if the deps install had actually gotten to its "ch <- nil"?

markdryan: Right.  You still need the other case statement.  So

case <-doneCh:
  cancelFunc()
  <-ch
case err := <- ch:
  cancelFunc()
}
```

Comment (markdryan) [inline ciao-launcher/main.go]: cancelFunc has not effect on ch.  However, by cancelling the context you're causing CommandContext to exit early and hence InstallDeps to return.  This will result it the inner go routine writing to the ch channel.  When it does that we now its safe to exit.

so

```
case <-doneCh:
  cancelFunc()
  <-ch
```
says cancel the long running function and wait for the cancel to finish.

ciao-launcher/main.go
tpepper: Should show two calls to launcherDeps.Append() and one call to InstallDeps().

ciao-launcher/main.go
tpepper: added this in latest rev

markdryan:
@tpepper This is working great.  I uninstalled one of the packages needed by launcher, ran launcher and did a ctrl-c while it was installing.  Everything quit cleanly.

markdryan:
> Can we drop the environment patch before merging? I'm really not convinced it adds anything useful - okay it does no harm either but it won't fix the bug :-)

Agreed.  Let's drop this.

tpepper:
I've dropped the env patch and taken issue #511 out of the PR description.  Adding commentary over in that issue as to why.

markdryan:
@tpepper Is there any reason not to merge this?

tpepper:
Only reason not to have merged it is it keeps needing rebased...I'm rebasing daily, but this PR's ready. Other will follow for version and alt repo handling.

markdryan:
Looks good to me.  Tested the rebased version locally and it's still working.

PR #579, opened 2016-09-16
Updated ssntp/README.md added link to ciao-cert

pdxjohnny:
Added a link to the ciao-cert repo in the ssntp README to provided
instructions on certificate generation for ssntp.

Signed-off-by: John Andersen <john.s.andersen@intel.com>

### DIFF
--- a/ssntp/README.md
+++ b/ssntp/README.md
@@ -97,6 +97,11 @@ UUIDs.
 3. Connection is successfully established. Both ends of the connection
    can now asynchronously send SSNTP frames.
 
+## SSNTP certificates ##
+
+SSNTP uses ciao-cert to generate the certificates it needs to communicate. They
+can be generated with instructions found in [ciao-cert] (https://github.com/01org/ciao/tree/master/ciao-cert).
+
 ## SSNTP frames ##
 
 Each SSNTP frame is composed of a fixed length, 8 bytes long header and


ssntp/README.md
```diff
@@ -97,6 +97,31 @@ UUIDs.
...
 
+## CIAO Cert ##
+
+SSNTP uses ciao-cert to generate the certificates it needs to communicate. They
+can be generated as follows.
+
+```
+sudo mkdir -p /etc/pki/ciao/
+sudo mkdir -p /var/lib/ciao/logs/scheduler
+sudo -E $GOPATH/bin/ciao-cert -directory /etc/pki/ciao -host localhost -server -role scheduler
```
sameo: There are more options to ciao-cert and this is only one example. So I would not include those into the overall SSNTP readme but rather somewhere in the wiki.
Also, there are a few examples here as well: https://github.com/01org/ciao/tree/master/ciao-cert

tpepper: I agree.

It will be easier to maintain if ssntp docu notes that it depends on certificates with roles embedded and links to the docu on creating those, eg:

https://github.com/01org/ciao/blob/master/ciao-cert/README.md
https://godoc.org/github.com/01org/ciao/ciao-cert

(note the second of those is empty...maybe that's the place to put some attention)

pdxjohnny:
Good idea I will change it so it links to that existing documentation. The
only reason I didn't do that originally was because I had not come across it

pdxjohnny:
I updated so that it links to the ciao-cert repo

Issue #580, opened 2016-09-19
Testutils unit tests hang occassionally.


markdryan:
This hang occurred on a go tip build.

```
testutil           TestTenantAdded                          0.00s      PASS    

testutil           TestReconnects                           N/A        NOT RUN 

testutil           TestTenantRemoved                        N/A        NOT RUN 

testutil           TestPublicIPAssigned                     N/A        NOT RUN 

testutil           TestNewSsntpTestControllerConnectionArgs N/A        NOT RUN 

testutil           TestControllerErrorChan                  N/A        NOT RUN 

testutil           TestControllerErrorChanTimeout           N/A        NOT RUN 

testutil           TestControllerEventChan                  N/A        NOT RUN 

testutil           TestControllerEventChanTimeout           N/A        NOT RUN 

testutil           TestControllerCmdChan                    N/A        NOT RUN 

testutil           TestControllerCmdChanTimeout             N/A        NOT RUN 

testutil           TestControllerCloseChans                 N/A        NOT RUN 

testutil           TestStartIdentityServer                  N/A        NOT RUN 

testutil           TestServerStatusChan                     N/A        NOT RUN 

testutil           TestServerStatusChanTimeout              N/A        NOT RUN 

testutil           TestServerErrorChan                      N/A        NOT RUN 

testutil           TestServerErrorChanTimeout               N/A        NOT RUN 

testutil           TestServerEventChan                      N/A        NOT RUN 

testutil           TestServerEventChanTimeout               N/A        NOT RUN 

testutil           TestServerCmdChan                        N/A        NOT RUN 

testutil           TestServerCmdChanTimeout                 N/A        NOT RUN 

testutil           TestServerCloseChans                     N/A        NOT RUN 

testutil           TestRoleToTestCert                       N/A        NOT RUN 
```

And here are the stack traces

```
SIGABRT: abort

PC=0x45d151 m=0

goroutine 0 [idle]:

runtime.futex(0x8b65d0, 0x0, 0x0, 0x0, 0x0, 0x7f2a00000000, 0x0, 0x0, 0x7ffda061a168, 0x40eeab, ...)

	/home/travis/.gimme/versions/go/src/runtime/sys_linux_amd64.s:387 +0x21

runtime.futexsleep(0x8b65d0, 0x0, 0xffffffffffffffff)

	/home/travis/.gimme/versions/go/src/runtime/os_linux.go:45 +0x62

runtime.notesleep(0x8b65d0)

[...]

PR #582, opened 2016-09-19
[DO NOT MERGE] testutil: increase Results channel timeouts

tpepper:
Back in commit 0e16917936eabe454ada1df8bc0bc601d4a4a954 I rationlized why
we sometime see timeouts on the testutil Results channels, but took a stab
at a reasonable timeout.  In the meantime we have a considerable amount of
travis failures where, eg:

ciao-scheduler     TestReconnects                           26.03s     FAIL
ciao-scheduler     TestTenantAdded                          25.00s     FAIL
ciao-scheduler     TestTenantRemoved                        25.00s     FAIL

we clearly hit the 25s timeout.

Therefore this patch bumps the timeout to 60 seconds.  If it keeps causing
problems, I'll just remove the timeout completely.  This is purely test
code, and go-test does have it's own hang detection.

Fixes https://github.com/01org/ciao/issues/580

### DIFF
--- a/testutil/agent.go
+++ b/testutil/agent.go
@@ -124,7 +124,7 @@ func (client *SsntpTestClient) GetCmdChanResult(c *chan Result, cmd ssntp.Comman
 		if result.Err != nil {
 			err = fmt.Errorf("Client error sending %s command: %s\n", cmd, result.Err)
 		}
-	case <-time.After(25 * time.Second):
+	case <-time.After(60 * time.Second):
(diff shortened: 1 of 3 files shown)


tpepper:
So wouldn't you know it:  The first CI run with this still had a timeout at 60s.

https://travis-ci.org/tpepper/ciao/jobs/161103970

I need to pour through the log and re-check I haven't missed an actual race in this test case.

PR #583, opened 2016-09-20
ciao-storage: Use image-format 2

### LINKED ISSUE #528: Ciao is using a deprecated --image-format
The ceph storage driver currently specifies the "--image-format", "1" option when creating new ceph volumes.  The reason it does this is that we have problems mapping rbd images that used image-format 2 on some of our Ubuntu dev machines.  image-format 1 is deprecated so we should move back to image-format 2.

Changing to image-format 2 may require us to make some manual modifications to any ceph clusters running Ubuntu.

albertomurillo:
rbd image format 1 is deprecated but image
format 2 provides different image features
that are not compatible with the kernel rbd client.

This change uses image-format 2 and the layering
image feature which is the only image feature supported
by the rbd kernel client.

Fixes #528

### DIFF
--- a/ciao-storage/ceph.go
+++ b/ciao-storage/ceph.go
@@ -41,11 +41,16 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 
 	var cmd *exec.Cmd
 
+	// imageFeatures holds the image features to use when creating a ceph rbd image format 2
+	// Currently the kernel rdb client only supports layering but in the future more feaures
+	// should be added as they are enabled in the kernel.
+	imageFeatures := "--image-feature layering"
+
 	if imagePath != nil {
-		cmd = exec.Command("rbd", "--keyring", d.SecretPath, "--id", d.ID, "--image-format", "1", "import", *imagePath, ID)
+		cmd = exec.Command("rbd", "--keyring", d.SecretPath, "--id", d.ID, imageFeatures, "import", *imagePath, ID)
 	} else {
 		// create an empty volume
-		cmd = exec.Command("rbd", "--keyring", d.SecretPath, "--id", d.ID, "--image-format", "1", "create", "--size", strconv.Itoa(size), ID)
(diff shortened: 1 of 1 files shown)


markdryan:
@albertomurillo Thanks for looking at this.

Initially I had thought that the issue was just with Ubuntu.  Are you seeing it with other distros as well?  I don't really understand what is meant by the kernel rbd client.  If it's just Ubuntu specific maybe this isn't the right fix.

markdryan:
Are you saying it's a limitation of the rbd tool itself?

albertomurillo:
There are several ways to mount a rados block device (rbd) to a linux hosts

1. Using the kernel client (rbd module) [1]
2. Using the userspace tool rbd-fuse [2]
3. Using rbd-ndb userspace tool that maps an rbd to a kernel nbd block device. [3][4]

The limitation is on the kernel driver which is not updated as often as the userspace
tools.

[1] http://docs.ceph.com/docs/master/man/8/rbd/
[2] http://docs.ceph.com/docs/master/man/8/rbd-fuse/
[3] http://docs.ceph.com/docs/master/man/8/rbd-nbd/
[4] https://www.sebastien-han.fr/blog/2016/04/04/ceph-jewel-preview-map-rbd-devices-on-nbd/

markdryan:
@albertomurillo Thanks.  I'll check out the links and test tomorrow.

markdryan:
@kaccardi What do you think about this?  From what I read @albertomurillo is correct.  The only image-format 2 feature currently supported in any released kernels is layering, so using image-format 2 with all features enabled isn't going to work on any OS.   The problem with this patch though is that in two years time when all major distros support all these features, our code will explicitly disable them.

The other option would be to fix this by disabling the unsupported features in our ceph configs on each node but this would complicate our setup.

I'm inclined to say let's accept this patch.  It's better than what we currently have.  We can revisit the issue in a couple of years when the rbd kernel modules catch up with the client.

I haven't tested the patch itself but I did check that I could map an image-format=2 volume when I specified --image-feature layering on the command line.

PR #586, opened 2016-09-20
Single VM: Fix dev environment

### LINKED ISSUE #518: ciao-cli instance delete --all does not remove network interfaces
After ciao-cli instance delete --all on a node the network interfaces created for the node are left behind.
This seems to be regression. This can be tested with the single vm script which tears down all the instances.
Nothing has changed in networking in a while.

markdryan:
This PR updates the singlevm scripts so that they can be used as a development environment.  In short:

1. The cluster is no longer torn down when the setup.sh exits.  Instead all we do is to delete the test instances created by setup.sh.  The cluster remains up and functional.
2. Some information is printed out when the script ends informing the user that he needs to source a file containing some environment variables needed to communicate with ciao and also to teardown the cluster.
3. The cleanup.sh script has been updated to correctly teardown the cluster.

The PR also fixes #518 which left bridges and tap devices hanging around.

### DIFF
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -1,7 +1,18 @@
 #!/bin/bash
 
+if [ -z "$HOSTS_FILE_BACKUP" ]; then
+    echo "HOSTS_FILE_BACKUP" is not set.
+    echo "Please run"
+    echo ""
+    echo ". ~/local/demo.sh"
+    exit 1
+fi
+ciao_gobin="$GOPATH"/bin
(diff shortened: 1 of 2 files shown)


markdryan:
@mcastelino Could you take a look at these changes?  I'm able to use the single vm to run my own tests and clean up nicely with these patches.  If the patches are merged I'll update the wiki.

mcastelino:
@markdryan LGTM. Will do a quick test on my end and will merge it.

mcastelino:
@markdryan tested on Ubuntu 14.04 and works fine

PR #588, opened 2016-09-20
Enable TLS verification between cli and controller

rbradford:
This PR fixes "singlevm" test mode to have valid certificates between cli and controller - a prerequisite for enabling TLS verification.

It then goes ahead and enables TLS verification between ciao-cli and ciao-controller.


mcastelino:
@rbradford LGTM. Will verify and merge

PR #594, opened 2016-09-22
test-cases: Fix package list

markdryan:
test-cases contained some uncessary code for handling the list of packages
passed to it on the command line.  This code was unnecessary as go list
already does this processing for us, and it does it better.  So this patch
simply removes the custom handling, and test-cases now behaves like all
the other go tools, which is what users expect.

The commit also updates the usage message to explain how to specify packages
on the test-cases command line.


markdryan:
@bryteise I saw you had some questions about test-cases on irc last night.  This PR should clear up the confusion.  When merged, test-cases will behave like all the other go tools, when processing package lists.  Let me know if it's still not working for you.

bryteise:
@markdryan that clears things up nicely, thanks!

PR #596, opened 2016-09-22
documentation: enable swagger to use vendoring

leoswaldo:
Swagger is a tool used to generate documentation based on
declarative comments in the code, the problem with it was that it
had an issue with vendoring. We have fixed that issue in
https://github.com/yvasiyarov/swagger/pull/120
and now we can use our vendor directory. This fix was need because
we were bringing dependencies with 'go get' which in some PRs were
causing to fail for time outs, an issue aside of the PRs.

In this change we add the required argument to enforce to check for
vendor directory and remove the 'go get'.


markdryan:
Looks good to me.  Bet this speeds up our build a bit as well.

PR #598, opened 2016-09-22
Revert "Enable TLS verification between cli and controller"

mcastelino:
Reverts 01org/ciao#588


mcastelino:
@kaccardi @rbradford This PR seems to cause issues with the cluster setup. Reverting this for now till we root cause it. 
Note: I did not test on a full cluster before I merged this change in.

rbradford:
I tested this behaviour against a cluster build using the new ansible method. There must be something different between that and the existing test clusters.

PR #600, opened 2016-09-26
Merge supplied command line CA file with system pool

rbradford:
When enabling TLS, which this PR also does, this should allow configurations where keystone and controller cert are different (with one being in system store and the other on the command line.)

Tested against singlevm and against CNW cluster.


mcastelino:
@rbradford @kaccardi looks like merging this PR broke BAT. Can you look at it please.

Issue #591, opened 2016-09-21
launcher sometimes hangs on startup


markdryan:
Sometimes, maybe about 1 in 20 times, I can get launcher to hang on startup.  When I quit it it does not quit cleanly.  Here are the stack trace and logs.

```
sudo ./ciao-launcher --logtostderr -cert tests/ciao-launcher-server/cert-client-localhost.pem -cacert tests/ciao-launcher-server/CAcert-server-localhost.pem --with-ui=nc  -v=2 
I0920 04:28:10.437134   23977 main.go:612] Starting Launcher
I0920 04:28:10.437412   23977 main.go:547] Initial nofile limits: cur 1024 max 65536
I0920 04:28:10.437520   23977 main.go:559] Updated nofile limits: cur 65536 max 65536
I0920 04:28:10.437524   23977 main.go:630] Launcher will allow a maximum of 13107 instances
I0920 04:28:10.444279   23977 ssntp.go:889] SSNTP Info: c2c07791-7b45-4c0f-a26f-aa3d232541ef connecting to localhost:8888
I0920 04:28:10.475788   23977 ssntp.go:889] SSNTP Info: Connected
I0920 04:28:10.476455   23977 ssntp.go:889] SSNTP Info: Sending CONNECT
I0920 04:28:10.477653   23977 ssntp.go:889] SSNTP Info: Waiting for CONNECTED
I0920 04:28:10.478964   23977 ssntp.go:889] SSNTP Info: Received CONNECTED frame:
	Major 0
	Minor 1
	Type STATUS
	Op CONNECTED
	Role Scheduler-
	Source 8c949471-8427-4ce4-9c14-f5c0c12451bc
	Destination c2c07791-7b45-4c0f-a26f-aa3d232541ef

I0920 04:28:10.481664   23977 ssntp.go:889] SSNTP Info: Done with connection
I0920 04:28:10.482348   23977 main.go:145] connected
I0920 04:28:10.482366   23977 ssntp.go:889] SSNTP Info: Waiting for next frame
^CI0920 04:28:10.781203   23977 main.go:577] Received terminating signal.  Waiting for server loop to quit
W0920 04:28:11.781862   23977 main.go:587] Server Loop did not exit within 1 second quitting
panic: Server Loop did not exit within 1 second quitting

goroutine 1 [running]:
panic(0x7a8d20, 0xc420100350)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
main.startLauncher(0x0)
[...]

PR #601, opened 2016-09-26
Fix launcher deadlock

rbradford:
This change addresses the deadlock in #591. Tested via singlevm & and ciao-launcher-server/ciaolc

### DIFF
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"runtime/debug"
 	"sync"
 	"syscall"
 	"time"
@@ -563,7 +564,7 @@ DONE:
 			glog.Flush()
 
 			/* We panic here to see which naughty go routines are still running. */
-
+			debug.SetTraceback("all")
 			panic("Server Loop did not exit within 1 second quitting")
 		}
 	}
--- a/ciao-launcher/main.go
+++ b/ciao-launcher/main.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


ciao-launcher/main.go
```diff
@@ -390,50 +389,51 @@ func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
...
-			if err != nil {
-				glog.Errorf("Failed to start network: %v\n", err)
-				client.conn.Close()
-				break DONE
-			}
-			defer shutdownNetwork()
+		ovsCh = startOverseer(&wg, client)
+	case <-doneCh:
+		client.conn.Close()
+		<-dialCh
```
markdryan: I think we need a return here.  Otherwise we'll carry on down to the for loop and possibly call client.conn.Close again.  I'm not sure whether it's okay to do this or not.  Returning here should be safe as the overseer has not yet started so we don't need to wait on the waitgroup.  Also by returning here we do not print out the log overseer has exitted, which in this case is not strictly correct as it never started.

ciao-launcher/main.go
```diff
@@ -563,7 +564,7 @@ DONE:
 			glog.Flush()
 
 			/* We panic here to see which naughty go routines are still running. */
-
+			debug.SetTraceback("all")
```
markdryan: Hidden away in a dark corner of the launcher docs we have this.

https://github.com/01org/ciao/blob/master/ciao-launcher/doc.go#L47

Could you just delete those two sentences.

ciao-launcher/main.go
rbradford: Done.

ciao-launcher/main.go
rbradford: Done!

markdryan:
Looks good to me.  Tested locally and I can confirm it fixes the issue.  Thanks!  This was potentially quite a serious issue.

@mcastelino We're good to merge.

PR #605, opened 2016-09-27
Fix logging wrappers to give correct source information

rbradford:
Using the *Depth variants of the glog functions makes the log messages through the wrapper have the line number they were called at, not the line number inside the wrapper.

### DIFF
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -102,23 +102,27 @@ const (
 )
 
 func debugf(format string, args ...interface{}) {
-	glog.V(2).Infof("ciao-cli DEBUG: "+format, args...)
+	if glog.V(2) {
+		glog.InfoDepth(1, fmt.Sprintf("ciao-cli DEBUG: "+format, args...))
+	}
 }
 
 func infof(format string, args ...interface{}) {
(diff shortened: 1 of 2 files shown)


ciao-cli/main.go
```diff
@@ -102,23 +102,27 @@ const (
...
 func errorf(format string, args ...interface{}) {
-	glog.Errorf("ciao-cli ERROR: "+format, args...)
+	glog.ErrorDepth(1, fmt.Sprintf("ciao-cli ERROR: "+format, args...))
 }
 
 func fatalf(format string, args ...interface{}) {
-	glog.Fatalf("ciao-cli FATAL: "+format, args...)
+	glog.FatalDepth(1, fmt.Sprintf("ciao-cli FATAL: "+format, args...))
 	os.Exit(1)
 }
```
kaccardi: I know this isn't part of your patch, but I notice that os.Exit(1) is being called unnecessarily here as glog.FatalDepth will call os.Exit(255) according to the godocs.

rbradford: I popped another commit on to this PR to do just this.

PR #617, opened 2016-09-30
[DONOTMERGE] Enable floating IPs on CIAO CNCI agent

leoswaldo:
This PR will adds support of Floating ips to the CIAO CNCI agent:

CNCI agent will be ready to support operations to assign and release public IP as it is indicated to them.

Overall design:
Two user define chains into nat table of iptables

- ciao-floating-ip-pre added to PREROUTING chanin
- ciao-floating-ip-post added to POSTROUTING chain

We need to create two different chains, one for incoming routing and another for outgoing, so whenever any hook trigger the corresponding chains (POST/PRE routing) they will trigger ones we created and follow the traversal order, looking into rules that will match the traffic. The user define chains will have priority since we are adding them as first rule (target) for all network protools.

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -267,6 +267,31 @@ func (client *ssntpClient) RestartInstance(instanceID string, nodeID string) err
 	return err
 }
 
+func (client *ssntpClient) AssignFloatingIP(floatingIP payloads.FloatingIP) error {
+	floatingIPCmd := payloads.PublicIPCommand{
+		PublicIP:     floatingIP.AssignFloatingIP.Address,
+		PrivateIP:    floatingIP.AssignFloatingIP.InternalIP,
+		TenantUUID:   floatingIP.AssignFloatingIP.TenantUUID,
+		InstanceUUID: floatingIP.AssignFloatingIP.InstanceUUID,
+	}
+
+	payload := payloads.CommandAssignPublicIP{
+		AssignIP: floatingIPCmd,
+	}
+
+	y, err := yaml.Marshal(payload)
+	if err != nil {
+		return err
+	}
+
+	glog.Info("Assing Floating IP %s to internal %s", floatingIP.AssignFloatingIP.Address, floatingIP.AssignFloatingIP.InternalIP)
+	glog.V(1).Info(string(y))
+
+	_, err = client.ssntp.SendCommand(ssntp.AssignPublicIP, y)
+
+	return err
+}
+
 func (client *ssntpClient) EvacuateNode(nodeID string) error {
 	evacuateCmd := payloads.EvacuateCmd{
 		WorkloadAgentUUID: nodeID,
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -86,6 +86,17 @@ func (c *controller) deleteInstance(instanceID string) error {
 	return nil
 }
 
+func (c *controller) assignFloatingIP(floatingIP payloads.FloatingIP) error {
+	// validate the tenant exists
+	_, err := c.ds.GetTenant(floatingIP.AssignFloatingIP.TenantUUID)
+	if err != nil {
+		return err
+	}
+
+	go c.client.AssignFloatingIP(floatingIP)
+	return nil
+}
+
 func (c *controller) confirmTenant(tenantID string) error {
 	tenant, err := c.ds.GetTenant(tenantID)
 	if err != nil {
--- a/networking/ciao-cnci-agent/client.go
+++ b/networking/ciao-cnci-agent/client.go
@@ -200,7 +200,8 @@ func processCommand(client *ssntpConn, cmd *cmdWrapper) {
 			glog.Infof("Processing: CiaoCommandAssignPublicIP %v", c)
 			err := assignPubIP(c)
 			if err != nil {
-				glog.Infof("Error Processing: CiaoCommandAssignPublicIP %v", err)
+				glog.Errorf("%v", err)
+				glog.Errorf("Error Processing: CiaoCommandAssignPublicIP %v", err)
 			}
 		}(cmd)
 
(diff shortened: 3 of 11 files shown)


networking/ciao-cnci-agent/client.go
```diff
@@ -200,7 +200,8 @@ func processCommand(client *ssntpConn, cmd *cmdWrapper) {
 			glog.Infof("Processing: CiaoCommandAssignPublicIP %v", c)
 			err := assignPubIP(c)
 			if err != nil {
-				glog.Infof("Error Processing: CiaoCommandAssignPublicIP %v", err)
+				glog.Errorf("%v", err)
```
mcastelino: any reason you are printing it twice.

networking/ciao-cnci-agent/network.go
```diff
@@ -25,6 +25,7 @@ import (
 
 	"gopkg.in/yaml.v2"
 
+	"github.com/coreos/go-iptables/iptables"
```
mcastelino: it may make sense to actually put the code in firewall.go and call it from here... as it may be useful in other places.

Take a look at following places where the palce holders for this code already exists
https://github.com/01org/ciao/blob/master/networking/libsnnet/firewall.go#L281
https://github.com/01org/ciao/blob/master/networking/libsnnet/firewall.go#L341

All the code you wrote here should move into firewall.go and modify the existing placeholders

networking/ciao-cnci-agent/client.go
leoswaldo: My mistake, didn't see the err was already printed out.

networking/ciao-cnci-agent/network.go
leoswaldo: Got it

networking/ciao-cnci-agent/network.go
```diff
@@ -321,13 +321,13 @@ func assignPubIP(cmd *payloads.PublicIPCommand) error {
...
 	prIP, puIP, err := unmarshallPubIP(cmd)
-
 	if err != nil {
-		glog.Errorf("cnci.releasePubIP invalid params %v %v", err, cmd)
+		return fmt.Errorf("cnci.releasePubIP invalid params %v %v", err, cmd)
 	}
 
-	if enableNetwork {
-		glog.Infof("cnci.releasePubIP success %v %v %v", prIP, puIP, cmd)
+	err = gFw.PublicIPAccess(libsnnet.FwEnable, prIP, puIP, gCnci.ComputeLink[0].Attrs().Name)
```
mcastelino: I think there is a typo here. It should be libsnnet.FwDisable

leoswaldo: good catch

ciao-controller/client.go
```diff
@@ -267,6 +267,31 @@ func (client *ssntpClient) RestartInstance(instanceID string, nodeID string) err
 	return err
 }
 
+func (client *ssntpClient) AssignFloatingIP(floatingIP payloads.FloatingIP) error {
+	floatingIPCmd := payloads.PublicIPCommand{
+		PublicIP:     floatingIP.AssignFloatingIP.Address,
+		PrivateIP:    floatingIP.AssignFloatingIP.InternalIP,
+		TenantUUID:   floatingIP.AssignFloatingIP.TenantUUID,
+		InstanceUUID: floatingIP.AssignFloatingIP.InstanceUUID,
```
mcastelino: We should make providing the InstanceUUID optional. When we support k8s services, there will not be a instance behind the InternalIP. Hence allow for the case where the InstanceUUID is left un specified.

ciao-controller/command.go
```diff
@@ -86,6 +86,21 @@ func (c *controller) deleteInstance(instanceID string) error {
...
 
+func (c *controller) assignFloatingIP(floatingIP payloads.FloatingIP) error {
+	// get node id.  If there is no node id we can't send a delete
+	i, err := c.ds.GetInstance(floatingIP.AssignFloatingIP.InstanceUUID)
+	if err != nil {
+		return err
+	}
+
+	if i.NodeID == "" {
+		return errors.New("Cannot delete instance not assigned to Node")
```
mcastelino: Is this a typo?

ciao-controller/command.go
```diff
@@ -86,6 +86,21 @@ func (c *controller) deleteInstance(instanceID string) error {
 	return nil
 }
 
+func (c *controller) assignFloatingIP(floatingIP payloads.FloatingIP) error {
+	// get node id.  If there is no node id we can't send a delete
+	i, err := c.ds.GetInstance(floatingIP.AssignFloatingIP.InstanceUUID)
```
mcastelino: As mentioned before we should not assume there exists a instance behind a given floating IP. We can start with this method, but longer term we may need two types of floating IP's. The ones with an instance behind them and ones where they are for virtual service end points.

ciao-controller/client.go
leoswaldo: with the latest changes, it wont be necessary, we'll request only the TenantUUID as required

ciao-controller/command.go
leoswaldo: yeah, it was a typo, that code section was modified, so it's not there anymore, thanks Manohar

ciao-controller/command.go
leoswaldo: with latest changes we verify the tenant not the instance, fixed by now
thanks

openstack/compute/api.go
```diff
@@ -27,6 +27,8 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/gorilla/mux"
+
+	"github.com/01org/ciao/payloads"
 )
```
kaccardi: I'd like to avoid putting payloads into this file if possible. the Service interface implementation should be able to parse the openstack struct.

kaccardi:
I'll pull this branch and resubmit for review with changes.

mcastelino:
@kaccardi @tpepper closing this PR. We will do a resubmit from the branch at https://github.com/mcastelino/ciao/tree/floating_ip

PR #618, opened 2016-09-30
Support for qemu daemonization and socket based consoles

sameo:
This PR enables support for qemu socket based consoles and daemonization.
It also cleans some dead code up.


kaccardi:
looks good to me.

PR #619, opened 2016-09-30
ciao-controller: Honor compute API port override

kaccardi:
The code to start the compute API service was using the const
defined in the openstack package for the compute service and not
allowing the value to be overriden via the configuration command.
Set the default value of ciao-controller's computeAPI port to be
that const from the openstack package and use this value for
starting the service.


markdryan:
Looks good to me.

PR #625, opened 2016-10-01
[WIP] boot from volume

kaccardi:
These are patches to support booting from volume via a workload template.

### DIFF
--- a/_release/bat/release_test.py
+++ b/_release/bat/release_test.py
@@ -405,10 +405,6 @@ class BATTests(unittest.TestCase):
         """Start one instance of all workloads"""
         self.failUnless(launch_all_workloads())
 
-    def test_start_all_workloads10(self):
-        """Start 10 instances of all workloads"""
-        self.failUnless(launch_all_workloads("10"))
-
     def test_get_cncis(self):
         """Start a random workload, then get CNCI information"""
         self.failUnless(start_random_workload())


markdryan:
The SingleVM has it's own copy of workload_resources.csv, as the CNCI requires less resources when running in the SingleVM.  This file will need to be updated as well.

markdryan:
Also, I'm having problems deleting a boot from volume instance.  Actually, the volume does seem to go away but I get this error.

```
ciao-cli instance delete --instance e37a7d7e-30f6-44aa-858e-781ad4672be9
F1003 15:43:31.481160   24847 instance.go:171] ciao-cli FATAL: Instance deletion failed: 204 No Content
goroutine 1 [running]:
```

markdryan:
We still need to update:  https://github.com/01org/ciao/blob/master/testutil/singlevm/tables/workload_resources.csv

markdryan:
The command to create the rbd from the qcow image is still incorrect.  You need this:

```
	        rbdStr := fmt.Sprintf("rbd:rbd/%s:id=%s",ID,d.ID)
		cmd = exec.Command("qemu-img", "convert", "-O", "rbd", *imagePath, rbdStr)
```

markdryan:
I wasn't able to attach a volume created from an image to an instance.  The volume was created correctly but remained for ever in Attaching.

```
Volume #1
	Name             [bootest]
	Size             [3 GB]
	UUID             [80facdc6-cd16-483c-99fd-376253222835]
	Status           [attaching]
	Description      []
```

Here's how I created the volume:

```
ciao-cli volume add -name bootest -size 3 -source /var/lib/ciao/images/73a86d7e-93c0-480e-9c41-ab42f69b7799
Created new volume: 80facdc6-cd16-483c-99fd-376253222835
```

markdryan:
Controller unit tests don't seem to be compiling.

```
# github.com/01org/ciao/ciao-controller
./controller_test.go:1139: undefined: context in context.ds
./controller_test.go:1144: undefined: context in context.image
./controller_test.go:1168: undefined: context
```

kaccardi:
@markdryan we do still need the workload template changes to test boot from volume. Creating the boot able volume with ciao-cli just allows you to then use that volume if you want to boot from an existing volume rather than create one on the fly.

kaccardi:
@markdryan how were you trying to attach the volume you created.

kaccardi:
@markdryan I rebased to master and then fixed the issues with controller_test.

ciao-controller/instance.go
```diff
@@ -166,23 +191,71 @@ func getStorage(c *controller, wl *types.Workload, tenant string) (payloads.Stor
 			return payloads.StorageResources{}, err
 		}
 
+		// don't you need to add support for indicating whether
```
kaccardi: I expect the original failure was due to https://github.com/01org/ciao/pull/621 which was present in this PR till I rebased just now. The issue of continuing to issue the start request despite the failure has been addressed in a separate commit.

kaccardi:
@markdryan I wish I could respond to all your comments inline, this is irritating. But, I filed an issue for the error you were seeing when you tried to delete an instance which was booted from volume. https://github.com/01org/ciao/issues/652 The instance was deleted properly, but ciao-cli expects a 200 - since I recently modified the compute API to send responses that matched the openstack compute api spec, this is not what we send on successful delete - we actually send 204.

markdryan:
@kaccardi The unit tests failures are due to a compile error in the launcher unit tests.

```
015,#16167 /home/travis/gopath/src/github.com/01org/ciao/ciao-launcher/overseer.go:#16043,#16087

# github.com/01org/ciao/ciao-launcher

ciao-launcher/docker_test.go:111: cannot use s (type dockerTestStorage) as type storage.BlockDriver in field value:

	dockerTestStorage does not implement storage.BlockDriver (missing CopyBlockDevice method)

ciao-launcher/docker_test.go:191: cannot use s (type dockerTestStorage) as type storage.BlockDriver in field value:

	dockerTestStorage does not implement storage.BlockDriver (missing CopyBlockDevice method)
```

There's a type in the launcher unit tests, dockerTestStorage, that implements the BlockDriver interface.  As this PR adds a new method to this interface, dockerTestStorage no longer implements all the methods in BlockDriver and this is causing the launcher unit tests to fail to compile.  So we need to update dockerTestStorage.  The docker unit tests don't call this function so we can just do nothing and return nil.

ciao-controller/tables/workload_storage.csv
```diff
@@ -0,0 +1 @@
+79034317-3beb-447e-987d-4e310a8cf410,"",1,1,0,"image",73a86d7e-93c0-480e-9c41-ab42f69b7799
```
markdryan: I killed controller, restarted it and then checked the db.  I had two identical rows in the workload_storage table.

markdryan:
@kaccardi 

> @markdryan how were you trying to attach the volume you created.

I attached it after the instance had booted via ciao-cli volume attach

ciao-storage/ceph.go
```diff
@@ -42,7 +42,8 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 	// Currently the kernel rdb client only supports layering but in the future more feaures
 	// should be added as they are enabled in the kernel.
 	if imagePath != nil {
-		cmd = exec.Command("rbd", "--id", d.ID, "--image-feature", "layering", "import", *imagePath, ID)
+		rbdStr := fmt.Sprintf("rbd:rbd/%s:id=%s", ID, d.ID)
+		cmd = exec.Command("qemu-img", "convert", "-f", "qcow2", "-O", "rbd", *imagePath, rbdStr)
```
markdryan: If we specify -f qcow2 then all source images need to be in qcow2 format.  If we remove these two arguments, qemu-img will autodetect the source type for us.

kaccardi:
@markdryan I was not able to duplicate the problem with the attaching volume staying in the attaching state forever with this version of the PR. Hopefully the problem was fixed.

markdryan:
@kaccardi I'm still seeing the following storage related problems:

1. Volumes created from a source qcow image cannot be attached to an existing VM instance via ciao-cli volume attach.  They remain in the attaching state.  I have no problems with empty volumes.
2. When I try to attach the volume in this way the names of all the volumes, the attaching one and all other pre-existing volumes, get messed up when I do a ciao-cli volume list, i.e., the names seem to get randomly assigned to the volumes.
3. I get an error when deleting an instance which has a volume attached.
4. I can get duplicate entries in the workload_storage table by editing workload_storage.csv and restarting controller.
5. I have BAT tests failures on Single VM due to timeouts. 

1,2 and 3 are unrelated to this PR I think and shouldn't prevent us from merging.  2 and 3 already have bugs.  So perhaps I just need to enter a bug for 1 and we can stop discussing it here.

That just leaves us with 4 and 5 which I guess we need to fix before merging.  Does that sound reasonable?

ciao-storage/ceph.go
```diff
@@ -42,7 +42,8 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 	// Currently the kernel rdb client only supports layering but in the future more feaures
 	// should be added as they are enabled in the kernel.
 	if imagePath != nil {
-		cmd = exec.Command("rbd", "--id", d.ID, "--image-feature", "layering", "import", *imagePath, ID)
+		rbdStr := fmt.Sprintf("rbd:rbd/%s:id=%s", ID, d.ID)
+		cmd = exec.Command("qemu-img", "convert", "-O", "rbd", *imagePath, rbdStr)
```
markdryan: We need to add qemu-utils to deps.go.

rbradford:
I think it's important to address #675 as part of this change. I ran into this problem due to issues outside of controller (image store) and also from the lack of qemu-img in osprepare. If the creation of the volume from an image fails the controller shouldn't forward the request onto ciao-launcher where it will be immediately rejected.

It would be nice to give feedback to the user that the instance creation fails but I don't know how that would integrate with #673.

Issue #536, opened 2016-09-07
Make SingleVM useful for development


markdryan:
Right now SingleVM is more of a BAT test rather than a development environment.  When you start the single VM it creates 4 instances which have some quite heavy workloads.  Once these are finished it tears everything down.  As a developer, I would just like all the ciao services to be installed and to be able to run my own tests without having to wait for the standard SingleVM tests to complete.  I also I don't want everything pulled down.  Basically, I'd like an option so that the script stops after the ciao services have been started, and prints something out the stdout saying set these environment variables to use ciao-cli.

Also, I think this should be the default behaviour.  The current behaviour where we run all the tests should only be enabled if we specify a --bat flag.

markdryan:
Perhaps we could simply place the bat tests in a separate script which could be run when singlevm starts up.  So maybe we could have

setup to initialise everything
run_bat to run the tests
teardown to stop everything

markdryan:
The other thing that's a bit nasty about the script is that it temporarily modifies your /etc/hosts file.  This means that it can't be run directly on a development machine.  It needs to be run inside a VM or a container.  How far are we off getting all this to work inside a container?  Can we just put all this stuff inside an image and upload to docker hub?

mcastelino:
@markdryan agree. Will add a BAT and standalone mode.

@kaccardi @tpepper We need a way to delete an tenant and associated CNCI to ensure the cluster can be torn down properly (vs just killing the CNCI).

The /etc/hosts change should not be required. We need to setup certificates and DNS resolution properly in both directions so that TLS will not complain.

markdryan:
Most of this is already done.  We just need to make the automated tests optional.

PR #630, opened 2016-10-04
Single VM: Enable Ceph and script refactor

markdryan:
This PR makes some changes to the Single VM scripts.  The first set of changes focus on tidying up the scripts:

1. The cleanup code is now in one place.
2. Misleading instructions from the controller logs are no longer written to the screen
3. The code to remove all docker containers has been removed as it's not needed.
4. Fixed some other bugs related to cleanup
5. The standard tests run by setup.sh have been moved to a new script, verify.sh and are now optional.  This is desirable when using Single VM for development as you don't want to run these scripts each time you run ./setup.sh.

The second set of changes enable ceph inside Single VM.

Fixes PR #536

### DIFF
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -12,7 +12,7 @@ sudo killall ciao-scheduler
 sudo killall ciao-controller
 sudo killall ciao-launcher
 sleep 2
-sudo "$ciao_gobin"/ciao-launcher --hard-reset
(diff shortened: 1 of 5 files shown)


markdryan:
@mcastelino  Could you review this one please?

mcastelino:
@markdryan LGTM. Have not tried it on my system yet. But the changes make sense.

mcastelino:
Verified on FC24.

PR #632, opened 2016-10-04
ciao-image,networking: Add Generic persistent data support with BoltDB

obedmr:
This PR add support for persistent data with BoltDB key/value database.

It includes:
### DB Layer
- BoltDB key/value database
- Generic Encode/Decode 
### For Image Service
- Write a new image
- Delete image
- Get image
- Get all images
### For networking (docker_plugin)
- Adapt docker_plugin to support the generic DB layer


mcastelino:
@obedmr I implemented a go maps to gob+boltDb storage implementation for networking. I need to make it a little more generic to be agnostic of the actual map used.  Maybe it is time I did that so that it can be used generically.

https://github.com/01org/ciao/blob/master/networking/libsnnet/docker_database.go

The only part of this that is not generic is
https://github.com/01org/ciao/blob/master/networking/libsnnet/docker_database.go#L80

obedmr:
@mcastelino could you take a look on this?

obedmr:
@mcastelino, finally, I made it work with `èncoding/gob``, I'd say mapRebuild is more for a more specific case, what do you think? For example: in ``ciao-image``if the service crashes, you can just restart the service. Memory is used just as a temporal cache, if something is requested and it's not in cache, it will look for it at bolt DB and will update cache db.

mcastelino:
@obedmr cache on demand will work.. however having an option to pre-cache is nice... specially if you are in RO mode most of the time

ciao-image/datastore/boltdb.go
```diff
@@ -0,0 +1,92 @@
...
+func (b *Boltdb) Get(ID string) (Image, error) {
+	b.DbInit(b.DbDir, b.DbFile)
+	defer b.DbClose()
+
+	img := Image{}
+	data, err := b.DbGet("images", ID)
+	if err != nil {
+		return img, fmt.Errorf("Error on image retrieve: %v ", err)
+	}
+	vr := bytes.NewReader(data.([]byte))
```
markdryan: This will panic if the conversion fails.  For data read in off the disk this might not be the correct behaviour.  A corrupt database might lead to a image service panic.  Why don't you just pass []bytes to DbGet and DbPut.  This way you wouldn't need the casts at all?

ciao-image/datastore/posix.go
```diff
@@ -44,5 +44,12 @@ func (p *Posix) Write(ID string, body io.Reader) (int64, error) {
 func (p *Posix) Delete(ID string) error {
 	imageName := path.Join(p.MountPoint, ID)
 
-	return os.Remove(imageName)
+	_, err := os.Stat(imageName)
+	if !os.IsNotExist(err) && err != nil {
```
markdryan: I'm not sure the logic is correct here.  If os.IsNotExist returns true, i.e., the file is known to not exist we then proceed to try and delete it which doesn't make any sense.  You probably want

```
    if err != nil {
        if os.IsNotExist(err) {
            err  = nil
       }
      return err
   }
```

ciao-image/datastore/posix.go
```diff
@@ -44,5 +44,12 @@ func (p *Posix) Write(ID string, body io.Reader) (int64, error) {
...
+		return err
+	}
+
+	os.Remove(imageName)
```
markdryan: What about checking the error from os.Remove?  if you want to ignore it, fair enough but assign it to _ to show that you really want to do this and that this is not just an oversight.

.travis.yml
```diff
@@ -59,7 +59,7 @@ script:
    - sudo mkdir -p /var/lib/ciao/instances
    - sudo chmod 0777 /var/lib/ciao/instances
    - test-cases -v -timeout 9 -text -coverprofile /tmp/cover.out -short github.com/01org/ciao/ciao-controller/...
-   - test-cases -v -timeout 9 -text -coverprofile /tmp/cover.out -append-profile -short github.com/01org/ciao/ciao-launcher github.com/01org/ciao/ciao-scheduler github.com/01org/ciao/payloads github.com/01org/ciao/configuration github.com/01org/ciao/testutil github.com/01org/ciao/ssntp/uuid github.com/01org/ciao/qemu github.com/01org/ciao/openstack/...
+   - test-cases -v -timeout 9 -text -coverprofile /tmp/cover.out -append-profile -short github.com/01org/ciao/ciao-launcher github.com/01org/ciao/ciao-scheduler github.com/01org/ciao/payloads github.com/01org/ciao/configuration github.com/01org/ciao/testutil github.com/01org/ciao/ssntp/uuid github.com/01org/ciao/qemu github.com/01org/ciao/openstack/... github.com/01org/ciao/ciao-image/...
```
markdryan: We're going to need some unit tests for the new database package as well.

ciao-image/datastore/boltdb.go
```diff
@@ -0,0 +1,92 @@
...
+	"bytes"
+	"encoding/gob"
+	"fmt"
+
+	"github.com/01org/ciao/database"
+)
+
+// Boltdb implements the DataStore interface for persistent data
+// on a key/value database with BoltDB
+type Boltdb struct {
```
markdryan: Why is this struct called BoltDb?  Isn't it agnostic about the backend used?

obedmr:
@markdryan, I'm currently OoO, early on next week (Monday), it's my first item to do on my Agenda. Expect the updated PR on Monday or Tuesday.

markdryan:
@obedmr Great.  BTW the image service has been integrated into Single VM.  This means that we'll need to update the Single VM scripts to ensure that the new db gets deleted when we tear down the single vm or restart it.

ciao-image/datastore/boltdb.go
obedmr: @markdryan, yeah, it should be more agnostic, /me changing it to ``MetaDs``

obedmr: done.

obedmr:
@markdryan, PR was updated. Could you please review the unit testing for dbprovider, not sure if it's a good practice to repeat the code there. I was thinking that I may add some extra functions for not repeating code. For example, when testing dataDbAdd, I need to do all the things I did @ DbInit and DbInitTable

ciao-image/datastore/metads.go
```diff
@@ -0,0 +1,91 @@
...
+	m.DbInit(m.DbDir, m.DbFile)
+	defer m.DbClose()
+
+	img := Image{}
+	data, err := m.DbGet("images", ID)
+	if err != nil {
+		return img, fmt.Errorf("Error on image retrieve: %v ", err)
+	}
+	vr := bytes.NewReader(data.([]byte))
+	if err := gob.NewDecoder(vr).Decode(&img); err != nil {
```
mcastelino: @obedmr this feels wrong... we do the encoding in the database, but decoding here. Ideally the encode and decode should be transparent and done in the database layer.

obedmr: @mcastelino, you're right, you mentioned that earlier, /me fixing that tomorrow morning, will be taking the part you did @ docker plugin. Thanks

markdryan:
@obedmr I guess my main concern is that we now have two DB abstractions for bolt db,  one in database and one in networking/libsnnet.  Looking at the interfaces there seems to be a lot of overlap between these two abstractions.   Is there any thing preventing you from using the networking/libsnnet implementation in the image service (it would of course need to be moved to a separate package).  The nice thing about the existing networking/libsnnet code is that it is mature so we can re-use it risk free in other components.  Assuming you can re-use it, I think the best course of action would be to take the existing code from libssnet, move it to the database package, and add whatever is missing for the image service.

@mcastelino  Does this make sense to you?

I'd also like to see the interface renamed to indicate that it's a keystore db and not a relational db.  We are still using relational dbs in ciao so this will avoid confusion should we ever extract this code from controller.

obedmr:
@markdryan, I think that that would be the idea, to have a single BoltDB abstraction, that's why I've  been doing this. Once we're sure that this generic and boltdb database abstraction meets all that is needed by networking and ciao-image, I'd proceed to send the PR for changing the networking package for using it instead of using its own implementation. Actually, I took networking package implementation as base for my code, there are some pieces missing, but I'll be working on them and let @mcastelino for reviewing it.

@markdryan, Regarding the keystone db, I don't understand why it's mentioned here. This ciao-image db usage is not a keystone thing. It's used for image metadata information. Could you elaborate on that part? I'd like to see what is the Keystone relation here.

markdryan:
So are you planning to refactor the networking code to use the new DbProvider interface in this PR?

@obedmr key store not keystone.  We use two different types of dbs in ciao right now, relational and no sql.  If we were to genericize the db code in controller that uses a relational database and put it in github.com/01/org/ciao it would probably need to have its own interface.  What would be call it?  DbProvider is already taken?

obedmr:
@markdryan, yes, I'm trying to refactor the networking code with this.

Oh, I'm so sorry for my reading mistake, apologies, I don't know why I read it as Keystone.  I got your point about relational vs key/value databases. I think we could use the same ``DbProvider`` interface,  I think, the current DbProvider functions may also apply for a relational db. Not sure if any more complex DB operations (queries) is required in the controller code (I have not seen the db implementation there). What do you think? do you think it would have more functions or what could be different?

obedmr:
@markdryan, I think networking db code refactor could be in other PR, this could be out of this ciao-image scope. But yeah, will work on that once this dbProvider stuff is done.

ciao-image/datastore/cache.go
```diff
@@ -45,6 +45,7 @@ func (c *ImageCache) CreateImage(i Image) error {
 	c.lock.Lock()
 
 	c.images[i.ID] = i
+	c.metaDs.Write(i)
```
markdryan: Does Write return an error. If so we should trap it and maybe try to write before we update the map.  This way we have nothing to roll back.

ciao-image/datastore/posix.go
```diff
@@ -44,5 +44,15 @@ func (p *Posix) Write(ID string, body io.Reader) (int64, error) {
...
-	return os.Remove(imageName)
+	_, err := os.Stat(imageName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			err = nil
+		}
+		return err
+	}
+
+	os.Remove(imageName)
```
markdryan: We're still ignoring the error from os.Remove

ciao-image/main.go
```diff
@@ -43,8 +46,16 @@ func init() {
...
 func main() {
-	// TBD Select the right datastore interface
-	metaDs := &datastore.Noop{}
+	metaDs := &datastore.MetaDs{
+		DbProvider: database.NewBoltDBProvider(),
+		DbDir:      dbDir,
+		DbFile:     dbFile,
+	}
+	metaDsTables := []string{"images"}
+	metaDs.DbInit(metaDs.DbDir, metaDs.DbFile)
```
markdryan: And ignoring the errors from DbInit and DBTableInit.

A good utility to help you catch all these missing error checks is

https://github.com/kisielk/errcheck

If you know its safe to ignore the error you can explicitly cast it away using _ =

markdryan:
@obedmr I guess I feel that the refactoring of the docker network plugin to use the new generic db package should be done as a separate commit but as part of this PR, otherwise we might simply forget.  Testing should be easy.  You just need to boot up Single VM and run the verify.sh script.  This should test docker networking.

@mcastelino Do you agree, or would you prefer to migrate the docker networking code yourself in a separate PR?

obedmr:
@markdryan, sure, will be working on it Today. thanks for the single VM testing tip

mcastelino:
@markdryan @obedmr I agree with Mark that the changes to the network plugin has to be in the PR. That will be truly indicative if this implementation is truly generic.

Also to ensure the changes made are correct the networking unit tests as well the single vm tests should run.

That said, the docker plugin code manages the maps explicitly (with mutexes) and uses the data base seperately. @obedmr code does this differently. I wonder how much effort it will be to change the db implementation or the docker plugin to accommodate this change.

obedmr:
@mcastelino, I like the idea of having db+encode+decode+cache in same database abstraction layer. Let me do an approach and I'll update this PR

obedmr:
@mcastelino, @markdryan I updated this PR to make it more generic. I think, with this implementation will be easier to adapt the docker plugin to use this. @mcastelino, what do you think? is there something I'm missing? I'll start doing some test with the ``docker_plugin`` to use this DB layer.

obedmr:
@markdryan @mcastelino, when you have a chance could you take a look on it?

ciao-image/main.go
```diff
@@ -31,6 +32,8 @@ var identity = "https://localhost:35357/"
 var userName = "csr"
 var password = "hello"
 var mountPoint = "/var/lib/ciao/images"
+var dbDir = "/var/lib/ciao/images"
```
markdryan: I think it's probably best not to store the db in this directory, which up to now has just been used for actual images.  How about /var/lib/ciao/ciao-image/ciao-image.db

markdryan: We also need to modify singlevm to clean up this.  This should be done in setup.sh around line 119 after we delete the controller.db file.

database/dbprovider.go
```diff
@@ -0,0 +1,41 @@
+//
```
markdryan: We need to add a README.md file in the new database directory.

ciao-image/datastore/datastore_test.go
```diff
@@ -51,27 +54,29 @@ func testGetAll(t *testing.T, d RawDataStore, m MetaDataStore) {
...
-	images, err := cache.GetAllImages()
+	images, err := imageStore.GetAllImages()
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	if len(images) != 1 {
-		t.Fatalf("len is actually %d\n", len(images))
-	}
+	if fmt.Sprintf("%T", m) != "*datastore.Noop" {
```
markdryan: This is a bit odd.  Is the intention to simply check to see if m is not a *datastore.Noop?  if so it might be better to use a type assertion.

if _, ok := m.(*datastoreNoop); !ok {

}

ciao-image/datastore/metads.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+//  closeDb is a generic function to close every Db transaction
+func closeDb(metaDs *MetaDs) {
+	_ = metaDs.DbClose()
+}
+
+// Write is the metadata write implementation.
+func (m *MetaDs) Write(i Image) error {
+	defer closeDb(m)
```
markdryan: Is it correct to close the db if init fails?

ciao-image/datastore/metads.go
```diff
@@ -0,0 +1,100 @@
...
+
+	err := m.DbInit(m.DbDir, m.DbFile)
```
markdryan: Why do we open and close the database on each operation?  Is there any reason why we don't init the database once when we start the image service and close it when image service shutdowns.

ciao-image/datastore/metads.go
```diff
@@ -0,0 +1,100 @@
...
+	err = m.DbInit(m.DbDir, m.DbFile)
+	if err != nil {
+		glog.Errorf("Error on Db Initialization: %v", err)
+	}
+
+	var elements []interface{}
+	imageTable := &ImageMap{}
+	elements, err = m.DbProvider.DbGetAll("images", imageTable)
+
+	for _, img := range elements {
```
markdryan: You could avoid re-allocation here by doing

images = make([]Image, len(elements)

for i, img := range elements {
		image := img.(*Image)
 		images[i] = *image
}

ciao-image/datastore/store.go
```diff
@@ -0,0 +1,154 @@
...
+func (s *ImageStore) Init(rawDs RawDataStore, metaDs MetaDataStore) error {
+	s.metaDs = metaDs
+	s.rawDs = rawDs
+
+	return nil
+}
+
+// CreateImage will add an image to the datastore.
+func (s *ImageStore) CreateImage(i Image) error {
+	defer s.ImageMap.Unlock()
```
markdryan: I'd switch these two statements around.  It might matter if Lock were to panic.

ciao-image/datastore/store.go
```diff
@@ -0,0 +1,154 @@
...
+	s.ImageMap.Lock()
+
+	err := s.metaDs.Delete(ID)
+	if err != nil {
+		return image.ErrNoImage
+	}
+
+	err = s.rawDs.Delete(ID)
+	if err != nil {
+		return err
```
markdryan: We need to think how best to handle this case.  The user will receive an error telling him that the image could not be deleted.  However, if he does a GetImage after receiving the error the Image won't exist.  I'm not sure what the best plan would be here.  Try to delete the image first perhaps.

obedmr:
@markdryan thanks for the feedback, I will be working on this tomorrrow.

ciao-image/datastore/store.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	return nil
+}
+
+// UploadImage will read an image, save it and update the image cache.
+func (s *ImageStore) UploadImage(ID string, body io.Reader) error {
+	img, err := s.metaDs.Get(ID)
```
markdryan: Don't we need a lock here?

database/boltdb.go
```diff
@@ -203,28 +146,50 @@ func (db *dbProvider) DbDelete(table string, key string) (err error) {
 	return err
 }
 
-func (db *dbProvider) DbGet(table string, key string) (value interface{}, err error) {
+func (db *dbProvider) DbGet(table string, key string, dbTable DbTable) (interface{}, error) {
 
-	err = db.DB.View(func(tx *bolt.Tx) error {
+	elem := db.Cache[table][key]
```
markdryan: Do we allow people to store nil values in the db?  As far as I can tell this is not disallowed by DbAdd, which means this code isn't correct as it cannot distinguish between the case where a key is actually missing from the cache and a key is present in the cache but the associated object is nil.  To actually see whether something is stored in the cache or not use.

elem, present := db.Cache[table][key]
if present {
   return elem, nil
}

markdryan:
@obedmr I tested your PR today on Single VM and it worked fine.  The verify.sh script worked demonstrating that docker networking is still working and I noted that if I killed and restarted the image service it did not lose information about the files that it was managing.

I've made a number of minor comments some of which will need to be addressed before we can merge.

mcastelino:
@obedmr the cache design as implemented is not ideal IMO. It will result in two copies of that data in memory and in case of large data bases will be an issue. I implemented the cache using interfaces in the networking code. I see that you removed the interfaces and created a standalone cache. If you use interfaces you do not end up with two copies.

markdryan:
@obedmr Just an additional comment on the commit messages.   Many of them are missing descriptions and there are no references to existing github issues so it's not clear which bugs this PR will fix.  Could you add these?

markdryan:
And, I'm afraid you'll need to rebase.

database/boltdb.go
```diff
@@ -108,38 +73,13 @@ func (db *dbProvider) DbClose() error {
...
-	return err
-}
-
-func (db *dbProvider) DbTableInit(tables []string) (err error) {
+func (db *dbProvider) DbTablesInit(tables []string) (err error) {
 
 	glog.Infof("dbInit Tables := %v", tables)
-	for i, v := range tables {
-		glog.Infof("table[%v] := %v, %v", i, v, []byte(v))
+	db.Cache = make(map[string]map[string]interface{})
```
obedmr: Oh, interesting, I didn't noticed that, so, maybe there's no need of Cache for this DB layer, @mcastelino  is this good for the networking code? Is it really necesary to keep the map you had? I think we could delegate this caching task to Bolt.

mcastelino:
@obedmr I never had an internal cache. I had used interfaces to allow seamless mapping between the db and my data structures. That is the gap that this PR does not address. By removing the interfaces the networking code has to change more than I really like.

ciao-image/datastore/metads.go
obedmr: @markdryan, agree with open DB at the beginning and close it when service shutdowns.  /me applying the fix. thanks.

obedmr:
@mcastelino, ok, got it, I will bring those back and see how it works with the addition of this DB layer.

ciao-image/datastore/store.go
obedmr: Agree, let's first remove image from rawds, and if it fails, it keeps metads

database/dbprovider.go
obedmr: Done on latest commit.

obedmr:
@mcastelino @markdryan I did updates on PR in order to address the review comments. The only missing piece is the Single VM cleanup code. I will add it tomorrow. When you have a chance, could you take a look?

database/boltdb.go
```diff
@@ -203,28 +166,43 @@ func (db *dbProvider) DbDelete(table string, key string) (err error) {
...
+	err := db.DB.View(func(tx *bolt.Tx) error {
 		bucket := tx.Bucket([]byte(table))
 		if bucket == nil {
 			return fmt.Errorf("Bucket %v not found", table)
 		}
+		data := bucket.Get([]byte(key))
+		vr := bytes.NewReader(data)
 
-		val := bucket.Get([]byte(key))
-		if val == nil {
```
mcastelino: @obedmr was the nil check removed for a particular reason?

obedmr: @mcastelino, good catch, I missed that. You're right, I'm bringing it back now.

mcastelino:
@obedmr things look good from networking point of view. @markdryan Please merge once you re-review the changes

markdryan:
@obedmr We still need to fix the Single VM issue I think.  I don't see a commit for this.

database/boltdb.go
```diff
@@ -28,62 +28,26 @@ import (
...
-type tableBoltDB struct {
+type boltDB struct {
 	Name string
 	DB   *bolt.DB
 }
 
-func newTableBoltDb() *tableBoltDB {
-	return &tableBoltDB{
-		Name: "tableBolt.DB",
+type dbProvider boltDB
```
markdryan: I don't really understand why we define a new type here.  Why don't we just rename the boltDB struct dbProvider or even better dbProvider boltDB?

obedmr: Good point, removing that, mybe it was just a copy and paste issue from networking code: https://github.com/01org/ciao/blob/master/networking/libsnnet/database.go#L78

obedmr:
@markdryan I just added the ciao-image database removal instruction that was required for making sure that we leave clean ciao image environment from Single VM.

markdryan:
@obedmr Looks good.  You've answered all my comments and questions.  I'll give it one more test before merging.  I might not get this done until tomorrow morning though.

markdryan:
I'm afraid the new PR isn't working for me.  Just calling setup.sh inside SingleVM fails.  The reason seems to be that no error is returned if a key is not found in a table.  You also don't get an error if you try to delete an non existing image, e.g.,

```
# ciao-cli image delete --image idontexist
Deleted image idontexist
```

obedmr:
@markdryan it's fixed for deleting "not existing images" it returns a 404 HTTP code.

markdryan:
@obedmr It's still not working inside Single VM.  I can't merge this PR right now as it breaks Single VM.

obedmr:
@markdryan, it's strange, it was working for me, well, I'll take a look on it later Today.

obedmr:
@markdryan I just tested with single vm scripts and it seems to work, is there any particular error or issue you're getting?

Here/s te recording of my singlevm run: http://asciinema.org/a/b2c7v9pv5oujoiezala92q70v

markdryan:
@obedmr Still not working for me I'm afraid.  Here's the error I'm getting

```
F1108 18:29:26.376625     860 image.go:157] ciao-cli FATAL: Could not create image [Expected HTTP response code [201] when accessing [POST https://singlevm:9292/v2/images], but got 409 instead
Conflict
null]
goroutine 1 [running]:
github.com/01org/ciao/vendor/github.com/golang/glog.stacks(0x8fdc00, 0xc400000000, 0xd0, 0x226)
	/home/singlevm/hostgo/src/github.com/01org/ciao/vendor/github.com/golang/glog/glog.go:769 +0xa5
github.com/01org/ciao/vendor/github.com/golang/glog.(*loggingT).output(0x8e1840, 0xc400000003, 0xc42001f680, 0x8bd271, 0x8, 0x9d, 0x0)
```

markdryan:
I do get an error when deleting non-existent images though so that piece is working now.   I can't figure out why it is working for you.  Is your git status clean?

obedmr:
@markdryan I run it in a fresh VM, I 
```
go get -u github.com/01org/ciao/...
.
.
git remote add obedmr github.com/obedmr/ciao/ 
git fetch obedmr singlevm
```
Maybe, image is not ready when you're trying to create the image.

markdryan:
Maybe there's something wrong with my setup then.

@mcastelino Could you possibly give this a try?

The thing is I think I can see why it's failing and looking at the code I would expect it to fail.  It's failing for me here:

https://github.com/01org/ciao/blob/master/ciao-image/service/service.go#L50

When we add an image in setup.sh we specify an ID.  The image service makes sure that the ID doesn't already exist before adding the image.  It does this by looking up the image id in the datastore.  It only allows the image to be added if this Get on the image id fails.  But I think in the current code there is no failure.    DbGet just returns nil, nil, and so the image service returns an error as it thinks the image already exists.

If I modify DbGet to return an error instead of a nil, nil when the key does not exist,

https://github.com/01org/ciao/pull/632/files#diff-b11d2d27cf6458c0707ae9a0fbe82d3fR192

everything works.  What I don't understand is why this is working for you.  I can see from your video that the IDs are being specified when uploading the images.

markdryan:
@obedmr Do you checkout the singlevm branch before running setup?

obedmr:
@markdryan yeah, my bad, I checkout ``singlevm`` branch which is not the right one. The good one is the ``ciao-image``. I'm not working in the fix.

obedmr:
@markdryan, that was a really good catch, I hadn't tested that scenario (sending new images with IDs), I had to create a new [commit](https://github.com/01org/ciao/pull/632/commits/ae2aab79597998dc10dd8171834fe0645257dbfb) for it, becuase it was an issue that was there before. 

Here's my new video: 
https://asciinema.org/a/205qi6m70htvg901xq1p3pouj

markdryan:
Everything's working for me now.  Thanks for persevering with this @obedmr.

PR #634, opened 2016-10-05
Add support for software QEMU virtualisation

rbradford:
In order to aid testing of Single VM mode on Travis this change adds a ciao-launcher option that lets you enable software QEMU. A follow-up PR will need to switch Single VM mode to use "auto" mode.


mcastelino:
@rbradford  I need to create a CNCI based on debian/centos as clear linux requires KVM to be able run. I am working on this right now. Will push an alternate CNCI image out shortly.

markdryan:
Works great. Thanks!

PR #637, opened 2016-10-05
ciao-cli: Make volume and image service work with -ca-file

markdryan:
The volume and image service were not passing the CA certificate,
specified either on the command line or via an environment
variable, down to gophercloud.  This meant that these services
could not work inside the Single VM.  As we now have three different
gophercloud services requiring the same authentication code, the
original code that implemented this feature for the identity service,
has been moved to a new function.

### DIFF
--- a/ciao-cli/authenticate.go
+++ b/ciao-cli/authenticate.go
@@ -1,0 +1,48 @@
+//
+// Copyright (c) 2016 Intel Corporation
+//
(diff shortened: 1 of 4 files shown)


rbradford:
Looks good to me.

rbradford:
I successfully tested the volume & identity aspects of this change when I was testing PR#640

PR #640, opened 2016-10-06
controller, payloads, configuration, testutil: Enable volume service

markdryan:
This commit does three things:

1. It adds the port of the volume service API to the ciao-configuration file.
2. It modifies controller so that it passes the cluster configured port number
   to the identity service.
3. It updates the identity service so that it advertises the volume service.


markdryan:
This PR is needed to enable the volume service in SingleVM.

rbradford:
Tested on Ubuntu SingleVM in combination with PR#637. Tested by using ciao-cli to add, list, show and delete volumes.

PR #647, opened 2016-10-06
Networking: Cleanup the docker plugin database implementation

mcastelino:
Move to generic database implementation for the docker plugin database.
This allows for the reuse of this implementation to store any in memory
key value database to persistent store implemented with boltDb+gob

### DIFF
--- a/networking/libsnnet/docker_plugin.go
+++ b/networking/libsnnet/docker_plugin.go
@@ -137,13 +137,29 @@ var DockerPluginCfg = struct {
 	Timeout: 1 * time.Second,
 }
 
+//DbTable interface that needs to be supported
+//for the table to be handled by the database
+type DbTable interface {
+	// Creates the backing map
+	NewTable()
+
+	// Name of the table as stored in the database
+	Name() string
+
+	// Allocates and returns a single value in the table
+	NewElement() interface{}
+
+	// Add an value to the in memory table
+	Add(k string, v interface{}) error
+}
+
 // A DockerDBProvider represents a persistent data base provider
 // that can be used by the DockerPlugin to store its internal state
 type DockerDBProvider interface {
 	//Initializes the Database
 	DbInit(dir string) error
 	//Populates the DockerPlugin cache from the database
-	DbMapRebuild(table string, dockerMap interface{}) error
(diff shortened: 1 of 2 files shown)


mcastelino:
@markdryan could you review this please.

networking/libsnnet/docker_plugin.go
```diff
@@ -181,12 +197,66 @@ type DockerEpMap struct {
...
+}
+
+//NewElement allocates and returns an endpoint value
+func (d *DockerEpMap) NewElement() interface{} {
+	return &DockerEpVal{}
+}
+
+//Add adds a value to the map with the specified key
+func (d *DockerEpMap) Add(k string, v interface{}) error {
+	switch val := v.(type) {
```
markdryan: You could use a checked type assertion here instead, which would save you one line of code

```
val, ok := v.(*DockerEpVal)
if !ok {
    return fmt.Errorf("Invalid value type %t", v)
}
d.m[k] = val

**Comment by markdryan** [comment]
@mcastelino Looks good. One small comment on the code but it's stylistic so feel free to ignore.

I can't think of any better way to do this unless you were to just use a map of interface{} for all types, casting when you need to read the values from the maps, and passing either a reflect.Type into DbTableRebuild or an interface{} initialised to contain the correct type, a copy of which would then be passed to gob.Decode.

```
var typeHolder *d.DockerNwMap
d.DbMapRebuild(tableNetworkMap, &d.DockerNwMap, typeHolder)
```
The final parameter in DbMapRebuild here is an interface{}.

DockerNwMap would embed a generic structure containing the Mutex and a map[string]interface{} map and a method for reading a value from the map that performs the cast. You might also want to provide a setter to get type checking when writing to the map.

I'm not sure this is any better (and it may not work, I haven't tried it) but it might result in slightly less code and less work when adding a new type. On the other hand, you'll have to use Setters and Getters to write and read to the map.  Well you don't have to, but if you don't, you'll have casts all over the place, which you might get wrong,i.e, forget the *.

networking/libsnnet/docker_plugin.go
mcastelino: will fix this.. shorter the better

PR #651, opened 2016-10-07
Refactor ciao certificate generation

rbradford:
This PR forms part of the pre-work of a simpler deployment methodology. Through the commits in this PR the code for creating certificates is refactored out from ciao-cert and into the ssntp/certs package.

### DIFF
--- a/ciao-cert/main.go
+++ b/ciao-cert/main.go
@@ -26,26 +26,17 @@
 package main
 
 import (
-	"crypto/ecdsa"
-	"crypto/elliptic"
-	"crypto/rand"
-	"crypto/rsa"
 	"crypto/x509"
-	"crypto/x509/pkix"
-	"encoding/asn1"
 	"encoding/pem"
 	"flag"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"log"
-	"math/big"
-	"net"
 	"os"
 	"strings"
-	"time"
 
 	"github.com/01org/ciao/ssntp"
+	"github.com/01org/ciao/ssntp/certs"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


rbradford:
@sameo Is this API close to what you had in mind?

sameo:
@rbradford Yes, this looks really good.

ssntp/certs/certs.go
```diff
@@ -0,0 +1,240 @@
...
+	var priv interface{}
+	var err error
+
+	if ell == false {
+		priv, err = rsa.GenerateKey(rand.Reader, 2048)
+	} else {
+		priv, err = ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+	}
+
+	if err != nil {
```
markdryan: Maybe we could make this function simpler.  If GenerateKey always sets priv to nil when there is an error, which it really should, we could just write.

```
func generatePrivateKey(ell bool) (interface{}, error) {
      if ell == false {
		return rsa.GenerateKey(rand.Reader, 2048)
	} else {
		return ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
	}
}
```

ssntp/certs/certs.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// Create certificate signed by private key from serverCert
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, parentCert, publicKey(priv), serverPrivKey)
+	if err != nil {
+		return fmt.Errorf("Unable to create client certificate: %v", err)
+	}
+
+	// Write out certificate (including private key)
+	pem.Encode(certOutput, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
```
markdryan: We should probably check the error return by Encode here.

ssntp/certs/certs.go
```diff
@@ -0,0 +1,240 @@
...
+	derBytes, err := x509.CreateCertificate(rand.Reader, template, template, publicKey(priv), priv)
+	if err != nil {
+		return fmt.Errorf("Unable to create server certificate: %v", err)
+	}
+
+	// Write out CA cert
+	pem.Encode(caCertOutput, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+
+	// Write out certificate (including private key)
+	pem.Encode(certOutput, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
```
markdryan: We need to check the error here.

ssntp/certs/certs.go
```diff
@@ -0,0 +1,240 @@
...
+//    golang/src/pkg/crypto/tls/generate_cert.go
+//
+// which is:
+//
+//    Copyright 2009 The Go Authors. All rights reserved.
+//    Use of this source code is governed by a BSD-style
+//    license that can be found in the golang LICENSE file.
+//
+
+package certs
```
markdryan: I think we're going to need some unit tests for all this stuff, particularly as we are adding a new package.

ssntp/certs/certs.go
```diff
@@ -0,0 +1,240 @@
...
+
+func pemBlockForKey(priv interface{}) *pem.Block {
+	switch k := priv.(type) {
+	case *rsa.PrivateKey:
+		return &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(k)}
+	case *ecdsa.PrivateKey:
+		b, err := x509.MarshalECPrivateKey(k)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Unable to marshal ECDSA private key: %v", err)
+			os.Exit(2)
```
rbradford: I fixed a bunch of places in the code to return errors. Look like I missed this one. Fixed!

ssntp/certs/certs.go
rbradford: Fixed!

ciao-cert/main.go
```diff
@@ -202,138 +122,58 @@ func checkCompulsoryOptions() {
...
-			serverPrivKey, err = x509.ParsePKCS1PrivateKey(privKeyBlock.Bytes)
-		} else {
-			serverPrivKey, err = x509.ParseECPrivateKey(privKeyBlock.Bytes)
-		}
+		err = certs.CreateClientCert(template, *isElliptic, bytesCert, certOut)
 		if err != nil {
-			log.Fatalf("Could not get server private key %s", err)
+			log.Fatalf("Failed to create certificate: %v", err)
 		}
+		certOut.Close()
```
rbradford: Ditto!

ssntp/certs/certs.go
rbradford: All pem.Encode returns now error checked.

markdryan:
Passed Single VM verify.sh and BAT tests.

PR #653, opened 2016-10-10
Single VM: Fix a ceph related disk space leak

markdryan:
The ceph-demo docker container creates a new volume each time the container
is started.  This volume needs to be explicitly deleted.  It does not get
deleted when the container is deleted.  Overtime, this can lead to a
substantial waste of disk space.  This commit tells docker to delete the
volume created by the ceph-demo container when it itself is deleted.


mcastelino:
LGTM

PR #656, opened 2016-10-10
single-vm: move creation of ciao_bin directory

kaccardi:
Don't duplicate the mkdir of the ciao_bin directory, and instead
move the call to make the directory to the top of the script.


markdryan:
@kaccardi Thanks for this and sorry for breaking Single VM.  Tested with and without a ~/local directory.  Works fine.

PR #664, opened 2016-10-12
Single VM: Enable the image service

markdryan:
This commit enables the image service in Single VM.  It also modifies Single VM to use the image service to manages the example images that it downloads.  Changes to the fake identity service, ciao-cli and the image service itself were required to enable this.

### DIFF
--- a/ciao-image/service/service.go
+++ b/ciao-image/service/service.go
@@ -39,8 +39,21 @@ type ImageService struct {
 func (is ImageService) CreateImage(req image.CreateImageRequest) (image.DefaultResponse, error) {
 	// create an ImageInfo struct and store it in our image
 	// datastore.
+	id := req.ID
+	if id == "" {
+		id = uuid.Generate().String()
+	} else {
+		if _, err := uuid.Parse(id); err != nil {
+			return image.DefaultResponse{}, image.ErrBadUUID
+		}
+
+		if _, err := is.ds.GetImage(id); err == nil {
+			return image.DefaultResponse{}, image.ErrAlreadyExists
+		}
+	}
+
 	i := datastore.Image{
-		ID:         uuid.Generate().String(),
+		ID:         id,
 		State:      datastore.Created,
 		Name:       req.Name,
 		CreateTime: time.Now(),
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -10,6 +10,7 @@ fi
 ciao_gobin="$GOPATH"/bin
 sudo killall ciao-scheduler
 sudo killall ciao-controller
+sudo killall ciao-image
 sudo killall ciao-launcher
 sleep 2
 sudo "$ciao_gobin"/ciao-launcher --alsologtostderr -v 3 --hard-reset
@@ -18,3 +19,6 @@ sudo pkill -F /tmp/dnsmasq.macvlan0.pid
 sudo mv $hosts_file_backup /etc/hosts
 sudo docker rm -v -f ceph-demo
 sudo rm /etc/ceph/*
+sudo rm /etc/pki/ciao/controller_key.pem /etc/pki/ciao/ciao-image-key.pem
+sudo rm /etc/pki/ciao/controller_cert.pem /etc/pki/ciao/ciao-image-cacert.pem
+
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -21,4 +21,7 @@ sudo docker rm -v -f ceph-demo
 sudo rm /etc/ceph/*
 sudo rm /etc/pki/ciao/controller_key.pem /etc/pki/ciao/ciao-image-key.pem
 sudo rm /etc/pki/ciao/controller_cert.pem /etc/pki/ciao/ciao-image-cacert.pem
+sudo rm /var/lib/ciao/images/4e16e743-265a-4bf2-9fd1-57ada0b28904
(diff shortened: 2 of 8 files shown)


markdryan:
@mcastelino Is this change okay with you?

testutil/singlevm/cleanup.sh
```diff
@@ -21,4 +21,7 @@ sudo docker rm -v -f ceph-demo
 sudo rm /etc/ceph/*
 sudo rm /etc/pki/ciao/controller_key.pem /etc/pki/ciao/ciao-image-key.pem
 sudo rm /etc/pki/ciao/controller_cert.pem /etc/pki/ciao/ciao-image-cacert.pem
+sudo rm /var/lib/ciao/images/4e16e743-265a-4bf2-9fd1-57ada0b28904
```
mcastelino: @markdryan why are we doing this directly? Just to cleanup prior artifacts? I notice you moved the copy to the cli method

mcastelino:
@markdryan LGTM. Also tested on FC24

testutil/singlevm/setup.sh
```diff
@@ -146,6 +147,8 @@ openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout controller_key.pem -
 #Copy the certs
 sudo cp -f controller_key.pem /etc/pki/ciao
 sudo cp -f controller_cert.pem /etc/pki/ciao
+sudo ln -s /etc/pki/ciao/controller_key.pem /etc/pki/ciao/ciao-image-key.pem
+sudo ln -s /etc/pki/ciao/controller_cert.pem /etc/pki/ciao/ciao-image-cacert.pem
```
tpepper: Is there a reason these are symlinked?  As opposed to creating a unique/separate key and cacert for the image service?

testutil/singlevm/setup.sh
```diff
@@ -305,6 +294,23 @@ echo "$identity" >> "$ciao_env"
 identity_url=$(echo $identity | sed 's/^.*=//')
 "$ciao_bin"/run_image.sh $identity_url &> /dev/null
 
+sleep 1
```
markdryan: No right now.  The is just one of the many race conditions in single vm that we'll need to address at some point.  The problem is that without the sleep the ciao-cli image commands that follow will fail as they will be executed before the ciao-image service has started.  We might be able to fix this if we use systemd to launch the ciao services but this will need to be done as part of a bigger change.

ciao-image/service/service.go
```diff
@@ -39,8 +39,15 @@ type ImageService struct {
 func (is ImageService) CreateImage(req image.CreateImageRequest) (image.DefaultResponse, error) {
 	// create an ImageInfo struct and store it in our image
 	// datastore.
+	id := req.ID
+	if id == "" {
+		id = uuid.Generate().String()
+	} else if _, err := uuid.Parse(id); err != nil {
+		return image.DefaultResponse{}, image.ErrBadUUID
+	}
```
markdryan: Yes I think you're right.   I'll add this tomorrow.

testutil/singlevm/setup.sh
markdryan: Well @rbradford and I discussed this and we decided that it wasn't worth the extra effort of creating an extra certificate.  Does it really buy us anything in Single VM to use separate certs?

markdryan: Doing so would also complicate things I think.  I may be mistaken but I think using separate certs for ciao-image and ciao-controller would require us to create three certs, a cert for controller, a cert for ciao-image and a CA cert that is used to sign the other two certs that can be given to ciao-cli.

tpepper: Since cert management is complicated, and good security depends on it, I like the idea of the single VM modelling best practices and "real" workflows.  Otherwise we'd not call ciao-cert or openssl at all for efficiency and just checkout a pre-canned set of test-keys like some other OS's and build systems do.

tpepper:
I'm going to hold off merging until at least the additional validation check happens on the user input uuid in ciao-image/service/service.go

testutil/singlevm/cleanup.sh
markdryan: Cephfs is not enabled in Single VM.  There's no point as there is only one machine so we don't need a network file system and enabling it will just slow things down.  The image service doesn't care whether cephfs is running or not.

For the sake of experimentation I did try enabling cephfs in Single VM locally.  It does work but it slows down the setup.sh script.  In particular you need to wait 10 seconds after you start the  ceph docker container before you can mount cephfs.

markdryan: Right now there's no point in starting ciao-image in cleanup.sh as it doesn't have any persistent storage.  Attempts to delete the images via ciao-cli would fail as the newly started image service wouldn't know anything about them, i.e., it would think that there are 0 images.

However, we could also create a new teardown.sh, which could be used to cleanup a correctly initialised cluster using ciao-cli, i.e, 

ciao-cli image delete -all
ciao-cli volume delete -all
ciao-cli instance delete -all.

The existing cleanup.sh would be continue to be used to cleanup the cluster in case of error or interruption, using ciao-launcher --hard-reset, ciao-image --hard-reset.  But this would be done is a separate PR.

ciao-image/service/service.go
tpepper: Thanks!

testutil/singlevm/cleanup.sh
tpepper: Yeah...future PR/enhancement eventually.  In the meantime you've got the pragmatic option here.

PR #685, opened 2016-10-14
Single VM: Check docker network connectivity

mcastelino:
Check docker network connectivity vs running the performance
benchmark. This matches what we do for virtual machines.

Fix some issues in the verify script where it was not reporting
failures with an exit code.


rbradford:
Tested on my Ubuntu 16.04 single VM test.

PR #687, opened 2016-10-15
Provide a generic database implementation.

mcastelino:
Make the networking database implementation generic so that it can be used to store data for any component in Ciao. This database implementation can be used to persist any table map, array, slice to a database backed by boltDB+gob.

We may want to consider moving this to a generic util package for ciao.

    **Note: As the table name internally has changed, this will break
    any in place update of the CN launcher. For a proper update
    of the CN, it has to be hard reset to remove all instances
    and the database cleaned up.**

### DIFF
--- a/networking/libsnnet/docker_database.go
+++ b/networking/libsnnet/docker_database.go
@@ -28,30 +28,62 @@ import (
 	"github.com/golang/glog"
 )
 
-const (
-	dbFile = "docker_plugin.db"
-)
+//DbTable interface that needs to be supported
+//for the table to be handled by the database
+type DbTable interface {
+	// Creates the backing map
+	NewTable()
 
-type dockerBoltDB struct {
+	// Name of the table as stored in the database
+	Name() string
+
+	// Allocates and returns a single value in the table
+	NewElement() interface{}
+
+	// Add an value to the in memory table
+	Add(k string, v interface{}) error
+}
+
+// A TableDBProvider represents a persistent data base provider
+// that can be used to store map, arrays or any other type of
+// tables into a database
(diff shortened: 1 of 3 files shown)


mcastelino:
@markdryan Could you review this please.

markdryan:
@mcastelino , @obedmr We have almost identical database abstractions in this PR and in #632. We need to figure out which one to use, and also where to put it.

obedmr:
@markdryan, as far I know we'd be using the one we're creating at #632, once it's done, a new PR would come for updating this networking code, there's only one missing piece that @mcastelino needs there in order to use it. @mcastelino could you confirm?

mcastelino:
@obedmr It does not seem a completely transparent replacement right now. We can consider moving to your final implementation once you merge in. Also right now I need this to implement CNCI agent persistence.

mcastelino:
@obedmr also in your implementation does the user of the database need to know that gob is used to store the value? I am going through your implementation, maybe I am mistaken, but I think the user of the db would need to decode?

networking/libsnnet/docker_database.go
```diff
@@ -28,30 +28,62 @@ import (
...
+	Name() string
+
+	// Allocates and returns a single value in the table
+	NewElement() interface{}
+
+	// Add an value to the in memory table
+	Add(k string, v interface{}) error
+}
+
+// A TableDBProvider represents a persistent data base provider
```
mcastelino: @rbradford Good catch. fixed.

mcastelino:
@rbradford fixed the review comments.

PR #695, opened 2016-10-17
testutil: singlevm: use hostname to get primary interface

kaccardi:
use hostname -i to get the ciao ip address.

### DIFF
--- a/testutil/singlevm/setup.sh
+++ b/testutil/singlevm/setup.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 ciao_host=$(hostname)
-ciao_ip=$(ip route get 8.8.8.8 | head -1 | cut -d' ' -f8)
+ciao_ip=$(hostname -i)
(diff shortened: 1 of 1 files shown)


markdryan:
@mcastelino Does this look okay to you?

rbradford:
Not sure how this PR will interact with #696 as from the man page

>        -i, --ip-address
              Display the network address(es) of the host name. Note that this works only if the host name can be resolved. Avoid  using
              this option; use hostname --all-ip-addresses instead.

It could be that the change to modify /etc/hosts is giving the correct forward resolution that makes this work. Deferring to @mcastelino

mcastelino:
@rbradford good observation. I have seen this fail sometimes... I think I have an idea to address this whole linking to the host business.. I will update https://github.com/01org/ciao/pull/695 to add this.

This make the single vm fully independent of the host (except the hostname)

mcastelino:
@kaccardi @rbradford fixed this issue with an alternate implementation which eliminates the need to use the host IP completely.  https://github.com/01org/ciao/pull/696

Once that PR gets merged, I think we can drop this one?

kaccardi:
ok, I'll close this pull request.

PR #696, opened 2016-10-17
Single VM: Eliminate the need to modify /etc/hosts or use host IP

mcastelino:
Specify the IP address when generating the TLS certificates. This allows for successful reverse TLS DNS verification. This eliminates the need to modify the /etc/hosts file

Also delink the scripts from using the host IP address and use the virtual IP and interface setup by the scripts.


mcastelino:
@sameo @kaccardi can you review this please.

mcastelino:
@obedmr, can you also confirm if this fixes the reason you had to backup and restore /etc/hosts?

testutil/singlevm/cleanup.sh
```diff
@@ -1,11 +1,6 @@
 #!/bin/bash
 
-if [ ! -z $1 ]; then
-    hosts_file_backup=$1
-else
-    . ~/local/demo.sh
-    hosts_file_backup=$HOSTS_FILE_BACKUP
-fi
+. ~/local/demo.sh
```
rbradford: Is this necessary? I don't thing anything in this script actually depends on any of the env variables in the that file any longer.

mcastelino: @rbradford leaving it in place now, as it may be needed if we want to use the image service interface to cleanup the images vs just rm as we do today.

rbradford:
Tested on Ubuntu 16.04 after running cleanup on previous tree.

PR #697, opened 2016-10-19
test-cases, travis: Add support for TAP reports

markdryan:
This commit adds test-cases support for the generation of TAP reports. It also takes
the oppurtunity to rationalize the options for specifying the output format.  There
is now one format option, --format, which accepts a number of different values;
html, tap, colour-text and text.  The default is colour-text.

This change requires a modification to the travis script that executes test-cases,
as the old formatting option used by travis.yml no longer exists.

### DIFF
--- a/test-cases/test-cases.go
+++ b/test-cases/test-cases.go
@@ -135,32 +135,55 @@ const htmlTemplate = `
 </html>
 `
 
+var newLineRegexp = regexp.MustCompile(`(\r\n)|[\n\r]`)
+
+type formatType string
+
+const (
+	formatText       formatType = "text"
+	formatColourText            = "colour-text"
+	formatHTML                  = "html"
+	formatTAP                   = "tap"
+)
+
+func (f *formatType) String() string {
+	return string(*f)
+}
+func (f *formatType) Set(val string) error {
+	v := formatType(val)
+	if v != formatText && v != formatColourText && v != formatHTML && v != formatTAP {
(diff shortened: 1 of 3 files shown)


test-cases/test-cases.go
```diff
@@ -672,6 +695,33 @@ func runTests(tests []*PackageTests) (int, *bytes.Buffer, error) {
...
+			} else {
+				testName = newLineRegexp.ReplaceAllString(testName, " ")
+			}
+			fmt.Printf("%d - %s\n", i+1, testName)
+			i++
+		}
+	}
+	if i > 0 {
+		fmt.Printf("1..%d\n", i)
+	}
```
rbradford: In a valid tap file isn't this before the test results?

markdryan: After is okay as well.  Outputting at the bottom here as this is what the release_test.py script does.  The goal is to replace this script without breaking the testing infrastructure built around it.

test-cases/test-cases.go
```diff
@@ -135,32 +135,55 @@ const htmlTemplate = `
 </html>
 `
 
+var newLineRegexp = regexp.MustCompile(`(\r\n)|[\n\r]`)
+
+type formatType string
+
+const (
+	formatText       formatType = "text"
+	formatColourText            = "colour-text"
```
markdryan: Test-cases has always been a UK English application.  This commit is replacing a --colour option with a --format=colour-text, so it's just maintaining the status quo.

PR #699, opened 2016-10-19
ciao-controller: api: Set 'limit' value properly for instances

### LINKED ISSUE #698: ciao-controller api: pager doesn't limit number of instances returned
When calling /{tenant}/servers/detail?limit=value&offset=value with query parameters 'limit' and 'offset' the "limit" value is not working properly.

Expected behavior is to returned the number of registries (instances) specified in limit.

Actual behavior is that ciao returns all instances. Which is currently affecting the paging feature on ciao-webui.

jorgevgut:
Fixes #698 Set limit value when a pager is being used on servers/detail endpoint.
Returns correct number of instances in order for the pager feature to work on
ciao-webui.

Signed-off-by: Jorge Villalobos <jorge.villalobos.gutierrez@intel.com>

### DIFF
--- a/openstack/compute/api.go
+++ b/openstack/compute/api.go
@@ -288,12 +288,22 @@ func pagerQueryParse(r *http.Request) (int, int, string) {
 	// we only support marker and offset for now.
 	if values["marker"] != nil {
 		marker = values["marker"][0]
-	} else if values["offset"] != nil {
-		o, err := strconv.ParseInt(values["offset"][0], 10, 32)
-		if err != nil {
-			offset = 0
-		} else {
-			offset = (int)(o)
+	} else {
+		if values["offset"] != nil {
+			o, err := strconv.ParseInt(values["offset"][0], 10, 32)
+			if err != nil {
+				offset = 0
+			} else {
+				offset = (int)(o)
+			}
+		}
+		if values["limit"] != nil {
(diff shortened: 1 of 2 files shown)


markdryan:
It also looks like travis is complaining about the formatting.  Could you run go fmt and resubmit?

jorgevgut:
Sure @rbradford @markdryan I'm on it!

jorgevgut:
I have fixed formatting and added a couple of tests to verify the pager query func works properly (it's the one that was failing). Inlcuding offset and limit

tpepper:
The unit test is sufficient I suppose.  It would be nice to have a little more stress/variation/coverage of the underlying function's conditional behaviors and edges, but that's a criticism for a large part of our unit tests.

@jorgevgut ideally the format change would not be in the second patch, but rather an amended version of the first patch.  If you "git format-patch -1", that'll save your "top" one patch, then you can git reset --hard back to the commit id of your first patch.  Edit the one file to remove the space that fmt wasn't liking.  Then git add the edited file.  Then "git commit --amend".  Then "git am" the one patch you'd previously saved.  Next git push --force to your branch.  This PR will notice the push and that the first patch is superseded by a newer one and we'll get a nice stack of two distinct patches that are merge ready.

jorgevgut:
Thanks for the tip @tpepper !! you're right, with the amend code history looks so much cleaner now!

Issue #676, opened 2016-10-14
ciao-cli instance list order is unhelpful


rbradford:
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP         SSH PORT
1 365c7ff0-0388-41e5-8730-31df0149fe9a active 172.16.0.4 10.103.239.150 33004
2 faf95e1d-4804-415a-bdcd-94ee7190f245 active 172.16.0.2 10.103.239.150 33002
3 fe55b055-dcca-4d69-9a7a-78c9f9cb5aaf active 172.16.0.3 10.103.239.150 33003

The number in the first column implies some kind of ordering. However in this example the first is the most recent, followed by the oldest and then then lastly is the the intermediate instance launched.

It might also be helpful to show the time that the instance was launched.

PR #701, opened 2016-10-20
Order instances in ciao-cli by creation time

rbradford:
Fixes: #676

### DIFF
--- a/ciao-controller/types/types.go
+++ b/ciao-controller/types/types.go
@@ -87,6 +87,7 @@ type Instance struct {
 	CNCI        bool                `json:"-"`
 	Usage       map[string]int      `json:"-"`
 	Attachments []StorageAttachment `json:"-"`
+	CreateTime  time.Time           `json:"-"`
 }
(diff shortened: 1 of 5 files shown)


ciao-controller/types/types.go
```diff
@@ -87,6 +87,7 @@ type Instance struct {
 	CNCI        bool                `json:"-"`
 	Usage       map[string]int      `json:"-"`
 	Attachments []StorageAttachment `json:"-"`
+	CreateTime  time.Time           `json:"create_time"`
```
kaccardi: why are we tagging this with a json tag? what api marshals this struct out?

rbradford: @markdryan and I couldn't find where that struct was marshalled out at all. I'll update the patch to not marshal this one and open an issue about removing the marshalling if not necessary.

kaccardi: it's probably left over from when we had the debug UI.

Issue #644, opened 2016-10-06
volume names are deleted when updating volume state


kaccardi:
I noticed that the name of the volume seems to disappear in ciao-cli volume list after it has been detached.

kaccardi:
check the json the cli receives to determine if this is a cli problem or controller issue.

markdryan:
Actually, this is not related to detaching volumes.  It seems to happen all the time.  In the example below I've added two volumes to two separate instances, one of which was a container ( so the attach should have failed) and one was a Fedora VM.  I did not specify a name when creating the first volume.  I did when specifying the second.  Here's the output of two back to back calls to ciao-cli volume list

```
# ciao-cli volume list
Volume #1
	Name             []
	Size             [1 GB]
	UUID             [f888ae99-b5ea-4274-b556-1eee3ddfd571]
	Status           [in-use]
	Description      []

Volume #2
	Name             [Container Test]
	Size             [1 GB]
	UUID             [c6ee5b28-97fd-438d-bf8e-6bbc0800ffac]
	Status           [attaching]
	Description      []

# ciao-cli volume list
Volume #1
	Name             []
	Size             [1 GB]
	UUID             [c6ee5b28-97fd-438d-bf8e-6bbc0800ffac]
	Status           [attaching]
	Description      []

Volume #2
	Name             []
	Size             [1 GB]
	UUID             [f888ae99-b5ea-4274-b556-1eee3ddfd571]
	Status           [in-use]
	Description      []
```

Note in the second call to ciao-cli volume list the name of the second volume is missing.

PR #703, opened 2016-10-20
ciao-controller: Fix volume detail "corruption" in ListVolumesDetail

rbradford:
Various members of the Volume structure require a pointer. These were
being filled by taking the address of the member from the loop variable.
Unfortunately because this loop variable is reused for each iteration of
the loop all these pointers were ultimately the same resulting in
duplicated values in the volume details list.

The solution is to copy the value to a local variable and take the
address of that.

Fixes: #644


ciao-controller/openstack_volume.go
```diff
@@ -297,14 +297,17 @@ func (c *controller) ListVolumesDetail(tenant string) ([]block.VolumeDetail, err
 		vol.ID = data.ID
 		vol.Size = data.Size
 		vol.OSVolTenantAttr = data.TenantID
-		vol.CreatedAt = &data.CreateTime
+		createdAt := data.CreateTime
+		vol.CreatedAt = &createdAt
```
tpepper: I get slightly confused after so many years in a non-garbage collected language, but:

This means we make a copy, then store a link to that copy.  Repeatedly.  But I assume the structure has pointers to avoid copies.  Would it then be more direct to not have the block.VolumeDetail structure use pointers for these members?

rbradford: Or better ... maybe I can iterate over the slice by index and take the addresses of the members in the original slice.

rbradford:
Updated this change to take addresses of members of the underlying structures.

Issue #689, opened 2016-10-17
ciao-cli volume list order is unhelpful


markdryan:
The order in which volumes are output by the ciao-cli command is unhelpful and confusing.  Volumes should be ordered by some criteria, perhaps by size of creation time.  Here's an example:

```
# ciao-cli volume list
Volume #1
    Name             []
    Size             [1 GB]
    UUID             [f888ae99-b5ea-4274-b556-1eee3ddfd571]
    Status           [in-use]
    Description      []

Volume #2
    Name             [Container Test]
    Size             [1 GB]
    UUID             [c6ee5b28-97fd-438d-bf8e-6bbc0800ffac]
    Status           [attaching]
    Description      []

# ciao-cli volume list
Volume #1
    Name             []
    Size             [1 GB]
    UUID             [c6ee5b28-97fd-438d-bf8e-6bbc0800ffac]
    Status           [attaching]
    Description      []

Volume #2
    Name             []
    Size             [1 GB]
    UUID             [f888ae99-b5ea-4274-b556-1eee3ddfd571]
    Status           [in-use]
    Description      []
```

markdryan:
Related to #676 and #688

Issue #702, opened 2016-10-20
storage api: Support paging for listing volumes


jorgevgut:
Support pagination on listVolumes and listVolumesDetail functions on block api.
This will enable UI to correctly display multipages, which will increase user experience when high number of volumes are listed.

PR #705, opened 2016-10-20
ciao-cli: Report volumes in creation time order

rbradford:
This change takes advantage of the CreatedAt already populated in the
volume details as part of the openstack API.

Fixes: #689

### DIFF
--- a/ciao-cli/volume.go
+++ b/ciao-cli/volume.go
@@ -20,6 +20,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sort"
+	"time"
 
 	"github.com/rackspace/gophercloud"
 	"github.com/rackspace/gophercloud/openstack"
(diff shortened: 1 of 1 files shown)


kaccardi:
I know ciao-cli doesn't currently have a single unit test. Is it possible to start adding them as we add new functionality?

markdryan:
@kaccardi I think the unit testing needs to be done separately.  None of the ciao-cli functions are designed to be unit tested so they'd need to be rewritten.  As this PR is just modifying an existing function and not adding a new one, writing the function in question to be unit testable would obscure the purpose of this PR.  I suggest opening a separate ticket for ciao-cli unit testing if one does not already exist.

markdryan:
Strictly speaking the sorting should probably be done in the controller.  The openstack API supports this.  Unfortunately, gophercloud does not, so there's no easy way for us to do this right now, apart from submitting a PR to gophercloud and waiting.   The current randomness of the volume list command is quite confusing right now so it makes sense to me to merge this PR as is, with the proviso that we will need to revisit the code when #702 is implemented.

Issue #690, opened 2016-10-17
ciao-cli should be scriptable


markdryan:
Ciao-cli should accept a -f <go-template> parameters that allows callers to format the output of a ciao-cli query command in the manner that most suits their use cases, e.g.,

ciao-cli volume list -f '{{.UUID}}: "{{.Name}}"'

might produce the following output

fcf32992-0feb-4435-ad3a-92c923c5c6b7: data
503d45be-5e44-4fc1-a3ec-3851b43fe7df: media

The ciao-cli help needs to be updated with the Go structures expose by each command.  See go list for more details.

markdryan:
This would be very useful for the BAT tests which currently parse the output of ciao-cli.

PR #731, opened 2016-10-21
ciao-cli: Make ciao-cli scriptable

markdryan:
This commit adds a -f parameter that accepts a go template to the following
commands:

- event list
- image list
- tenant list
- instance list
- instance show
- node list
- node show
- tenant list
- tenant show
- volume list
- volume show
- workload list

As an example, to print the name of the workload with an ID of
ca957444-fa46-11e5-94f9-38607786d9ec on might type

ciao-cli workload list -f '
{{- range .}}
  {{- if eq .ID "ca957444-fa46-11e5-94f9-38607786d9ec" }}
    {{- .Name | printf -}}
  {{end -}}
{{end }}
'

The output would be

Docker Debian latest

Fixes: #690

### DIFF
--- a/ciao-cli/README.md
+++ b/ciao-cli/README.md
@@ -291,3 +291,89 @@ $GOBIN/ciao-cli -username admin -password ciao event list -all
 ```shell
 $GOBIN/ciao-cli event list
 ```
+
+## Scripting with ciao-cli
+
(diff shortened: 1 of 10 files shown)


markdryan:
@tpepper Good point.  I've updated the README.md.

PR #735, opened 2016-10-21
Add missing bus parameter for a CharDevice

sboeuf:
When creating a CharDevice, we need to add a "bus" parameter so that it can match the serial pci device previously created.


tpepper:
Since it's optional, let's condition it on if cdev.Bus != ""

tpepper:
gofmt caught some spaces instead of tabs...trivial fix

PR #739, opened 2016-10-25
send volume attachments with server info

### LINKED ISSUE #654: ciao-cli instance show does not show volumes
I created a boot from volume instance.  The instance was created and booted correctly.  But when I did a ciao-cli instance show, no information about the volume was listed.  I'd expect to see the volumes attached to an instance reported by this command.

kaccardi:
Allow ciao-cli to report the attached volumes when displaying instance details.

Fixes #654

### DIFF
--- a/ciao-cli/instance.go
+++ b/ciao-cli/instance.go
@@ -54,6 +54,7 @@ const (
 	}
 	SSHIP   string                                // Instance SSH IP address
 	SSHPort int                                   // Instance SSH Port
+	OsExtendedVolumesVolumesAttached []string     // list of attached volumes
 }`
 )
 
@@ -505,6 +506,10 @@ func dumpInstance(server *compute.ServerDetails) {
 	if server.SSHIP != "" {
 		fmt.Printf("\tSSH IP: %s\n", server.SSHIP)
 		fmt.Printf("\tSSH Port: %d\n", server.SSHPort)
+	}
+
+	for _, vol := range server.OsExtendedVolumesVolumesAttached {
+		fmt.Printf("\tVolume: %s\n", vol)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ciao-cli/instance.go
```diff
@@ -506,6 +506,10 @@ func dumpInstance(server *compute.ServerDetails) {
 		fmt.Printf("\tSSH IP: %s\n", server.SSHIP)
 		fmt.Printf("\tSSH Port: %d\n", server.SSHPort)
 	}
+
+	for _, vol := range server.OsExtendedVolumesVolumesAttached {
+		fmt.Printf("\tVolume: %s\n", vol)
+	}
```
rbradford: @markdryan  has some feedback on this wrt. to his template changes.

ciao-cli/instance.go
```diff
@@ -506,6 +506,10 @@ func dumpInstance(server *compute.ServerDetails) {
 		fmt.Printf("\tSSH IP: %s\n", server.SSHIP)
 		fmt.Printf("\tSSH Port: %d\n", server.SSHPort)
 	}
+
+	for _, vol := range server.OsExtendedVolumesVolumesAttached {
```
markdryan: We also need to update the instanceTemplateDesc constant.  Otherwise users who type ciao-cli instance list --help will not see that the new field, OsExtendedVolumesVolumesAttached, is available to templates.

kaccardi: @markdryan please see updated commit.

ciao-cli/instance.go
```diff
@@ -54,6 +54,7 @@ const (
 	}
 	SSHIP   string                                // Instance SSH IP address
 	SSHPort int                                   // Instance SSH Port
+	OsExtendedVolumesVolumesAttached []string     // list of attached volumes
```
tpepper: I'm surprised the CI tools don't complain about a stutter here.  Is "OsExtendedVolumesAttached" sufficiently descriptive or do we need the 2x Volumes?

kaccardi: @tpepper it's named the way it is to match the openstack variable name as defined in their spec.

Issue #645, opened 2016-10-06
volume attach to a container doesn't fail properly


kaccardi:
Attaching a volume to a container should fail but it doesn't.  No error is reported and the volume is perpetually stuck in the Attaching state.  Launcher seems to be sending an error so this may be a problem with controller

markdryan:
Here's an example work flow.

```
# ciao-cli workload list
Workload 1
	Name: Fedora 23 Cloud
	UUID:69e84267-ed01-4738-b15f-b47de06b62e7
	Image UUID: 73a86d7e-93c0-480e-9c41-ab42f69b7799
	CPUs: 2
	Memory: 128 MB
Workload 2
	Name: Clear Cloud
	UUID:e35ed972-c46c-4aad-a1e7-ef103ae079a2
	Image UUID: df3768da-31f5-4ba6-82f0-127a1a705169
	CPUs: 2
	Memory: 128 MB
Workload 3
	Name: Docker Ubuntu latest
	UUID:ca957444-fa46-11e5-94f9-38607786d9ec
	Image UUID: fa7d86d8-fa46-11e5-8493-38607786d9ec
	CPUs: 2
	Memory: 128 MB
Workload 4
	Name: Docker Iperf
	UUID:ab68111c-03a6-11e6-87de-001320fb6e31
	Image UUID: b5e696b8-03a6-11e6-a424-001320fb6e31
	CPUs: 2
	Memory: 128 MB

# ciao-cli instance add --workload ca957444-fa46-11e5-94f9-38607786d9ec --instances 1
Created new instance: db2afe25-1f2f-43f4-b13b-e83edfa063f8

# ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 3bdf019a-c460-4deb-932d-131bc3ed0fd7 active 172.16.0.2 198.51.100.52 33002
2 db2afe25-1f2f-43f4-b13b-e83edfa063f8 active 172.16.0.3 198.51.100.52 33003

# ciao-cli volume add -name "Container Test"
Created new volume: c6ee5b28-97fd-438d-bf8e-6bbc0800ffac

# ciao-cli volume attach --instance db2afe25-1f2f-43f4-b13b-e83edfa063f8 --volume c6ee5b28-97fd-438d-bf8e-6bbc0800ffac
Attached volume: c6ee5b28-97fd-438d-bf8e-6bbc0800ffac

# ciao-cli volume list
Volume #1
	Name             []
	Size             [1 GB]
[...]

PR #740, opened 2016-10-25
ciao-controller: datastore: update attachments when deleting instance

kaccardi:
When we delete an instance, we should update the attachments list
to make sure any attachments are deleted and the volume state
is reset to available.

Possibly Fixes: #645

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1042,6 +1042,8 @@ func (ds *Datastore) deleteInstance(instanceID string) error {
 		glog.V(2).Info("deleteInstance: ", err)
 	}
 
+	ds.updateStorageAttachments(instanceID, nil)
+
 	return err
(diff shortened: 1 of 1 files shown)


kaccardi:
@rbradford it might be difficult to test till we get some of the other boot from volume PRs pushed.

tpepper:
Based on code inspection, this code change would mostly trigger when controller gets an Event Notify for ssntp.InstanceDeleted:

EventNotify()
  client.ctl.ds.DeleteInstance(event.InstanceDeleted.InstanceUUID)
    DeleteInstance()
      ds.deleteInstance(instanceID)

In that case launcher's sent that event, so it feels pretty innocuous for controller to update it's internal data.

There's also a number of launch failure cases which would trigger this, but I'm thinking those didn't hit launcher or would have had cleanup post-error.

Given that it fails safe to merge even ahead of a SingleVM test point.

PR #741, opened 2016-10-25
ciao-storage: do not use rbd import

kaccardi:
Instead of using rbd import to create volumes from images,
use qemu-img convert.

### DIFF
--- a/ciao-storage/ceph.go
+++ b/ciao-storage/ceph.go
@@ -42,7 +42,8 @@ func (d CephDriver) CreateBlockDevice(imagePath *string, size int) (BlockDevice,
 	// Currently the kernel rdb client only supports layering but in the future more feaures
 	// should be added as they are enabled in the kernel.
 	if imagePath != nil {
-		cmd = exec.Command("rbd", "--id", d.ID, "--image-feature", "layering", "import", *imagePath, ID)
+		rbdStr := fmt.Sprintf("rbd:rbd/%s:id=%s", ID, d.ID)
+		cmd = exec.Command("qemu-img", "convert", "-O", "rbd", *imagePath, rbdStr)
 	} else {
 		// create an empty volume
 		cmd = exec.Command("rbd", "--id", d.ID, "--image-feature", "layering", "create", "--size", strconv.Itoa(size)+"G", ID)


tpepper:
I did some experimenting yesterday and I can get drastic speed differences in rbd import and qemu-img convert depending on the -S sparse setting.  In all cases, after uploading the rbd and qemu-img reported size of the rbd image as the same.  It would be good to know if one or the other is faster because it transmits and stores less data when a sparse image is encountered.  It would be beneficial to intentionally not be packaging and shipping lots of zeros around.

kaccardi:
@tpepper This isn't about optimization, it's about functionality. rbd import doesn't create the filesystem the way that launcher needs it to. qemu-img convert actually takes the qcow and turns it into something launcher can boot. The change was required. I updated the commit message.

tpepper:
I'm able to boot images that are rbd imported.

kaccardi:
@markdryan will need to review and comment as he requested this change.

tpepper:
Based on what I understand of possible image formats, what I can infer from the semi-disjoint rbd import and qemu-img convert -O rbd, and what qemu seems to be able to do with rbd -drive...we may as well merge this despite the lingering concerns I have.  We're liable to swizzle this stuff more over time and what's here is one way of having things function.  The commit message does capture the idea that there are different formats that need handled differently or functionality wont be present.

markdryan:
@tpepper  The nice thing about using qemu-img is that it gives us support for all the following formats for free:  `bochs vvfat rbd vpc parallels tftp ftp ftps raw https qcow dmg http qcow2 quorum null-aio cloop vdi iscsi null-co vhdx blkverify file vmdk host_cdrom blkdebug host_device sheepdog qed nbd`

It's also safer as it ensures that the image we're storing in rbd image is in raw format.  If we just do an rbd import, how do we know what we're importing is in RAW format and will ultimately boot?

Also, if we just use rbd import we will need to manually convert all our images to raw format before we can boot from them.  This would not work right now on single VM for instance, which pulls down qcow images for some workloads e.g., clear.

tpepper:
@markdryan that's precisely the larger design question here?  All the docu says only use raw so a pre-conversion at some point is needed if we want to do that.  Doing it out of band relative to the launch will be performance important.  But it means we would only store images as raw and would only launch raw images.  If we declare our backing store is raw only, our front end utilities should automatically convert user uploads and inform the user we're doing that for them, instead of making them manually do it.

PR #744, opened 2016-10-25
Add storage workload for testing purposes

kaccardi:
This PR must not be merged until all the other boot from volume prs are merged. It adds a test workload to create a bootable volume based on an image. Without the other PRs merged, the bat tests will fail.


rbradford:
I've tested this against single VM and a full cluster (combined with #741)

rbradford:
I can't seem to be able to SSH into the instances created with the new workload. I'm trying to get qemu console so I can see if there is anything obvious.

kaccardi:
merge after #741

_release/bat/release_test.py
```diff
@@ -405,10 +405,6 @@ def test_start_all_workloads(self):
         """Start one instance of all workloads"""
         self.failUnless(launch_all_workloads())
 
-    def test_start_all_workloads10(self):
```
tpepper: My comment on this from the prior PR is lost since it was closed.  I've opened https://github.com/01org/ciao/issues/749 to track re-adding a test case to launch a herd of cattle in BAT.

Issue #747, opened 2016-10-26
ciao-cli makes http requests before parsing commands


rbradford:
This means that if you have slow network connection it can take a long time to get feedback that you've made an error in your chosen command line.

PR #748, opened 2016-10-26
ciao-cli: Parse the subcommand before identifying with keystone

rbradford:
This change moves the execution of common code that talks to keystone
and validates the tenant setup to after the subcommand arguments are
parsed but before the command itself is run.

This means you can get instant feedback of command line parsing errors
without any network induced delay.

Fixes: #747

### DIFF
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -51,6 +51,7 @@ func (c *command) run(args []string) error {
 		c.usage(cmdName)
 	}
 	args = subCmd.parseArgs(args[2:])
+	prepareForCommand()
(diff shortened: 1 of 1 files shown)


markdryan:
Definitely a big improvement although I'm guessing it doesn't catch all errors ahead of time, e.g.,

ciao-cli instance add --instances 1

will fail as no workload is specified.  Looking at the code this error won't be caught until after prepareForCommand is called as the check for the missing workload is in run.  I guess to fix it completely we'd need to add an error to parseArgs or add a new validateArgs method to subCommand and move the argument checks out of run.

Issue #652, opened 2016-10-07
ciao-cli expects StatusAccepted for instance Delete, when it should expect NoContent


kaccardi:
The OpenStack API for deleting an instance should return 204, StatusNoContent according to http://developer.openstack.org/api-ref-compute-v2.1.html#deleteServer. ciao-cli returns an error for instance delete if it does not return 200. Fix ciao-cli to expect the correct response.

PR #750, opened 2016-10-26
ciao-cli: expect NoContent from instance delete

kaccardi:
The api for deleting a server response with 204 in the case
of success, not 200.

Fixes: #652


tpepper:
Looks good to me and @markdryan confirmed it tests good for him.

Issue #746, opened 2016-10-26
Unable to delete instance immediately after creation


markdryan:
There seems to be a race condition with instance creation.  When writing the BAT tests I create an instance and then try to delete it straight away without first waiting for that instance to become active.  When a I do this I get the following error from ciao-cli:

```


```

If I insert a sleep of 2 seconds between instance creation and deletion I don't get the error.

kaccardi:
@markdryan I don't think I'd call this a bug - it is by design. You can't delete an instance that's "pending". You have to wait till it's either active, or exited or something. Allowing a delete prior to this opens up all kinds of race conditions. I also don't want to queue commands in controller for the same reason. So to me, this is operator error.

kaccardi:
@markdryan I do think status code 500 isn't the right response though.

markdryan:
But isn't this a race condition?  Currently, you need to wait an arbitrary amount of time before you can delete something?

kaccardi:
where would the race condition be? in any automated bat tests? yes - you have to keep checking and timeout after some predefined wait.

kaccardi:
the debug UI had this flow: get status (once time). If user requests a delete on an instance that at that moment in time active or exited, pop up error message and do not send the request. Then the user has to chose when to try again.

markdryan:
RIght.  That's where I'm seeing this.  In the BATs.  If this is by design feel free to close.  I can wait until I get a status in the BATs before deleting.  Although maybe we want to keep it open until we change the error code.

PR #754, opened 2016-10-27
ciao-controller: send ErrInstanceNotAssigned when needed

kaccardi:
When an operation is requested on an instance that is not yet
assigned to a node, send an explicit error message for this
condition from ciao-controller. In the openstack compute api,
detect this error message and send a 403 Forbidden error rather
than a 500 internal server error.

Fixes: #746


tpepper:
Lgtm

markdryan:
@kaccardi Sorry, I didn't get to this tonight.  I rebased my branch with the latest changes and now I'm having some issues.  I'll need to sort these out before I can re-create the test conditions needed to test the PR.

markdryan:
Tested the delete case.  The correct error is now returned.

PR #755, opened 2016-10-28
singlevm: Remove use of qemu-nbd to use loop device mountpoints

obedmr:
This implements a loop device based CNCI image mounting/handling
for CNCI certs and binary injection.
It removes `qemu*` dependencies, `losetup` and loop devices tools are
available by default on most of Linux distributions

Signed-off-by: Munoz, Obed N obed.n.munoz@intel.com

### DIFF
--- a/networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
+++ b/networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
@@ -63,19 +63,18 @@ fi
 
 echo -e "\nMounting image: $image"
 tmpdir=$(mktemp -d)
-sudo modprobe nbd max_part=63
-sudo qemu-nbd --format=raw -c /dev/nbd0 "$image" 
 
 #it can take some time for the device to get created
 retry=0
+loop=''
 until [ $retry -ge 3 ]
 do
-	sudo udevadm settle
-	sudo partprobe /dev/nbd0
-	sudo mount /dev/nbd0p$partition "$tmpdir" && break
-	let retry=retry+1
-	echo "Mount failed, retrying $retry"
-	sleep 10
+    if [ "$loop" == "" ]; then
+	loop=`sudo losetup -f --show -P $image`
+    fi
+    sudo mount ${loop}p$partition "$tmpdir" && break
+    let retry=retry+1
+    echo "Mount failed, retrying $retry"
 done
 
 if [ $retry -ge 3 ]
@@ -111,6 +110,6 @@ sudo rm -rf "$tmpdir"/var/lib/cloud
 #Umount
 echo -e "Done! unmounting\n"
 sudo umount "$tmpdir"
-sudo qemu-nbd -d /dev/nbd0
+sudo losetup -d $loop
 sudo rm -rf "$tmpdir"
 exit 0


markdryan:
@mcastelino Does this look okay to you?

networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
```diff
@@ -63,19 +63,18 @@ fi
 
 echo -e "\nMounting image: $image"
 tmpdir=$(mktemp -d)
-sudo modprobe nbd max_part=63
-sudo qemu-nbd --format=raw -c /dev/nbd0 "$image" 
 
 #it can take some time for the device to get created
 retry=0
+loop=''
 until [ $retry -ge 3 ]
```
mcastelino: @obedmr do we need the retry. I used to have it before for qemu-nbd. I do not think the loop device needs the re-try?

besides that this is a nice change to lower dependencies.

networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
```diff
@@ -63,19 +63,18 @@ fi
...
-	sudo udevadm settle
-	sudo partprobe /dev/nbd0
-	sudo mount /dev/nbd0p$partition "$tmpdir" && break
-	let retry=retry+1
-	echo "Mount failed, retrying $retry"
-	sleep 10
+    if [ "$loop" == "" ]; then
+	loop=`sudo losetup -f --show -P $image`
+    fi
+    sudo mount ${loop}p$partition "$tmpdir" && break
```
mcastelino: @obedmr do we need udevadm settle here? @tpepper ?

networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
obedmr: Sometimes, there's a latency issue that make it fail on first run, so there's a need of a second run.

mcastelino:
@obedmr @markdryan LGTM

markdryan:
There still seems to be one outstanding question about whether udevadm settle is needed or not.  @obedmr could you explain why this is no longer needed?

mcastelino:
@tpepper can you provide feedback on udevadm settle please

networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
tpepper: The udevadm settle shouldn't needed, nor the retry-loop.  But there's a race between finding a free loopdev, storing it in 'loop'.  Kpartx encapsulates logic to find and use a free loopdev, eg:

kpartx -sav $image
(parse the output to find which loopN was selected)
mount ${loop}p$partition "$tmpdir"  && break
echo "fatal/non-recoverable error got in the way"

tpepper: and later then for cleanup to not leak loopdevs:

kpartx -sdv $image

mcastelino: @obedmr can you address @tpepper comments. I will merge after that.

obedmr:
@markdryan @mcastelino @tpepper The use of kpartx adds another dependency for single VM usage, we remove qemu but we add kpartx, is that of for you?

tpepper:
I think I misread this earlier.  The losetup --show gets the dynamically used loopdev name into $loop. If that's failed, I think something bad has happened and doubt the retry is going to help.  But perhaps the retry helped in that the partition nodes were not in dev yet because that's asynchronous?

Regarding udevadm settle.  I didn't think it's needed, but it would be if the losetup causes device mapper to go asynchronously create /dev/mapper/loopNpM nodes.  I've never had the impression that is happening asynchronously, but it could be.  If so, the udevadm settle is needed.  But it MUST be after losetup and before mount to address the potential race were the node is not ready at mount time.

I would remove the retry loop and simply do losetup, settle, mount.  If that fails, something quite bad has happened.  That keeps our implementation nice and simple.  If we find we need more complex logic, I would pull in kpartx since it's got good retry/race handling logic and no point duplicating it.

tpepper:
I've created https://github.com/01org/ciao/pull/772 to address my concern.

PR #760, opened 2016-10-31
Rewrite BAT tests in GO

markdryan:
This PR rewrites the existing python based BAT tests in Go.  The Go tests are roughly
equivalent to the old BAT tests in release_test.py and are intended
to replace these tests.

For more information see the _release/bat/README.md file.

Going forward, BAT tests will be organised as follows:

- A new package under _release/bat will be created for each functional test
  area.  Currently, there is only one area, base, but more areas will be
  added soon.  Ultimately, we should have something like this:
  + _release/bat
    - base
    - storage
    - image
    - network
- A new general purpose package call bat has been created.  This package
  contains helper routines that are useful for writing BAT tests, e.g.,
  start and instance.  This package has been placed in the top level of
  the ciao repo as it is hoped that it may also be used one day in other
  parts of ciao, e.g., SingleVm.

The PR removes the old python BAT tests and makes a small change to ciao-cli node status command which was needed for one of the BAT tests.


markdryan:
This needs to be tested on the release cluster before it can be merged.  I'm also seeing occasional failures on Single VM.  I want to make sure they are not caused by the changes to the BAT tests before merging.

rbradford:
Works well for me in my SingleVM.

markdryan:
The occasional failures have been tracked down to #763.  With the fix #764 applied the tests pass consistently on my single vm instance. Just need to check on Jenkins now which I will do tomorrow.

markdryan:
This PR is ready to be merged now.

_release/bat/README.md
```diff
@@ -0,0 +1,71 @@
...
+can be run:
+
+* "CIAO_IDENTITY" - the URL and port number of your identity service
+* "CIAO_CONTROLLER" - the URL and port number of the ciao compute service
+* "CIAO_USERNAME" - a test user with user level access to a test tenant
+* "CIAO_PASSWORD" - your test user's password
+* "CIAO_ADMIN_USERNAME" - your cluster admin user name
+* "CIAO_ADMIN_PASSWORD" - your cluster admin password.
+
+## Running all the BAT tests
```
mcastelino: @markdryan do we need to do setup.sh when testing on single VM? Maybe good to add here.

mcastelino:
Verified on FC24 with Single VM.

PR #761, opened 2016-10-31
Ephemeral boot from volume

rbradford:
This is some early enabling work for the "Images as Volumes" project. With this change volumes created from an image specified in a storage workload will be deleted when the instance itself is deleted.

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -61,6 +61,27 @@ func (client *ssntpClient) CommandNotify(command ssntp.Command, frame *ssntp.Fra
 	glog.V(1).Info(string(payload))
 }
 
+func (client *ssntpClient) deleteEphemeralStorage(instanceID string) {
+	attachments, _ := client.ctl.ds.GetStorageAttachments(instanceID)
+	for _, attachment := range attachments {
+		if !attachment.Ephemeral {
+			continue
+		}
+		err := client.ctl.ds.DeleteStorageAttachment(attachment.ID)
+		if err != nil {
+			glog.Warningf("Error deleting attachment from datastore: %v", err)
+		}
+		err = client.ctl.ds.DeleteBlockDevice(attachment.BlockID)
+		if err != nil {
+			glog.Warningf("Error deleting block device from datastore: %v", err)
+		}
+		err = client.ctl.DeleteBlockDevice(attachment.BlockID)
+		if err != nil {
+			glog.Warningf("Error deleting block device: %v", err)
+		}
+	}
+}
+
 func (client *ssntpClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
 	payload := frame.Payload
 
@@ -73,6 +94,7 @@ func (client *ssntpClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
 			glog.Warning("Error unmarshalling InstanceDeleted")
 			return
 		}
+		client.deleteEphemeralStorage(event.InstanceDeleted.InstanceUUID)
 		client.ctl.ds.DeleteInstance(event.InstanceDeleted.InstanceUUID)
 	case ssntp.ConcentratorInstanceAdded:
 		var event payloads.EventConcentratorInstanceAdded
--- a/ciao-controller/internal/datastore/sqlite3db.go
+++ b/ciao-controller/internal/datastore/sqlite3db.go
@@ -219,6 +219,7 @@ func (d attachments) Init() error {
 		id string primary key,
 		instance_id string,
 		block_id string,
+		ephemeral int,
 		foreign key(instance_id) references instances(id),
 		foreign key(block_id) references block_data(id)
 		);`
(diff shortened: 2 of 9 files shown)


ciao-controller/client.go
```diff
@@ -73,6 +73,21 @@ func (client *ssntpClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
...
+					glog.Warningf("Error deleting block device from datastore: %v", err)
+				}
+				err = client.ctl.DeleteBlockDevice(attachment.BlockID)
+				if err != nil {
+					glog.Warningf("Error deleting block device: %v", err)
+				}
+			}
+		}
+
 		client.ctl.ds.DeleteInstance(event.InstanceDeleted.InstanceUUID)
```
kaccardi: I would have put the code to delete the block device into the where we update the attachments in datastore.go. Otherwise, you'll get to this section:
// update the state of the volume.
                        bd.State = types.Available
                        err = ds.UpdateBlockDevice(bd)
                        if err != nil {
                                glog.Warning(err)
                        }

And if the block device was deleted, you'll have issues.

ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -2305,7 +2306,14 @@ func (ds *sqliteDB) deleteBlockData(ID string) error {
...
-	err := ds.create("attachments", a.ID, a.InstanceID, a.BlockID)
+	var ephemeral int
+
+	if a.Ephemeral {
+		ephemeral = 1
+	} else {
+		ephemeral = 0
+	}
+	err := ds.create("attachments", a.ID, a.InstanceID, a.BlockID, ephemeral)
 	ds.dbLock.Unlock()
```
kaccardi: I don't think you need to convert your bool to an integer for sqlite3 - I believe it translates that for you (from a bool to an int).

ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -2328,8 +2337,12 @@ func (ds *sqliteDB) getAllStorageAttachments() (map[string]types.StorageAttachme
...
 	for rows.Next() {
 		var a types.StorageAttachment
+		var ephemeral int
 
-		err = rows.Scan(&a.ID, &a.InstanceID, &a.BlockID)
+		err = rows.Scan(&a.ID, &a.InstanceID, &a.BlockID, &ephemeral)
+		if ephemeral == 1 {
+			a.Ephemeral = true
+		}
 		if err != nil {
```
kaccardi: you can scan into a bool - sqlite does the type conversion for you.

ciao-controller/internal/datastore/sqlite3db.go
rbradford: Alas not, you get an error about not finding a column called false.

ciao-controller/internal/datastore/sqlite3db.go
rbradford: This one is fixed now.

ciao-controller/client.go
rbradford: Sorted by removing the attachment added which then makes ds.deleteInstance happy.

rbradford:
Unit tests added as best as I could. Some parts of the flow aren't easily testable via unit tests but will be easy to add BAT tests for.

rbradford:
HI @kaccardi i've pushed a bunch of updates to cover what we covered on IRC. Could you re-review?

ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -2304,9 +2307,26 @@ func (ds *sqliteDB) deleteBlockData(ID string) error {
 }
 
 func (ds *sqliteDB) createStorageAttachment(a types.StorageAttachment) error {
-	ds.dbLock.Lock()
-	err := ds.create("attachments", a.ID, a.InstanceID, a.BlockID)
-	ds.dbLock.Unlock()
+	datastore := ds.getTableDB("attachments")
+
+	ds.tdbLock.Lock()
+
```
kaccardi: For the "attachments" table, you need to use dbLock, not tdbLock. It would have been nice if I'd designed "getTableDB" to also return the lock you need to use, but alas.

kaccardi: everything else looked good as far as I could tell.

kaccardi: I filed an issue for the badly designed function.
https://github.com/01org/ciao/issues/775

rbradford: Hi @kaccardi - updated.

ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -2304,8 +2307,25 @@ func (ds *sqliteDB) deleteBlockData(ID string) error {
...
 func (ds *sqliteDB) createStorageAttachment(a types.StorageAttachment) error {
+	datastore := ds.getTableDB("attachments")
+
 	ds.dbLock.Lock()
-	err := ds.create("attachments", a.ID, a.InstanceID, a.BlockID)
+
+	tx, err := datastore.Begin()
+	if err != nil {
+		ds.tdbLock.Unlock()
+		return err
```
kaccardi: unlocking the wrong lock here.

PR #764, opened 2016-11-01
ciao-launcher: Fix docker network locks

### LINKED ISSUE #763: Deleting a VM can disabling containers for an entire subnet
1. Create a container
2. Create a VM
3. Delete the container
4. Delete the VM
5. Create a new container

If all the containers and VMs are created on the same subnet the final attempt to create a container will fail.   The reason is that the launcher code still thinks the subnet exists and hence the docker network exists and will not try to create a new one.

markdryan:
This commit fixes #763, an issue that was introduced by some poor optimization.
When we create a container on a subnet on a CN for the first time we need
to create a new docker network.  This network needs to be created after the vnic
but before the container is created.  It follows that we need to be careful when
creating multiple containers in parallel.  The docker network will be created
during the creation of one of these containers.  The instance go routines
controlling the creation of the other containers need to block until the network
has been created, otherwise they will fail.  The optimisation deleted by this PR
was trying to reduce the amount of time the lock would be held for.
Unfortunately, the code did not consider the case that when instances are
deleted, the final instance deleted for a subnet on a CN might be a VM and not a
container.  When this happens, it becomes impossible to start a new container
for that subnet as the container code thinks that the network for that subnet
already exists, when in fact the underlying bridge has been deleted.

This commit fixes the problem by removing the optimisation and placing Vnic
and docker network creation inside a critical section.  As we do not currently
allow parallel vnic creation, this shoudln't actually slow anything down.

Fixes #763

### DIFF
--- a/ciao-launcher/docker_network.go
+++ b/ciao-launcher/docker_network.go
@@ -32,41 +32,13 @@ type dockerNetworkState struct {
 	err  error
 }
 
-var dockerNetworkMap struct {
-	sync.Mutex
-	networks map[string]*dockerNetworkState
-}
-
-func init() {
-	dockerNetworkMap.networks = make(map[string]*dockerNetworkState)
-}
+var dockerNetworkLock sync.Mutex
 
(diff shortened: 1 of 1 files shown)


mcastelino:
LGTM. Will test and merge

PR #765, opened 2016-11-03
bat: Mention Single VM in the README

markdryan:
Update the README.md file to mention that the BAT tests can be run
with Single VM and that if you are using Single VM you can initialise
all the necessary environment variables by sourcing ~/local/demo.sh

### DIFF
--- a/_release/bat/README.md
+++ b/_release/bat/README.md
@@ -18,6 +18,11 @@ existing test case runner, test-cases.
 
 ## Set up
 
+The BAT tests require a running ciao cluster to execute.  This can be a
+full ciao cluster running on hundreds of nodes or a Single VM ciao cluster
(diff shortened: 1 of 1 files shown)


markdryan:
@mcastelino Does this address your comments on the BAT PR?  I also updated the Single VM wiki.

mcastelino:
@markdryan Perfect. I wonder if we need to keep verify.sh around anymore. The only thing it does beyond BAT is to check actual network connectivity to the instances. Maybe we add that to BAT and remove verify.sh?

markdryan:
@mcastelino Yes, that was my plan.  Create a _relase/bat/network package which contains the verify.sh tests.  I'll enter a ticket for this.

PR #766, opened 2016-11-03
ansible playbooks

albertomurillo:
This PR moves the ansible playbooks from github.com/clearlinux/clear-config-management to
ciao repository


markdryan:
@albertomurillo Thanks.  I think this will make deployment a lot easier and will make it less likely that we will break our deployment scripts when we add a new feature to ciao.  I have some review comments though:

1. You're adding multiple copies of the Apache 2.0 license to the repo. These are not needed as there is a copy in the top level of the ciao directory.
2. Some of the commits don't have a description.

I'd also like to see some comments from the other maintainers.  @tpepper , @kaccardi , @mcastelino are you happy for this to be merged.

_DeploymentAndDistroPackaging/ansible/roles/keystone/library/keystone
```diff
@@ -0,0 +1,1424 @@
+#!/usr/bin/python
```
markdryan: @tpepper , @kaccardi , @mcastelino , @rbradford are akk okay with this?

albertomurillo:
1.- Multiple Copies of apache license: There is one copy per role since it was expected to have the roles in ansible galaxy and each role has its own repository. Since this is not the case anymore I will remove roles/*/LICENSE files.

2.- some file have copyright header and others dont: The yaml file with copyright header are those from the roles (the ones the user shouldn't modify). The yaml files without copyright header are those from the playbook which is an example on how the user can call the roles. Should the later contain the copyright header too?

3.- Some of the commits don't have a description: Only d8be7c7e10d8caa00b200d0fec379869208d465f has no description because I thought the title was enough. I will modify the commit message to include a short description too.

markdryan:
@albertomurillo I think we should be consistent with the copyright headers.  Either all the .yml files should have them or none should have them.  Since they can contain code, I suggest we add the headers.

albertomurillo:
Copyright headers are now included in all yml files
https://github.com/01org/ciao/pull/766/commits/26f94ee0a2d4ea838f7a17ef653176be8bb00da7

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/files/tables/resources.csv
```diff
@@ -0,0 +1,5 @@
+1, instances
```
rbradford: Please use the .csv files already in the tree rather than a copy.

markdryan: Agree with this.  One of the nice things about having our ansible scripts in the ciao repo is that we can avoid this sort of duplication.  We should take advantage of this.

markdryan:
I've tested this PR this morning and I managed to install ciao on a cluster successfully.  I've found a few small issues but I'll enter them as separate bugs that can be fixed later.

1. The .gitignore needs to be updated to cover the files that ansible creates in the ciao directory.
2. The installation of the ceph configuration should not fail if the ceph files are not present.

PR #767, opened 2016-11-04
[WIP] Experiment with volume images

rbradford:
With this code the first time an image is used it will get uploaded to a volume. Subsequent will runs use the volume instead. The volumes are COWed before booting via rbd clone.

```
rob@ciao-singlevm:~/go/src/github.com/01org/ciao$ time ciao-cli instance add -workload 79034317-3beb-447e-987d-4e310a8cf410
Created new instance: b8461b07-1a99-42b3-b279-b8bf9952e713

real    1m35.642s
user    0m0.036s
sys     0m0.016s

rob@ciao-singlevm:~/go/src/github.com/01org/ciao$ time ciao-cli instance add -workload 79034317-3beb-447e-987d-4e310a8cf410
Created new instance: dd4f5f2a-7497-418d-a19e-5c75187573e3

real    0m0.355s
user    0m0.036s
sys     0m0.048s

rob@ciao-singlevm:~/go/src/github.com/01org/ciao$ time ciao-cli instance add -workload 79034317-3beb-447e-987d-4e310a8cf410
Created new instance: fc01d08b-8f8e-4f80-a4e2-e281c5b86cae

real    0m0.321s
user    0m0.052s
sys     0m0.032s

rob@ciao-singlevm:~/go/src/github.com/01org/ciao$ ciao-cli volume list
Volume #1
        Name             [Volume for image: 73a86d7e-93c0-480e-9c41-ab42f69b7799]
        Size             [0 GB]
        UUID             [73a86d7e-93c0-480e-9c41-ab42f69b7799]
        Status           []
        Description      []

Volume #2
        Name             []
        Size             [0 GB]
        UUID             [37efc42b-8f71-4d14-aeac-69eacb1bb3cb]
        Status           [in-use]
        Description      []

Volume #3
        Name             []
        Size             [0 GB]
        UUID             [e9b6884f-980b-4a81-b517-7d01c561a715]
        Status           [in-use]
        Description      []

Volume #4
        Name             []
        Size             [0 GB]
        UUID             [0885b1aa-03c1-4542-9fd1-85c6b1fc5f08]
        Status           [in-use]
        Description      []

```

Obviously needs quite a lot of tweaking and unit tests.

### DIFF
--- a/ciao-storage/block.go
+++ b/ciao-storage/block.go
@@ -25,7 +25,9 @@ var (
 
 // BlockDriver is the interface that all block drivers must implement.
 type BlockDriver interface {
-	CreateBlockDevice(image *string, sizeGB int) (BlockDevice, error)
+	CreateBlockDevice(volumeUUID string, image *string, sizeGB int) (BlockDevice, error)
+	CreateBlockDeviceFromSnapshot(volumeUUID string, snapshotID string) (BlockDevice, error)
+	CreateBlockDeviceSnapshot(volumeUUID string, snapshotID string) error
 	DeleteBlockDevice(string) error
 	MapVolumeToNode(volumeUUID string) (string, error)
 	UnmapVolumeFromNode(volumeUUID string) error
--- a/ciao-controller/controller_test.go
+++ b/ciao-controller/controller_test.go
@@ -497,7 +497,7 @@ func TestDetachVolume(t *testing.T) {
 }
 
 func addTestBlockDevice(t *testing.T, tenantID string) types.BlockData {
-	bd, err := ctl.CreateBlockDevice(nil, 0)
+	bd, err := ctl.CreateBlockDevice("", nil, 0)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/ciao-controller/instance.go
+++ b/ciao-controller/instance.go
@@ -23,7 +23,9 @@ import (
 	"net"
 	"time"
 
+	"github.com/01org/ciao/ciao-controller/internal/datastore"
 	"github.com/01org/ciao/ciao-controller/types"
+	"github.com/01org/ciao/ciao-storage"
 	"github.com/01org/ciao/payloads"
 	"github.com/01org/ciao/ssntp/uuid"
 	"github.com/golang/glog"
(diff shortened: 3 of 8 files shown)


rbradford:
@kaccardi & @tpepper FYI, would be nice to see some architectural feedback.

ciao-storage/block.go
```diff
@@ -25,7 +25,9 @@ var (
 
 // BlockDriver is the interface that all block drivers must implement.
 type BlockDriver interface {
-	CreateBlockDevice(image *string, sizeGB int) (BlockDevice, error)
+	CreateBlockDevice(string, *string, int) (BlockDevice, error)
```
tpepper: It's useful self-documenting style to keep the human readable variable name, in my opinion.

rbradford: This code has a mix of with and without. I've put it back.

rbradford:
> 1) Could we start from a simpler pure "boot from volume" where the image is uploaded and no CoW is made? Implement straight boot from the volume.

The problem is that you'd only be able to create a single instance from the workload as the rbd image cannot be shared by multiple VMs. I feel that is quite a significant regression!

I think in your second sentence you make the same conclusion:

> On the otherhand, since we don't yet have an easy way to do workload definitions, the 1:1:1 mapping of workload-to-instance-to-bootvolume makes it awkward in a pure boot from volume workflow today, where having an implicit snap always in there means you don't have a worry about many:1:1 mappings of workload-to-instance(s)-to-bootvolume(s).

For now the snapshots are an implementation detail of creating a new volume from an image. I hope that in the future some of this functionality will be reusable for snapshots exposed to the 

> Related to 2, the commit oneliner summary is bugging me. Perhaps "ciao-controller: Boot from image volume with CoW"? Or "ciao-controller: Boot from image CoW"?

Sure sounds good!

tpepper:
{quote}
The problem is that you'd only be able to create a single instance from the workload as the rbd image cannot be shared by multiple VMs. I feel that is quite a significant regression!
{quote}

While I do have the same conclusion about the 1:1ness I'm not viewing it as a regression.  I see it even as a feature.  What makes boot-from-volume distinct from boot-from-image.  Even if an image becomes a volume, then I see boot-from-volume as the path to boot from ar/w rootfs backing store.  Where boot-from-image would be the boot from r/o rootfs backing store (with r/w overlay).  Those are the two most primary use cases for a pet-VM imho: fully custom (ie: 1:1) and shared image.

I suspect this gives the boot-from-volume an IO performance advantage at runtime.  I expected boot-from-image with a frequently used in the cloud, cloud shared backing image should have an advantage in initial boot time if the virtualized/distributed storage system is any good.

rbradford:
> While I do have the same conclusion about the 1:1ness I'm not viewing it as a regression. I see it even as a feature. What makes boot-from-volume distinct from boot-from-image. Even if an image becomes a volume, then I see boot-from-volume as the path to boot from ar/w rootfs backing store. Where boot-from-image would be the boot from r/o rootfs backing store (with r/w overlay). Those are the two most primary use cases for a pet-VM imho: fully custom (ie: 1:1) and shared image.

I think we're in big confusion of terminology here.

To clarify going forth:

1. Boot from volume - the workload specifies an volume to boot from. You can only have a single instance with that volumes as rbd images can only be opened once.

2. Boot from image - the workload specifies an image to be copy-on-writed into a **newly created rbd image**

Historically boot from volume/boot from image meant different things but that's not helpful when we never have images on disk. (New world order.)

rbradford:
@kaccardi pointed out that booting from a volume specified in the workload is already implemented and I don't make any changes to that in this PR.

tpepper:
> booting from a volume specified in the workload is already implemented and I don't make any changes to that in this PR

I don't disagree.

Between these two scenarios:

1) Boot instances of a read-only common/shared workload image, with instance specific changes to the boot image stored in a user data volume
2) Boot instance of a read-write custom workload volume

do you see no need to implement the second, given the first?  They amount to almost the same thing from the user's perspective.  I say almost because there's a slight difference in what would happen in the user tries to download their instance image/volume to export it to another cloud.  In the first case you might want to export just the CoW overlay, or you might want the download to be the base image merged with the overlay.  In the second case you want the "merged" result.

I can go with the implementation in this PR as long as the cli ends up with sufficient knobs to allow downloading base image, overlay, and merged base image + overlay.

Similarly the workload definition implementation and instance start commands are going to vary slightly if you support both of these cases, or use the first case to emulate the second.

rbradford:
Replaced by #811

PR #771, opened 2016-11-04
ciao-launcher: Stop leaking docker networks

### LINKED ISSUE #768: ciao-launcher: Docker logical networks seem to be left behind after all instances are deleted
Once single VM is run 

 ./setup.sh
. ~/local/demo.sh
~/local/verify.sh

No docker networks related to ciao should be left behind, as all instances have been deleted.

docker network ls --filter driver=ciao

Will show ciao setup network left behind.

Either the network library is not reporting back the network status properly or we may have some other issue.

markdryan:
There were two issues here.  Commit 54d8e50 introduced a bug which
lead to a new docker network being created with every container,
which is not what we want.  The intended behaviour is to create a
new docker network for each new subnet.  This creation of extra
networks lead to a leak as they were not getting deleted as they
were not supposed to exist.

There was an existing bug which could result in docker network
leaks when you created a container, a VM, deleted the container
and then the VM in that order.

Both these issues have now been fixed.

Fixes #768


mcastelino:
LGTM. Will test and merge

PR #772, opened 2016-11-04
ciao-networking: simplify cnci image preparation mount

tpepper:
We have a race potentially between losetup and partition existance.  This
probably is what the retry loop managed to kind of work around in the past.
The udevadm settle was in the wrong location previously.  It needs to come
after creation and before use.  It would have coincidentally in the past,
IF you were retrying.

We can simplify by not looping and rather create, wait, use as a simple
series of commands.

If we find subsequent issues that need more logic in this flow, I suggest
we remove the losetup and udevadm settle, and instead do "kpartx -sav
$img" followed by the mount.  Then also on exit do "kpartx -sdv $img".

### DIFF
--- a/networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
+++ b/networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
@@ -64,21 +64,13 @@ fi
 echo -e "\nMounting image: $image"
 tmpdir=$(mktemp -d)
 
-#it can take some time for the device to get created
-retry=0
-loop=''
-until [ $retry -ge 3 ]
-do
-    if [ "$loop" == "" ]; then
-	loop=`sudo losetup -f --show -P $image`
-    fi
-    sudo mount ${loop}p$partition "$tmpdir" && break
-    let retry=retry+1
-    echo "Mount failed, retrying $retry"
-done
+loop=`sudo losetup -f --show -P $image`
+sudo udevadm settle
+sudo mount ${loop}p$partition "$tmpdir"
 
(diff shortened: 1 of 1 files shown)


networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
```diff
@@ -64,18 +64,9 @@ fi
...
-    if [ "$loop" == "" ]; then
-	loop=`sudo losetup -f --show -P $image`
-    fi
-    sudo mount ${loop}p$partition "$tmpdir" && break
-    let retry=retry+1
-    echo "Mount failed, retrying $retry"
-done
+loop=`sudo losetup -f --show -P $image`
+sudo udevadm settle
+sudo mount ${loop}p$partition "$tmpdir" && break
```
mcastelino: "&& break" must be a carry over from the retry loop

networking/ciao-cnci-agent/scripts/generate_cnci_cloud_image.sh
```diff
@@ -64,18 +64,9 @@ fi
...
 
 if [ $retry -ge 3 ]
```
mcastelino: we probably need a better way now to check if the mount actually went thro. Maybe look for some file that needs to be present.

tpepper: I replaced the now broken $retry check with a minimal sanity check.

mcastelino:
LGTM. Verified on FC24 bare metal.

PR #773, opened 2016-11-04
Single VM: Ensure no ciao network artifacts are left behind

mcastelino:
Ensure that no ciao network artifacts are left behind when all instances have been deleted.

Note: Similar logic needs to be in place for storage other artifacts.

This will help catch regressions of the type https://github.com/01org/ciao/issues/768

### DIFF
--- a/testutil/singlevm/verify.sh
+++ b/testutil/singlevm/verify.sh
@@ -126,6 +126,33 @@ else
 	echo "Container connectivity verified"
 fi
 
+#Clear out all prior events
+"$ciao_gobin"/ciao-cli event delete
+
+#Wait for the event count to drop to 0
+retry=0
+ciao_events=0
+
+until [ $retry -ge 6 ]
+do
+	ciao_events=`"$ciao_gobin"/ciao-cli event list | grep "0 Ciao event" | wc -l`
+
+	if [ $ciao_events -eq 1 ]
+	then
+		break
+	fi
+
+	let retry=retry+1
+	sleep 1
+done
+
+if [ $ciao_events -ne 1 ]
+then
+	echo "FATAL ERROR: ciao events not deleted properly"
+	"$ciao_gobin"/ciao-cli event list
+	exit 1
+fi
+
 #Now delete all instances
 "$ciao_gobin"/ciao-cli instance delete --all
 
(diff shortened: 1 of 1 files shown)


testutil/singlevm/verify.sh
```diff
@@ -137,3 +137,37 @@ fi
...
+until [ $retry -ge 6 ]
+do
+	#Verify that there are no ciao related artifacts left behind
+	ciao_networks=`sudo docker network ls --filter driver=ciao -q | wc -l`
+
+	if [ $ciao_networks -eq 0 ]
+	then
+		break
+	fi
+	sleep 1
```
markdryan: I think we need to increase retry here otherwise we'll get an endless loop in the error case, i.e., in the case docker networks were not deleted.

markdryan:
One other issue with these checks is that it prevents us from moving verify.sh into the BAT tests as these checks will only work on SingleVM.  I'm not saying that we shouldn't merge this, but I'm just pointing out that we can no longer make verify.sh a BAT test unless we can detect inside the BAT test that we're running on SingleVM.  Any idea how best to do this?

testutil/singlevm/verify.sh
markdryan: It might be better to wait until the instance deleted event is received, but this isn't currently working, see #776.

mcastelino: @markdryan good catch.. I will fix the end less loop. issue. We can add the deleted instance event once it works.

markdryan: actually it should work now.  A submitted a patch yesterday for this that got merged.

mcastelino: @markdryan I will update this PR and add the event check also. That will make this test much cleaner. I think in that case we can even eliminate the retry as the docker networks should have been cleaned up.

mcastelino:
@markdryan all review comments addressed. Also verify.sh is also re-entrant w.r.t events. Hence the event clearing.

testutil/singlevm/verify.sh
mcastelino: @markdryan PR updated to address the review comments

testutil/singlevm/verify.sh
```diff
@@ -126,6 +126,33 @@ else
...
+#Clear out all prior events
+"$ciao_gobin"/ciao-cli event delete
+
+#Wait for the event count to drop to 0
+retry=0
+ciao_events=0
+
+until [ $retry -ge 6 ]
+do
+	ciao_events=`"$ciao_gobin"/ciao-cli event list | grep "0 Ciao event" | wc -l`
```
markdryan: Note you could do this

ciao-cli event list -f '{{len .}}

but what you have will work just as well.

testutil/singlevm/verify.sh
markdryan: The original problem is still present I think.  I thought the plan was to get rid of this loop.

markdryan:
BTW, I entered a new bug #792 to cover the fact that we can't actually tell what an event is when it arrives and to which instance it pertains.  Knowing this information would make the use of the event log more secure.  I doubt this will get fixed in time for this PR though.  We can update the verify.sh at a later stage.

testutil/singlevm/verify.sh
mcastelino: @markdryan will add another PR to actually peek into the event and verify the data.

mcastelino:
@markdryan fixed the retry issue.

PR #777, opened 2016-11-07
ciao-controller: Store correct tenant ID for instance deleted events

### LINKED ISSUE #776: Deleted Instances events are logged with the wrong tenant id
This issue is easy to reproduce on Single VM

```
# ciao-cli instance add --workload 69e84267-ed01-4738-b15f-b47de06b62e7 --instances 2
# ciao-cli instance delete --all
# ciao-cli event list
0 Ciao event(s):
```
I would expect to see two delete events listed.  Instead none are listed.  The issue seems to be that events in the log are filtered by tenant-id, but the controller stores the instance id in the tenant id field for instance deleted events.  This is evident by looking at the sqlite3 db

```
sqlite> select * from log;
id|tenant_id|type|message|timestamp
1|c526680a-a392-4391-96e4-623bf2cbc71d|info|Deleted Instance c526680a-a392-4391-96e4-623bf2cbc71d|2016-11-07 13:38:10
2|a833ccbe-c77e-4dbe-b566-89e15270ea84|info|Deleted Instance a833ccbe-c77e-4dbe-b566-89e15270ea84|2016-11-07 13:38:10
```

If you compare the tenant-id field to the message you see it's actually storing the instance-id.

markdryan:
The controller was incorrectly storing the instance ID in the tenant ID
field of the log table for instance deleted events.  This meant that it
was not possible for the user to view these events using ciao-cli event
list as events are filtered by tenant ID, and the IDs stored did not
correspond to any tenants.

Fixes #776

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -993,7 +993,7 @@ func (ds *Datastore) DetachVolumeFailure(instanceID string, volumeID string, rea
 	ds.db.logEvent(i.TenantID, string(userError), msg)
 }
 
-func (ds *Datastore) deleteInstance(instanceID string) error {
+func (ds *Datastore) deleteInstance(instanceID string) (string, error) {
 	ds.instanceLastStatLock.Lock()
 	delete(ds.instanceLastStat, instanceID)
 	ds.instanceLastStatLock.Unlock()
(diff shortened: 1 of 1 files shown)


markdryan:
@kaccardi Is this change okay for you?

PR #785, opened 2016-11-08
[DONOTMERGE] Add initial support for ciao external IPs API

kaccardi:
Modify ciao-controller to add a stubbed out api for external IPs and pools. Modify ciao-cli to use the API to allow the user to interact with the pools and external IPs resources.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -25,21 +25,27 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
+	"fmt"
+	"net/http"
 	"os"
 	"strconv"
 	"sync"
 
+	"github.com/01org/ciao/ciao-controller/api"
 	datastore "github.com/01org/ciao/ciao-controller/internal/datastore"
 	image "github.com/01org/ciao/ciao-image/client"
 	storage "github.com/01org/ciao/ciao-storage"
 	"github.com/01org/ciao/openstack/block"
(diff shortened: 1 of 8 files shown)


ciao-controller/main.go
```diff
@@ -61,6 +68,7 @@ var servicePassword = ""
 var volumeAPIPort = block.APIPort
 var computeAPIPort = compute.APIPort
 var imageAPIPort = osimage.APIPort
+var controllerAPIPort = api.Port
```
markdryan: Will this require an update to our fake identify service?

kaccardi: not so far. I've tested with single VM - it does require some modification to our fake identity server, but not for this reason. The modification is required because of the cli changes I made to support privileged vs. non-privileged routes. Our fake identity service sets the token to always be in the project "ilovepuppies", but I made privileged actions require the "admin" project.

kaccardi: it will also require some configuration changes since we'll want this passed with the config from scheduler.

kaccardi:
@markdryan we are so close to being able to submit a real PR for all the external IP support that I think I'd rather wait to merge this until we have everything together. But please let me know if you would rather have it earlier and I'll revisit.

markdryan:
@kaccardi No rush.  I can wait.

mcastelino:
@kaccardi we can close this right?

PR #796, opened 2016-11-09
ansible: Remove .DS_Store

dlespiau:
That file caches some metadata for Finder and shouldn't have been
committed to the git repository.

Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>


tpepper:
Looks good to me. Known CI race caused travis to mark this as failed, unrelated to the patch in question here.  I've triggered re-running the test, but this is good to merge.

albertomurillo:
+1

PR #778, opened 2016-11-07
A more generic ciao vendor

dlespiau:
Sitting next to Mark, I've been looking at ciao-vendor for the cc-oci-runtime vendoring, but did things by hand until now: I have a hand written packages.json and manually culled deps in vendor/.

It's time to bite the bullet and make ciao-vendor usable for other projects. With this PR it now works for cc-oci-runtime!

```
$ ciao-vendor packages
Package											Status		Repo				Version		License
github.com/01org/cc-oci-runtime/vendor/github.com/davecgh/go-spew/spew			Vendored	github.com/davecgh/go-spew	6d212800	ISC
github.com/01org/cc-oci-runtime/vendor/github.com/hyperhq/runv/hyperstart/api/json	Vendored	github.com/hyperhq/runv		28d48ac6	Apache v2.0
github.com/01org/cc-oci-runtime/vendor/github.com/pmezard/go-difflib/difflib		Vendored	github.com/pmezard/go-difflib	792786c7	BSD
github.com/01org/cc-oci-runtime/vendor/github.com/stretchr/testify/assert		Vendored	github.com/stretchr/testify	976c720a	MIT
```

PR #798, opened 2016-11-09
contributing: Add an exception for commit message summaries

dlespiau:
I have a couple of trivial commits (eg. fixing a typo in a commit) that
don't really need a summary. Not even sure what I'd say!

    ciao-vendor: Fix depdenency typo

    Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>

And currently that's preventing a PR to be merged.

So, proposing to be a bit more precise in CONTRIBUTING.md.

Signed-off-by: Damien Lespiau <damien.lespiau@intel.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,6 +43,9 @@ For example:
 	Signed-off-by: Samuel Ortiz <sameo@linux.intel.com>
 ```
 
+Under exceptional circumstances, if the patch is fixing a typo in a comment for
+instance, it is acceptable to omit the detailed explanation.
+
 ## Pull requests
 
 We accept github pull requests.


tpepper:
I'm annoyed by exceptions.  What's wrong with
```
component name: fix typo

Change fo to foo.

signed-off-by: a@b.com
```

Then you know you don't run afoul of pedantic filters.

dlespiau:
I'm annoyed by pedantic filters.

Nothing is fundamentally  wrong with:
```
component name: fix typo

Change fo to foo.

signed-off-by: a@b.com
```
 
And, as a reviewer, I will not point out that the subject isn't very descriptive and could be improved for better git shortlog's. I will not point out that the word after the component name should be capitalised. I will not point out either that Signed-off-by starts with a capital S. I will also try very hard to not point out the name and '<' and '>' are missing from the email address.

What I'm arguing is that having unmovable dogmatic rules on quite trivial issues without applying common sense and blocking PRs on such things, especially for new contributors, is detrimental to the project and that there is nothing wrong either with:
```
component name: Fix fo typo

Signed-off-by: A <a@b.com>
```

Not trying to be difficult, I will eventually concede if facing a wall. I'm just trying to do my bit on the code review culture that IMHO is often forgetting what's really important and focuses on trivial issues.

tpepper:
@dlespiau please keep it up.  IMHO our goal with the doc was to avoid having review _avoided_ by people choosing to focus on the less than meaningful.  We have the same goal.

Go vet and the other tools in travis are a HUGE win in that regard because they enforce a uniform coding style and you get feedback and aligned ahead of submitting a PR or soliciting review, assuming you're using travis.  I like git commit hooks, but recently ran into somebody who only did short logs and was trained to write "..." in the full message by a tool that required non-whitespace there.  Nothing's perfect in automation gates.

In the abstract, with respect to commit shortlogs and messages, my goal when writing them is actually less to get past a pedantic reviewer, but to really to enable an archaeologist supporting our code to understand how and why something happened.  A recent example might be the difference between dbLock and tdbLock in ciao-controller.  These don't have comments in the source and looking at the sources it's not immediately clear which protects what and after some analysis even seems that they're used inconsistently.  @rbradford accidentally used one instead of the other in a commit, but in an error path so the code would test good since our error path test coverage is poor.  If his commit got in, he might do a short "fix typo" message and done.  But somebody a year in the future grepping through the short log and then looking at commit message details for relevant bits and then looking at patches and git blame at certain historical points, trying to understand which lock is for which things might not see this commit if they're grepping for tdbLock.  Maybe a series of longer commit messages popping periodically over time with variations on "this is confusing and I got confused and oops here's the typo / confusion fix" points a future maintainer at a few moments of janitorial work to remove the point of confusion.  Ie: the good answer could be to pause and realize we have a support issue and fix it, but that's often a back seat relative to fixing the primary issue at hand.  I've been through many many long term support situations like this dealing with open source code for the past two decades now.  More verbosity usually feels ridiculous at the moment, but at the cost of a few second today I find a steady trail of quality, grep-able, informative how/why cookie crumbs in shortlogs and commit messages saves my future self lots and lots of time in the future.  I value that investment.

tpepper:
@markdryan has suggested we might should modify our text to be more explicitly clear that we prefer to have commit message body's not just shortlogs.

If we go that route, perhaps additional text like the following helps clarify the rationale:

    A typo fix might say just that in the short log, but it's useful to also have a body that includes the from/to text for git log searchability.  This can help spot trends, find similarly broken code, or help a new fixer do the same as a prior fixer.

kaccardi:
for the record, I'm ok with this PR and also want to give you a high five for your broader philosophical point - but since I'm not the only gatekeeper, I will respect the views of the others here and not merge if there are objections.

markdryan:
My preference would be to continue to require bodies for all commit messages, even trivial ones.  I realise this is tedious for contributors and might seem overly pedantic.  However, I'm worried that if we start allowing exceptions we'll end up revisiting this issue over and over again as we debate whether each new trivial patch justifies an exception.

I therefore suggest that we update the current CONTRIBUTING.md file to explicitly mandate that bodies are supplied for each commit.  The current document is a little confusing, particularly as the blog post it links to describing how to write good commits, says that it's okay for trivial commits to exclude bodies.

dlespiau:
@kaccardi I'm glad the underlying message was not lost! :)

For the record, I'd like to provide more context:

- I submitted a PR (https://github.com/01org/ciao/pull/778) with somewhat fiddly details to take into account. Mark had 2 good comments (offline) that I addressed.
- That PR was still blocked because, in a commit message fixing a typo in a comment, I hadn't put a "detailed summary" of the change. I'd argue it's not necessary to block a PR on such a thing and it seems the link provided in CONTRIBUTING.md even agrees with me :)
- I did amend the PR as the goal here isn't to block it on trivial considerations. It's been merged since.
- The larger point though is to combat the tendency in software projects to follow arbitrary rules and focus  on unnecessary details instead of real problems. It creates hostile environments and new contributors can be driven away. This case is a bit contrived but I have no doubt people have experienced that in the past.
- To compare and contrast, I've contributed a few patches to gnome-builder (https://bugzilla.gnome.org/show_bug.cgi?id=756591). Coming from a kernel background, this was so refreshing I had to congratulate Christian for making new contributors welcome.
    - There were typos in the commit message, not even a mention of it! (I know, crazy)
    - Additional work was outlined, but the commit wasn't blocked on this new work.
    - "I like it." always make people feel good

Don't get me wrong, Ciao is already much better than the kernel!

I can't be the one adding more "unnecessary" rules to a project, so feel free to amend CONTRIBUTING.md with clearer expectations and close this PR if that's the path chosen.

kaccardi:
Thanks @dlespiau - and please feel free to continue to suggest improvements the ciao process. I submitted an issue to track progress of the contributing update. #821

PR #803, opened 2016-11-09
ansible: Retry Create custom projects task

### LINKED ISSUE #788: Ansible fails creating keystone projects in first run (ClearLinux)
This happens only when running the keystone container in clear-containers.

Before task "Create custom services"
Another task is run to ensure port 5000 is up

```
  - name: Wait for keystone to be ready
    wait_for: host="{{ keystone_fqdn }}" port=5000 state=started
```

An issue to clearcontainers has been opened at
https://github.com/01org/cc-oci-runtime/issues/349

In the meantime i will work on a workaround

albertomurillo:
The keystone role has a task to wait for port 5000 to be open before
trying to send request to create services, projects, etc. But there
is an issue with clearcontainers [1] where the port is reported to be
open even if the application has not been started inside the container.

This change adds a workaround to retry the Create custom project task
to avoid the scripts from failing when running in clearlinux with
clearcontainers.

[1] https://github.com/01org/cc-oci-runtime/issues/349

Fixes #788

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/roles/keystone/tasks/access.yml
+++ b/_DeploymentAndDistroPackaging/ansible/roles/keystone/tasks/access.yml
@@ -41,6 +41,14 @@
       login_project_name: "admin"
       insecure: yes
     with_items: "{{ keystone_projects | default([]) }}"
+    # TODO: Remove the following workaround
+    # There is an issue with clearcontainers where it says the port is open
+    # even when it is not ready yet.
+    # https://github.com/01org/cc-oci-runtime/issues/349
+    register: result
+    until: result.failed == false or result.failed is undefined
+    retries: 10
+    delay: 1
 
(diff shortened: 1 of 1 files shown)


kaccardi:
@albertomurillo you say this fixes 788 - but then the comment implies it's a workaround. I'm not sure if we should mark it "fixes" if it doesn't really fix the issue unless you are planning on opening a separate one.

albertomurillo:
Issue #788 is a symptom of a bug in clear containers: https://github.com/01org/cc-oci-runtime/issues/349

This PR contains a workaround to deal with the bug in clearcontainers and should be removed once clearcontainers's bug is fixed.

albertomurillo:
Also if the bug gets fixed in clearcontainers, the workaround (which is only a retry) will not affect the functionallity

Issue #614, opened 2016-09-29
Single VM Dev Env: Use keystone container for identity services


mcastelino:
Currently the single VM tests use a fake identity service. As we now have a dockerized keystone instance we should provision and use this keystone in single VM for identity services.

markdryan:
I'll take this one.  I need to do this to fully test ceph inside the Single VM.  Right now ciao-cli can't find the volume service.  It's probably just as easy to enable keystone in Single VM as it is to update the testutils/identity.go file.

PR #804, opened 2016-11-09
singlevm: Add keystone support to setup.sh

bwarden:
Enable using keystone identity service for singlevm test case instead of
ciao-controller's internal service.

Details:
* Clean up ports assigned to different modules.
* Set up environment for keystone.
* Generate configuration.yaml programmatically from setup.sh so we can
populate it with system-specific values.
* Re-use ciao certs for keystone to ensure a common CA.
* If system uses /usr/sbin/update-ca-certificates, copy the controller
cert to /usr/local/share/ca-certificates/ciao.crt and run it, adding the
controller's certificate as a trusted CA for the system. This allows
ciao-image to run.
* Start keystone container from setup.sh. Clean it up in cleanup.sh.
* Ensure that keystone starts before launching ceph container so that
keystone can listen on port 5000.
* Remove --single from ciao-controller command line to disable internal
identity server, define keystone identity URL explicitly in environment.
* After launching container, wait until keystone service starts
responding to queries.
* Create/populate expected entries in keystone via openstack util:
  - projects
  - services
  - roles
  - users
  - entities

Signed-off-by: Brett T. Warden <brett.t.warden@intel.com>


bwarden:
#614

testutil/singlevm/setup.sh
```diff
@@ -79,6 +82,33 @@ done
...
+  controller:
+    compute_ca: /etc/pki/ciao/controller_cert.pem
+    compute_cert: /etc/pki/ciao/controller_key.pem
+    identity_user: csr
+    identity_password: hello
+  image_service:
+    type: glance
+    url: https://${ciao_host}
+  launcher:
+    compute_net: [198.51.100.1/24]
```
mcastelino: @bwarden we should be able to use ${ciao_subnet} as it is set to ciao_subnet=198.51.100.1/24

bwarden: Will replace in compute_net and mgmt_net.

testutil/singlevm/setup.sh
```diff
@@ -274,7 +352,92 @@ else
...
+sudo docker run -d -it --name keystone \
+    -p $keystone_int_port:5000 \
+    -p $keystone_ext_port:35357 \
+    -e IDENTITY_HOST="$ciao_host" -e KEYSTONE_ADMIN_PASSWORD="${OS_PASSWORD}" \
+    -v $MYSQL_DATA_DIR:/var/lib/mysql \
+    -v ${OS_CACERT}:/etc/nginx/ssl/keystone_cert.pem \
+    -v ${ciao_pki_path}/${ciao_image_key}:/etc/nginx/ssl/keystone_key.pem clearlinux/keystone
+
+echo -n "Waiting for keystone identity service to become available"
+try_until=$(($(date +%s) + 30))
```
obedmr: I'd suggest a shorter solution, it may require a little change to make it fancier :).

```
until /bin/bash -c "source openrc ; openstack user list"
do
    echo Waiting keystone ...
done
```

markdryan:
This is not working for me.  The script fails when trying to upload an image:

```
Uploading test images to image service
---------------------------------------------------------------------------------------
F1110 10:10:43.668554   12820 image.go:157] ciao-cli FATAL: Could not create image [Post https://singlevm:9292/v2/images: dial tcp 10.0.2.15:9292: getsockopt: connection refused]
goroutine 1 [running]:
```

The reason is that both ciao-image and ciao-controller fail to start.  From the controller logs.

```
F1110 10:10:37.244323   12412 main.go:191] Unable to authenticate to Keystone: Post https://singlevm:35357/v3/auth/tokens: EOF
goroutine 1 [running]:
```

markdryan:
Would it be possible to separate all the keystone stuff out into a separate file?

testutil/singlevm/setup.sh
```diff
@@ -142,22 +208,34 @@ then
...
-sudo cp -f controller_cert.pem /etc/pki/ciao
-sudo ln -s /etc/pki/ciao/controller_key.pem /etc/pki/ciao/ciao-image-key.pem
-sudo ln -s /etc/pki/ciao/controller_cert.pem /etc/pki/ciao/ciao-image-cacert.pem
-
+sudo cp -f ${ciao_controller_key} ${ciao_pki_path}
+sudo cp -f ${ciao_controller_cert} ${ciao_pki_path}
+sudo ln -s ${ciao_pki_path}/${ciao_controller_key} ${ciao_pki_path}/${ciao_image_key}
+sudo ln -s ${ciao_pki_path}/${ciao_controller_cert} ${ciao_pki_path}/${ciao_image_cert}
+#Update system's trusted certificates
+CACERT_PROG=$(type -p update-ca-certificates)
```
markdryan: What's the motivation for installing these temporary certs in the system cert store rather than just using CIAO_CA_CERT_FILE as we've done in the past? For people who run setup.sh on bare metal it's probably best if it makes as few changes to the underlying OS as possible.

Right now we don't delete these certs when we do a cleanup.sh so this means that these temp certs are in the system store for ever.

testutil/singlevm/setup.sh
```diff
@@ -79,6 +118,33 @@ done
...
+  launcher:
+    compute_net: [${ciao_subnet}]
+    mgmt_net: [${ciao_subnet}]
+    disk_limit: false
+    mem_limit: false
+  identity_service:
+    type: keystone
+    url: https://${ciao_host}:${keystone_ext_port}
+EOF
+) > $conf_file
```
markdryan: Should this file be stored under ~/local along with everything else?  Right now it's being created in my ciao source folder.

tpepper:
{quote}
using CIAO_CA_CERT_FILE
{quote}

Yes that will save complexity dramatically.  Adding to the system trust store varies with linux flavor.  And no cleanup required if just in the env var.

testutil/singlevm/setup.sh
```diff
@@ -14,12 +19,46 @@ ciao_src="$GOPATH"/src/github.com/01org/ciao
...
+export OS_IDENTITY_API_VERSION=3
+export OS_PASSWORD=${ciao_admin_password}
+export OS_AUTH_URL=https://"$ciao_host":$keystone_ext_port/v3
+export OS_USERNAME=${ciao_admin_username}
+export OS_KEY=
+export OS_CACERT=${ciao_pki_path}/${ciao_image_cert}
+export OS_PROJECT_DOMAIN_NAME=default
+
+# Save these vars for later use, too
+set | grep ^OS_ | while read VAR; do
```
mcastelino: @bwarden this is nice. we can do the same for ^CIAO_* also.

testutil/singlevm/setup.sh
markdryan: I see in the commit comments that this was done to get ciao-image to start.   I wonder if there's a way to get ciao-image to work without using the system certs.  It currently starts okay using the fake identity service using just CIAO_CA_CERT_FILE.  What problems were you seeing?

mcastelino:
@bwarden @kaccardi with this change we should be able to test tenant API's also in verify.sh
We should do that in a follow on PR, we should enhance this work to test with at least 2 tenants.

mcastelino:
@jorgevgut can you test this PR with your webui work please. We should now be able to use webui with singlevm.

bwarden:
Reworking... will first submit a PR for configuration changes, then another for keystone.

bwarden:
Replaced with https://github.com/01org/ciao/pull/813

PR #817, opened 2016-11-11
ansible: Optionally skip ceph configuration on ciao nodes

### LINKED ISSUE #782: Ansible playbook should not fail if the ceph directory is missing
The ansible scripts currently fail if there is no ceph directory under _DeploymentAndDistroPackaging/ansible.  This is reasonable for the first installation of ciao on a cluster but is not very convenient when reinstalling ciao, during development for example.  In such situations ceph is already successfully configured and does not need to be reconfigured.  It would be better if the ansible scripts simply skipped the ceph installation if the ceph directory is missing.  Perhaps they could perform some simple test on the cluster to make sure ceph is acutally working.

albertomurillo:
Ciao requires a ceph cluster to be available for all ciao nodes.

The deployment playbook requires that the systadmin provides ceph
configuration and secret files in order to configure ceph on all
ciao nodes.

This changes introduces `skip_ceph` variable which allows the deployment
to skip ceph configuration and leave that step to be performed manually
by the sysadmin.

Fixes #782

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/doc/development.md
+++ b/_DeploymentAndDistroPackaging/ansible/doc/development.md
@@ -1,8 +1,16 @@
 # Development Mode
-This playbooks installs ciao from the latest ciao packaged for your distro.
+By default, these playbooks install ciao on production environments from
+the latest packages available for each supported distro.
 
-If you want to try the latest features, there is a way to use this playbook to
-deploy ciao from github.com/01org/ciao master branch source code.
+However, there are some options that can be specified when running the playbooks
+to alter this default behaviour. These options are useful during development and
+are described in more detail below:
 
-Set `ciao_dev = True` in [group_vars/all](../group_vars/all) file or pass the
+## Deploy from source code
+In order to deploy ciao from the master branch in github set `ciao_dev = True` in [group_vars/all](../group_vars/all) file or pass the
 command line argument `--extra-vars "ciao_dev=true"`
+
+## Skip ceph configuration
+If you plan to manually setup ceph in your ciao nodes set `skip_ceph = True`
+in [group_vars/all](../group_vars/all) file or pass the command line argument
+`--extra-vars "skip_ceph=true"`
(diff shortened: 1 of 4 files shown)


_DeploymentAndDistroPackaging/ansible/doc/development.md
```diff
@@ -1,8 +1,15 @@
 # Development Mode
-This playbooks installs ciao from the latest ciao packaged for your distro.
+This playbooks are though to install ciao on production environments from the
```
markdryan: This sentence doesn't make sense.  How about

'By default, these playbooks install ciao on ..'

_DeploymentAndDistroPackaging/ansible/doc/development.md
```diff
@@ -1,8 +1,15 @@
...
+latest packages availabe for each supported distro.
 
-If you want to try the latest features, there is a way to use this playbook to
-deploy ciao from github.com/01org/ciao master branch source code.
+above you can finde some tips that can be useful if you plan to try new features
```
markdryan: This sentence is a bit confusing as well.  How about

'However, there are some options that can be specified when running the playbooks to alter this default behaviour.  These options are useful during development and are described in more detail below.'

albertomurillo:
@markdryan Text was updated, thanks for the suggestions.

_DeploymentAndDistroPackaging/ansible/doc/development.md
```diff
@@ -1,8 +1,16 @@
 # Development Mode
-This playbooks installs ciao from the latest ciao packaged for your distro.
+By default, these playbooks install ciao on ciao on production environments from
```
markdryan: here we have a repition of ciao on, e.g., 'ciao on ciao on'.  Is this intentional?

'install ciao in production environments' would be better I think

albertomurillo: definetly a mistake again

_DeploymentAndDistroPackaging/ansible/roles/ciao-common/README.md
```diff
@@ -10,6 +10,7 @@ The following variables are available for all ciao roles
 Variable  | Default Value | Description
 --------  | ------------- | -----------
 cephx_user | admin | cephx user to login into the ceph cluster
+skip_ceph_ceph | False | When set to true, ansible will not configure ceph on Ciao nodes
```
rbradford: What's skip_ceph_ceph?

albertomurillo: another typo :)

PR #824, opened 2016-11-14
Make Ciao images immutable

rbradford:


kaccardi:
@rbradford does the image service upload api need to be modified as well to enforce immutability? And should the ceph driver be modified to make images read only?

rbradford:
@kaccardi Not sure if the image upload API can be made to enforce the immutability. Will check.

The ceph driver already makes rbd images used for images protected which will ensure that they cannot be modified. (They can be deleted if the reference count on them hits zero.)

rbradford:
@kaccardi I've updated the patch to deny an override in the posix backend. The new ceph backend already does this.

ciao-image/datastore/posix.go
```diff
@@ -29,6 +30,9 @@ type Posix struct {
 // If the image already exists it will be overridden.
 func (p *Posix) Write(ID string, body io.Reader) (int64, error) {
```
kaccardi: can you change the comment?

rbradford:
@kaccardi fixed.

PR #827, opened 2016-11-15
[RFC]testutil: Add snake oil cert files for roles

bryteise:
The current test utility support passing role based snake oil
certificates for testing but only only as strings. Some of the APIs that
need to be tested require these to be passed as file paths instead so
this patch adds that capability.

The fake ca certs were generated with the following process using
ciao-cert from commit 3bb33ad64383f0712d0f68e7e25379a7be636254:

$ ./ciao-cert -directory ../testutil -host localhost -server -role \
scheduler -ip 127.0.0.1

$ ./ciao-cert -directory ../testutil -host localhost -role $role \
-server-cert ../testutil/cert-Scheduler-localhost.pem -ip 127.0.0.1

The RoleToTestCertPath function the change implements will give the path
to the CA certificate generated when creating the server certificate and
the path to the certificate matching the requested role. If there is no
certificate path matching the specified role an error will be returned
as well.


bryteise:
@rbradford This really should have been marked RFC. It is part of an effort for adding unit tests to tracing which isn't merged yet.

Anyway the programmatic generation I didn't know about before so I explored using that. It does seem to add a fair amount of extra overhead for each use in verbosity and in run time so I'm not really sure it is a better approach (four tests used to run in ~.4s now run in 2-3s). I pushed a branch here(https://github.com/bryteise/ciao/blob/gen-certs/trace/tracer_test.go#L210) if you want to look at my implementation.

tpepper:
Can you make the certs in a TestMain() once and use them in all the unit tests?

I do like the idea of not having any snakeoil certs checked in, but still having programmed usable certs for coherent unit tests.

If the total decrease across ciao is that 0.4s becomes 3s, I'd take the slowness.

bryteise:
@tpepper I may be not quite understanding your suggestion but I'm understanding it as:
1) All tests needing certs calls a Generate() function that either creates the certs or verifies they exist and then returns
2) Have a gitignore for the snakeoil temporary certs (there isn't an obvious clean up point)

I'm a little worried about stale certs but it isn't anything a git clean can't solve so I'm okay with this approach.

bryteise:
@rbradford I'm actually kind of in favor of checking the certs into the tree because there is much less room for false positive test failures due to certificate generation problems. Could you or @tpepper and in some reasoning here? I do have a patch that does on the fly test certificate generation but was having trouble justifying it based on cluttering the git tree.

markdryan:
BTW launcher currently has some test certs checked into the tree which are used for testing.  I may remove these based on the outcome of this conversation.

I don't think we need to worry too much about people trying to use our test certs on real clusters as they will only work on local host.  The main reason I can think of to not check the certs in is that it clutters up the source tree it's easy to forget to regenerate them when the cert-tool changes, but then the tests would probably fail.

Perhaps we could add a helper function that makes it easier to generate temporary certs, say, certs.CreateTempSet.  Then all you'd need to do would be to

given

```
       type  TempCerts struct {
           CAPath string
           serverPath string
           clientPath string
       }
```

```
	template, err := certs.CreateCertTemplate(ssntp.SCHEDULER, "test", "test@test.test", []string{"localhost"}, []string{"127.0.0.1"})
        paths, cleanupFN, err := certs.CreateTempSet(template, false)
       defer cleanupFN()
```

paths is an instance of TempCerts and cleanupFN is a function that removes all three certs when invoked.

bryteise:
@markdryan You might want to look at https://github.com/bryteise/ciao/commit/aecff89cc443f7fdab2e17961db4ff23793d894f for my alternate approach. The commit describes my reasoning pretty well I think for why I avoided the cleanup function but it could be added. Regardless it generates all the different role certificates in addition to the CA at once.

tpepper:
My desires here are:

1) No duplicate certs checked in per component to make the component tests run.  Options are check in one comprehensive set in testutil, or modify everything to spawn certs on demand when they're needed.  The one-check-in-of-master-snakeoil-certs path would be nicer if it could include a git-hook to build/commit new certs on commits to the ciao-cert module, as @markdryan points out we humans aren't great at maintaining persistent state that matches the code expectations.  But, if we make mandatory changes to the key format, we complicate system update...new binaries gets deployed and the cluster can no longer form because old keys are present.  That may drive us in a direction where practically we don't modify how we create certs, aside from for optional additions, where if those features have unit tests which required pre-build snake-oil keys, the commiter's unit tests would likely fail until they also fixed the keys.  So practically it might not be a huge problem.

2) No persistent snake-oil.  Even though the keys are "localhost" only, localhost works on all hosts, so you can make a cluster with these.  But I guess you'd need to forward localhost ssntp traffic to your real (remote, not localhost) ssntp server.  So yeah @markdryan is right that this is probably a minor risk practically.  Still it's something we're liable to have to justify over and over again to a security reviewer who stopped and puked up a red flag upon seeing snake oil checked in.  It also feels kludgy for the same reason a tests-presumed set of keys in /etc/pki/ciao for is kludgy.  It feels cleaner and more complete and self-documenting if any/all component test prerequisites are established in a TestMain() and then cleaned up after m.Run().  I like https://github.com/bryteise/ciao/commit/aecff89cc443f7fdab2e17961db4ff23793d894f though I would prefer a RemoveTestCerts() helper available.  You get reuse across the component at least.

If the above seems like I'm equivocating, apologies.  I could go either way on this.  I like the idea of autogenerate and cleanup, but if it's too invasive for now to rip out the keys I can live with that.

bryteise:
Thanks @tpepper, that's exactly the kind of information I was looking for. Your 2 did not really seem worrying to me since these cert would sit under testutil instead of in a component's directory. It is very understandable to me to want to avoid the existence of the test keys outside of a test run however I just weighted this pretty low.

The ability to have self creating, self cleanup per component unit test is really a win and probably the most compelling reason to avoid checking in certs to me. I think only one set of certs needs to exist for any test though so I'm not sure per component generation has as much value.

I can't really speak to how often the cert layout has changed and how much cert regeneration would be needed. I guess if that is frequent it would be a problem but as you said would be git-hookable to avoid any real problems.

I'm also okay with either approach but as I said I had trouble making a good enough argument for my MakeCerts() change and wanted more feedback. Hopefully we can get to a consensus (I'll go back and add a RemoveTestCerts() and make sure that happens as part of MakeCerts() too).

Issue #820, opened 2016-11-14
ansible: /etc/systemd/system directory belongs to ciao user


albertomurillo:
`ciao-controller` role creates a `ciao-controler.service` sysemd unit under `/etc/sysemd/system` directory.

In ClearLinux, there could be the case that `/etc/systemd/system` directory does not previously exist due the stateless anture of the OS.`ciao-common` role deals with this issue by ensuring `/etc/systemd/system` exsits.

The problem is that the task that ensures the directory exists, also ensures other ciao directory are present and belongs to ciao user. 

https://github.com/01org/ciao/blob/master/_DeploymentAndDistroPackaging/ansible/roles/ciao-common/tasks/main.yml#L25-L35

PR #832, opened 2016-11-15
ciao-deployment: Fix ownership of /etc/systemd/system

erick0z:
Add a separate task in order to create the /etc/systemd/system
directory and make sure it belongs to root.

Fixes https://github.com/01org/ciao/issues/820


albertomurillo:
+1

Just the commit message, previous commits starts with `ansible:` 
not sure if we should keep using ansible or change to `ciao-deployment`

PR #835, opened 2016-11-16
ciao-down: Add ciao-down, a utility for building VMs for Single VM

markdryan:
ciao-down is a small utility for setting up a VM that contains everything
you need to run ciao's Single VM.  All you need to have installed on your
machine is:

1. Go 1.7 or greater
2. An ssh key

Then simply type

go get github.com/01org/ciao/testutil/ciao-down
$GOPATH/bin/ciao-down prepare

ciao-down will install some needed dependencies on your local PC such
as qemu and xorriso.  It will then download an Ubuntu Cloud Image and
create a VM based on this image.  It will boot the VM and install in that
VM everything you need to run ciao Single VM, including docker, ceph,
go, gcc, etc.  When ciao-down prepare has finished you can connect to the
newly created VM with

$GOPATH/bin/ciao-down connect

Your host's GOPATH is mounted inside the VM.  Thus you can edit your
the ciao code on your host machine and test in Single VM.

Follow the instructions at the prompt to start Single VM.

The nice thing about ciao-down is that it is built using existing ciao
packages.  It uses osprepare to install dependencies on the host, some
launcher code to create the images needed by the VMs, and the qemu
package to launch and manage the VMs.  All the rest is done via cloud-init.


markdryan:
Don't merge this yet.  There's still lots of cleanup and testing to do.  Plus the static analysis tools are checking.

Just posting to get some review comments and feedback.

markdryan:
Also, I've only tested this using an Ubuntu 16.04 host.  I haven't tested on Redhat or clear yet.

markdryan:
The GOPATH sharing doesn't seem to be working properly in Ubuntu 14.04.  The user and group ids of the files are incorrect in the SingleVM.  I guess this is due to the very old version of qemu used in this distro.  I'll add an option that allows users to disable the p9 file sharing.

markdryan:
@tpepper This is ready to merge now.  There's some more stuff I want to add like installing all the dependencies for the UI and some warnings if nested virtualization is not supported but what we have here is useful in its current form.

mcastelino:
@markdryan first of all, this is amazing...
Was testing this on FC24. Ran into a few issues

**When creating the VM:**
Downloading latest clear cloud image : [OK]
Setting git user.name : [FAIL]
Setting git user.email : [FAIL]
VM successfully created!



**When running the tests inside the VM:**

When running setup.sh

Uploading test images to image service
---------------------------------------------------------------------------------------
F1202 22:48:04.113230   22004 image.go:188] ciao-cli FATAL: Could not create image [Post https://singlevm:9292/v2/images: dial tcp 127.0.0.1:9292: getsockopt: connection refused]
goroutine 1 [running]:

**When running verify.sh**

F1202 22:49:48.894320   22724 instance.go:129] ciao-cli FATAL: HTTP Error [500] for [POST https://singlevm:8774/v2.1/b1661c2b06c54f338999ee703768389f/servers]: {"error":{"code":500,"name":"Internal Server Error","message":"Error when running: [rbd --id ciao clone 4e16e743-265a-4bf2-9fd1-57ada0b28904@ciao-image 6a583235-dad5-4a0f-94ce-c47018c73950]: exit status 2"}}
null

markdryan:
@mcastelino  Thanks for testing.  

I think the issue you are seeing with setup.sh 

> F1202 22:48:04.113230 22004 image.go:188] ciao-cli FATAL: Could not create image [Post
> https://singlevm:9292/v2/images: dial tcp 127.0.0.1:9292: getsockopt: connection refused]
> goroutine 1 [running]:

is a problem with Single VM rather than ciao-down.   I entered a bug for this yesterday but didn't get around to fixing it.  See #922.  It's basically a race condition in SingleVM.  setup.sh waits 5 seconds after starting controller before it tries to upload any images.  But sometimes, when the VM boots for the first time, this is not enough.  You can easily check this by doing a cleanup.sh and then setup.sh again.   I would expect image upload to work the second time around.

I haven't seen the git issue before but it isn't fatal.  Most of the time you won't be submitting patches from inside the VM as this can be done from the host.  I might ask you to help me debug the issue on Monday.

markdryan:
@mcastelino I've added two patches to the PR:

 - f78ecf4 fixes the image race condition in Single VM, by copying some of @bryteise's code.
-  26b8416 adds a --debug option to ciao-down prepare that allows you to dump out the cloudinit files

Could you try again running

```
./ciao-down delete  # to delete the old instance
./ciao-down prepare --debug
```

This time your setup.sh should succeed without any issues.  You should also see some debug on your terminal related to setting the git user name and email.  It would be interesting to know what those lines are.

mcastelino:
@markdryan My previous issues got fixed. But I know have a new error

```
F1205 17:56:48.537439   21905 instance.go:129] ciao-cli FATAL: HTTP Error [500] for [POST https://singlevm:8774/v2.1/064c9aab33af4edb8dd7a915e58a854c/servers]: {"error":{"code":500,"name":"Internal Server Error","message":"Failed to Launch CNCI for 064c9aab33af4edb8dd7a915e58a854c"}}
null
```

This is because KVM is not running inside the VM
```
E1205 17:56:48.475938   20410 qemu.go:1004] Unable to launch qemu: exit status 1
E1205 17:56:48.475977   20410 qemu.go:1006] Could not access KVM kernel module: No such file or directory
failed to initialize KVM: No such file or directory
```

When I check the qemu command line I see that you have done the the right thing
` -enable-kvm -cpu host`

When I check the CPU capabililites inside the VM I do not see VMX
`flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon rep_good nopl eagerfpu pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand hypervisor lahf_lm abm fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm xsaveopt arat
`

markdryan:
@mcastelino Thanks for the update. @rbradford ran into the same issue.  I can add a check to the start of ciao-down to abort if nested virtualization is not enabled in the host.  Right now we need it.  As soon as the CNCI switches to a non clear base, we can replace the error with a warning.  How does this sound?  I don't think ciao-down should try to mess around with the kernel boot parameters to enable this.  Do you agree?

mcastelino:
@markdryan Verified the on FC24. In FC24 nesting VT-x should be enabled explicitly
It can be done via

Adding `kvm-intel.nested=1` to GRUB_CMDLINE_LINUX
/etc/default/grub

And then generating the grub.cfg using (based on your BIOS, EFI or legacy)
```
grub2-mkconfig -o /boot/efi/EFI/fedora/grub.cfg
or
grub2-mkconfig -o /boot/grub2/grub.cfg
```

I think we should _not_ just document that we need nested vt support in KVM.
We should not automatically modify the users's system setting.

We should check in ciao-down if nested vtx is not enabled and bail out for now

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+// TODO: Code copied from launcher.  Needs to be moved to qemu
+
+func createRootfs(ctx context.Context, backingImage, instanceDir string) error {
+	vmImage := path.Join(instanceDir, "image.qcow2")
+	if _, err := os.Stat(vmImage); err == nil {
+		_ = os.Remove(vmImage)
+	}
+	params := make([]string, 0, 32)
+	params = append(params, "create", "-f", "qcow2", "-o", "backing_file="+backingImage,
+		vmImage, "60000M")
```
tpepper: This 60G number should be a variable and documented.  60G is a significant chunk.  It's big enough to make a VM able to run a long time with a bunch of ceph or docker left over trash, and then quietly die a few months down the line inexplicably.  But it also may be more than somebody wants to give to their Ciao development slowly incrementally filling up their disk.  I think single VM needs close to 20G for a full-ish Ubuntu install and various downloaded images and their copy into ceph, but am not sure the specific number.  Documenting this well will help the users and a config knob or greppable variable name will help them deal with future issues.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+	vmImage := path.Join(ws.instanceDir, "image.qcow2")
+	isoPath := path.Join(ws.instanceDir, "config.iso")
+	fsdevParam := fmt.Sprintf("local,security_model=passthrough,id=fsdev0,path=%s",
+		ws.GoPath)
+	args := []string{
+		"-qmp", fmt.Sprintf("unix:%s,server,nowait", socket), "-m", "8G", "-smp", "cpus=2",
+		"-drive", fmt.Sprintf("file=%s,if=virtio,aio=threads,format=qcow2", vmImage),
+		"-drive", fmt.Sprintf("file=%s,if=virtio,media=cdrom", isoPath),
+		"-daemonize", "-enable-kvm", "-cpu", "host",
+		"-net", "user,hostfwd=tcp::10022-:22", "-net", "nic,model=virtio",
```
tpepper: This should be a config knob port, or at least not conflict with 10022 which is used in https://download.clearlinux.org/image/start_qemu.sh and we have some overlap between people doing things with ClearLinux and with Ciao.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+		qmp.Shutdown()
+		return fmt.Errorf("VM is already running")
+	}
+
+	vmImage := path.Join(ws.instanceDir, "image.qcow2")
+	isoPath := path.Join(ws.instanceDir, "config.iso")
+	fsdevParam := fmt.Sprintf("local,security_model=passthrough,id=fsdev0,path=%s",
+		ws.GoPath)
+	args := []string{
+		"-qmp", fmt.Sprintf("unix:%s,server,nowait", socket), "-m", "8G", "-smp", "cpus=2",
```
tpepper: Can the 8G and 2 vcpu's be config knobs?  I'm liable to want to set those higher.  The cpu's I'd personally do higher be default.  The RAM I'd sometimes make more on my beefy dev box, but might also prefer less.  8G's a lot if you're running on an ultramobile laptop doing dev on the go and only have 8 or 16G total system RAM.  This also needs documented with respect to how it will interrelate with how many guest workload instances you can launch within the single VM an their cumulative resource consumption.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+
+func statusVM(ctx context.Context, instanceDir string) {
+	status := "ciao down"
+	ssh := "N/A"
+	socket := path.Join(instanceDir, "socket")
+	disconnectedCh := make(chan struct{})
+	qmp, _, err := qemu.QMPStart(ctx, socket, qemu.QMPConfig{}, disconnectedCh)
+	if err == nil {
+		status = "ciao up"
+		ssh = fmt.Sprintf("ssh 127.0.0.1 -p %d", 10022)
```
tpepper: port should be common variable, configurable

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+	}
+
+	w := new(tabwriter.Writer)
+	w.Init(os.Stdout, 0, 8, 0, '\t', 0)
+	fmt.Fprintf(w, "Status\t:\t%s\n", status)
+	fmt.Fprintf(w, "SSH\t:\t%s\n", ssh)
+	w.Flush()
+}
+
+func startHTTPServer(listener net.Listener, errCh chan error) {
```
tpepper: What does the http server serve?

testutil/ciao-down/download.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func downloadUbuntu(ctx context.Context, instanceDir string, cb progressCB) (string, error) {
+	imgPath := path.Join(instanceDir, "xenial-server-cloudimg-amd64-disk1.img")
+
+	if _, err := os.Stat(imgPath); err == nil {
+		return imgPath, nil
+	}
+
+	fmt.Println("Downloading Ubuntu 16.04")
```
tpepper: 16.04 should be in a separate variable or omitted.

testutil/ciao-down/download.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if err == nil && int(pr.downloaded/1000000)%10 != 0 {
+		downloadProgress(progress{downloadedMB: pr.totalMB, totalMB: pr.totalMB})
+	}
+
+	return
+}
+
+func downloadUbuntu(ctx context.Context, instanceDir string, cb progressCB) (string, error) {
+	imgPath := path.Join(instanceDir, "xenial-server-cloudimg-amd64-disk1.img")
```
tpepper: "xenial" should be in a separate variable or omitted.

testutil/ciao-down/download.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if _, err := os.Stat(imgPath); err == nil {
+		return imgPath, nil
+	}
+
+	fmt.Println("Downloading Ubuntu 16.04")
+
+	tmpImgPath := path.Join(instanceDir, "xenial-server-cloudimg-amd64-disk1.img.part")
```
tpepper: Same here.  The "xenial" string represents an unnecessary future maintenance point.

testutil/ciao-down/prepare.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	ws.HTTPSProxy = os.Getenv("HTTPS_PROXY")
+	if ws.HTTPSProxy == "" {
+		ws.HTTPSProxy = os.Getenv("https_proxy")
+	}
+
+	ws.NoProxy = os.Getenv("no_proxy")
+
+	pkPath := path.Join(ws.Home, ".ssh/id_rsa.pub")
```
tpepper: Is this trying to use my presumed ".ssh/id_rsa.pub"?  I don't usually have one of those, but rather use purpose built keys with a name corresponding to the usage.  Should be a config point or documented as required.  I'm starting to think I need a .ciao-downrc file.

testutil/ciao-down/prepare.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	ws.PublicKey = string(publicKey)
+
+	ws.ciaoDir = path.Join(ws.Home, ".ciao-down")
+	ws.instanceDir = path.Join(ws.ciaoDir, "instance")
+
+	data, err := exec.Command("git", "config", "--global", "user.name").Output()
+	if err == nil {
+		ws.GitUserName = strings.TrimSpace(string(data))
+	}
```
tpepper: if you want git user.name|email, it should be called out as desired, whether in the documentation, or in a printf here.

testutil/ciao-down/ciao_down.go
```diff
@@ -0,0 +1,320 @@
...
+	errCh <- err
+}
+
+func connect(errCh chan error) {
+	path, err := exec.LookPath("ssh")
+	if err != nil {
+		errCh <- fmt.Errorf("Unable to locate ssh binary")
+	}
+
+	err = syscall.Exec(path, []string{path, "127.0.0.1", "-p", "10022"},
```
mcastelino: @markdryan does it make sense to ensure the key does not get added to the known hosts
ssh -T -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i

testutil/ciao-down/prepare.go
mcastelino: We could generate a local keypair specific to this use like we do in verify.sh. It may ensure that we can always ssh in even in a non standard setup. Also it delinks us completely from the users setup.

testutil/ciao-down/cloudinit.go
```diff
@@ -69,6 +67,7 @@ write_files:
    path: /etc/apt/sources.list.d/docker.list
 
 runcmd:
+ - echo "127.0.0.1 singlevm" >> /etc/hosts
```
tpepper: This is dirty.  The name should be unique and configurable.  If you're adding to the file you need to remove from the file too.  It would be better to not touch the /etc/hosts file at all.

mcastelino: @tpepper is this not inside the VM?

testutil/ciao-down/README.md
```diff
@@ -0,0 +1,111 @@
+# ciao-down
+
+ciao-down is a small utility for setting up a VM that contains
+everything you need to run ciao's Single VM. All you need to have
+installed on your machine is:
+
+- Go 1.7 or greater
+- An ssh key
```
tpepper: The ssh key must be named "~/.ssh/id_rsa.pub"?

testutil/ciao-down/README.md
```diff
@@ -0,0 +1,111 @@
...
+package to launch and manage the VMs. All the rest is done via cloud-init.
+
+## ciao-down commands
+
+### prepare
+
+ciao-down prepare creates and configures a new ciao-down VM.  Currently,
+only one VM can be configured at any one time.  All the files associated
+with the VM are stored in ~/.ciao-down.  The first time you run ciao-down
+it will download the Ubuntu 16.04 cloud image which takes some time.  This
```
tpepper: Suggested edit 's/16.04/(currently version 16.04)/'

testutil/ciao-down/vm.go
```diff
@@ -0,0 +1,198 @@
...
+
+func statusVM(ctx context.Context, instanceDir string) {
+	status := "ciao down"
+	ssh := "N/A"
+	socket := path.Join(instanceDir, "socket")
+	disconnectedCh := make(chan struct{})
+	qmp, _, err := qemu.QMPStart(ctx, socket, qemu.QMPConfig{}, disconnectedCh)
+	if err == nil {
+		status = "ciao up"
+		ssh = fmt.Sprintf("ssh 127.0.0.1 -p %d", 10022)
```
tpepper: same comments throughout, as before the movement between files

testutil/ciao-down/cloudinit.go
```diff
@@ -55,6 +55,7 @@ write_files:
 
 runcmd:
  - echo "127.0.0.1 singlevm" >> /etc/hosts
```
tpepper: Oooh my earlier comment is irrelevant if the /etc/hosts is only modified inside the VM.  I'd still like a unique hostname.

tpepper:
All in all nice stuff!  I can merge if you prefer to incrementally make it more configurable over time or wait a day or two.

markdryan:
@mcastelino Thanks for the comments

> I think we should just document that we need nested vt support in KVM.
> We should not automatically modify the users's system setting.

> We should check in ciao-down if nested vtx is not enabled and bail out for now

Agreed.  I'll add another commit to check for this.

testutil/ciao-down/ciao_down.go
markdryan: Let me try.  That would simplify things.

testutil/ciao-down/ciao_down.go
```diff
@@ -0,0 +1,265 @@
...
+
+	_ = quitVM(ctx, ws.instanceDir)
+	err = os.RemoveAll(ws.instanceDir)
+	if err != nil {
+		errCh <- fmt.Errorf("unable to delete instance: %v", err)
+		return
+	}
+
+	knownHosts := path.Join(ws.Home, ".ssh", "known_hosts")
+	err = exec.Command("ssh-keygen", "-f", knownHosts, "-R", "[127.0.0.1]:10022").Run()
```
markdryan: @mcastelino has suggested using -o StrictHostKeyChecking=no when connecting, which would mean that this step would be no longer needed.  I'll try it out and if it works I'll remove this line.

testutil/ciao-down/ciao_down.go
```diff
@@ -0,0 +1,265 @@
...
+
+	errCh <- nil
+}
+
+func runCommand(signalCh <-chan os.Signal) error {
+	var err error
+
+	errCh := make(chan error)
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	switch os.Args[1] {
```
markdryan: Main doesn't have any args in Go.  The arguments are global variables owned by the os package and have no association with the main function.  Examples of the use of os.Args can be found in effective go.

testutil/ciao-down/download.go
markdryan: Actually, it represents an image on which we know that Single VM works flawlessly.   The explicit dependency is intentional.  The cloud-init script in cloud-init.go is also dependent on xenial.  It may work with other versions of Ubuntu but would certainly not work with Redhat or clear.

markdryan: I will move this stuff out into constants though so all these hard dependencies are easier to find.

testutil/ciao-down/prepare.go
markdryan: @tpepper , @mcastelino Agreed.  Let me do this.  I'll generate an explicit key pair as @mcastelino suggests.

testutil/ciao-down/prepare.go
markdryan: Sure.  I'll mention this in the documentation.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+ - echo "https_proxy={{.HTTPSProxy}}" >> /etc/environment
+{{end}}
+{{- if or (len .HTTPSProxy) (len .HTTPProxy) }}
+ - echo no_proxy="{{if len .NoProxy}}{{.NoProxy}},{{end}}singlevm"  >> /etc/environment
+{{end}}
+
+ - curl -X PUT -d "Downloading Go" 10.0.2.2:{{.HTTPServerPort}}
+ - echo "GOPATH={{.GoPath}}" >> /etc/environment
+ - echo "PATH=$PATH:/usr/local/go/bin:{{$.GoPath}}/bin"  >> /etc/environment
+ - {{template "PROXIES" .}}wget https://storage.googleapis.com/golang/go1.7.3.linux-amd64.tar.gz -O /tmp/go1.7.3.linux-amd64.tar.gz
```
markdryan: Good idea.  This should be easy to fix.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+ - sudo -u {{.User}} git config --global user.email {{.GitEmail}}
+{{end}}
+ - curl -X PUT -d "FINISHED" 10.0.2.2:{{.HTTPServerPort}}
+
+users:
+  - name: {{.User}}
+    gecos: CIAO Demo User
+    lock-passwd: true
+    shell: /bin/bash
+    passwd: $1$vzmNmLLD$04bivxcjdXRzZLUd.enRl1
```
markdryan: This hash needs to be removed.  It should never be used as lock-passwd is set to true.  I just left it in for debugging but now everything is working it can be deleted.

testutil/ciao-down/cloudinit.go
```diff
@@ -0,0 +1,427 @@
...
+    sudo: ALL=(ALL) NOPASSWD:ALL
+    ssh-authorized-keys:
+    - {{.PublicKey}}
+`
+
+const metaDataTemplate = `
+{
+  "uuid": "ddb4a2de-e5a5-4107-b302-e845cecd7613",
+  "hostname": "singlevm"
```
markdryan: Right now ciao-down only supports a single VM.  If you try to run ciao-down prepare twice you'll get an error the second time around.  So its not currently possible to get confused.  I'll enter an issue for updating ciao-down to support multiple VMs.

markdryan:
@tpepper and @mcastelino thanks for all the comments and suggestions.  I'll prepare some new commits that address most of your concerns today and enter issues for the bigger items.

testutil/ciao-down/ciao_down.go
tpepper: I meant to say main package's os.Arg's, but that then answers the question right away itself.  Arg's are in the os package and not the main package or tied to its main() function.

It sure feels weird if any piece of code anywhere could check if os.Args[42] is set and use it if so.

testutil/ciao-down/download.go
tpepper: I agree it's the one.  But at some point it will be superceded.  It would be nice to go find the one stub of code that says:

#this is the preferred, tested image
var imagePrettyName = "Ubuntu 16.04 Xenial"
var imageDownloadUrl = "https://..."

markdryan:
@tpepper This should be ready to merge now.  Unfortunately, there seems to be an issue with coveralls.

mcastelino:
@markdryan testing it on ubuntu 14.04 I get an error

```
Checking environment
Unable to generate SSH key pair : exit status 1
```

Did not debug it futher. The command works itself works in isolation
`ssh-keygen -f testmykey -t rsa -N ''`

tpepper:
Github had an outage earlier today in their API.  Perhaps that collided with the coveralls run?

mcastelino:
@markdryan Looks like the .ciao-down directory is not being created. Hence the failure of the ssh key generation.

testutil/ciao-down/download.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func getUbuntu(ctx context.Context, dest io.WriteCloser, cb progressCB) (err error) {
+	defer func() {
+		err1 := dest.Close()
+		if err == nil && err1 != nil {
+			err = err1
+		}
+	}()
+	req, err := http.NewRequest("GET", guestDownloadURL, nil)
```
mcastelino: @markdryan this fails on clear for some reason with error 

`Unable download file https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-disk1.img: Get https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-disk1.img: x509: certificate signed by unknown authority
`

curl to the same URL on the host works fine.

mcastelino:
On Ubuntu 14.04 I get another error once I get past the ssh key error

```
Failed to launch qemu : exit status 1, Could not access KVM kernel module: Permission denied
failed to initialize KVM: Permission denied

```

The user has password less sudo setup

testutil/ciao-down/download.go
mcastelino: @tpepper thank you for root causing this @markdryan this is broken cert paths

`symlinking /etc/ssl/certs -> /usr/share/ca-certs
`

markdryan: This seems to be fixed in the latest version of clear I tried.

PR #839, opened 2016-11-16
[WIP] ansible: Use cephfs for /var/lib/ciao/images

albertomurillo:


albertomurillo:
cephfs is not being used right now so i will close this PR

PR #848, opened 2016-11-18
ciao-cli: Improve template functionality

markdryan:
This PR adds two new functions to the template language understood
by ciao-cli:

- tojson outputs the specified object in json format, e.g., {{tojson .}}
- filter operates on an slice or array of structures.  It allows the caller
  to filter the input array based on the value of a single field.
  The function returns a slice containing only the objects that satisfy the
  filter, e.g.
- select operates on a slice of structs.  It outputs the value of a specified
  field for each struct on a new line , e.g.,
  {{select . "Name"}}

  ciao-cli image list -f '{{$x := filter . "Protected" "true"}}{{len $x}}'

  outputs the number of protected images maintained by the image service.

The image list command was also modified to called the template utility
functions in template.go.  This was necessary to ensure that our new
functions are available to the image list command.

Finally, a the createTemplate function was using the wrong name for the
templates it created.

### DIFF
--- a/ciao-cli/event.go
+++ b/ciao-cli/event.go
@@ -58,6 +58,7 @@ The template passed to the -f option operates on a
 	Message   string    // Event message
 }
 `)
+	fmt.Fprintln(os.Stderr, templateFunctionHelp)
 	os.Exit(2)
 }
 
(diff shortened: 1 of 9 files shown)


markdryan:
Going to make a quick change to the way errors are handled.  Don't merge yet.

markdryan:
Should be good to go now.

PR #850, opened 2016-11-18
ciao-controller: Change persistent data paths

obedmr:
In favor of cleaner deployment, we're changing default paths
for persistent data file of ciao-controller and ``ciao-image`` to
``/var/lib/ciao/data`` directory.

It will create the directories in case that they don't exists.
In case of a diferent path is desired, ``ciao-controller`` command
will require the proper parameters to configure the paths.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,10 @@ script:
    - go list ./... | grep -v github.com/01org/ciao/vendor | xargs go list -f '{{.Dir}}' | xargs -L 1 ineffassign
    - go list ./... | grep -v github.com/01org/ciao/vendor | xargs go list -f '{{.Dir}}' | xargs gofmt -s -l | wc -l | xargs -I % bash -c "test % -eq 0"
    - sudo mkdir -p /var/lib/ciao/instances
-   - sudo chmod 0777 /var/lib/ciao/instances
+   - sudo mkdir -p /var/lib/ciao/data/controller/{workloads,tables}
+   - sudo chmod 0777 -R /var/lib/ciao
+   - cp $GOPATH/src/github.com/01org/ciao/ciao-controller/workloads/* /var/lib/ciao/data/controller/workloads/
+   - cp $GOPATH/src/github.com/01org/ciao/ciao-controller/tables/* /var/lib/ciao/data/controller/tables/
    - test-cases -v -timeout 9 -coverprofile /tmp/cover.out -short github.com/01org/ciao/ciao-controller/...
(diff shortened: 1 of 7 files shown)


albertomurillo:
The data paths are specified in ansible role ciao-controller/templates/ciao-controller.service.j2
This change should update the systemd units too.

obedmr:
@albertomurillo could you please take a look on ``ciao-deployment`` related files.

obedmr:
@albertomurillo I did the updates in ansible for creating the required ciao-controller path, how is it for you?

albertomurillo:
+1
Looks good for ciao_dev, the packages should create those folders anyways in a tmpfiles systemd unit.

markdryan:
@kaccardi Is this okay for you?

kaccardi:
probably ok - but I was worried about moving the transient datastore out of /tmp - but I guess it doesn't matter much since we took the datastore access out of the fast path for the most part. there are some queries which do go directly to sql and don't use the cache, and these will slow I believe if we move the transient datastore out of /tmp. this is the events, node summary, and frame statistics. Is there a reason we can't just leave the transient datastore in tmp?

obedmr:
@kaccardi, the reason was just to keep all db files in a single location. But, if  you think it's better to keep transient db in /tmp, we can keep it there. What do you think?

tpepper:
@kaccardi is your desire in https://github.com/01org/ciao/pull/850#issuecomment-263371286 that the transient datastore be in a ram based tmpfs?

kaccardi:
@tpepper yes basically.

obedmr:
sure, so, /me updating PR, thanks

obedmr:
@kaccardi, it's done, I only changed the workloads and tables locations.

tpepper:
Hmm.  If we need the /tmp store on a tmpfs, we need to make it so.  Not all host OS's put /tmp on a tmpfs out of the box.

obedmr:
@kaccardi @tpepper  is this ``tmpfs`` setup planned to be in CIAO code? if so, I'd think it could be a separate piece of work we'd need to think.

tpepper:
Yes I agree.  If there's a requirement for tmpfs, it is a separate issue that needs opened and addressed.  It shouldn't block this PR.

obedmr:
@tpepper @kaccardi  I just created the issue for the ``tmpfs`` topic #897 
Please feel free to edit or comment on it.

@kaccardi is there something else you would like to add on this PR before proceeding to merge?

PR #852, opened 2016-11-22
ciao-cert: Rename signer flag for clarity

bryteise:
Previously the -server and -server-cert flag were used to indicate that
the certificate being generated would be either the signing certificate
or signed by the signing certificate. This change updates the flags to
be -anchor and -anchor-cert to separate the ssntp.SERVER role term from
the trust anchor term used in the same command and APIs.

---

The motivation for this is that I was incredibly confused by the term server when talking about the certificate and I thought there was some sort of relation between that and the ssntp role that was embedded in the certificate. Based on existing tests and my own findings that does not appear to be the case and I do not find other documents which contradict these. That being the case, simply renaming the options and a few functions/documentation avoids the possible confusion of these two distinct uses.

As an aside my findings are that the anchor doesn't need to be a specific role (it was largely hinted through documentation/use that the ssntp.SCHEDULER or ssntp.SERVER roles were supposed to be the anchor though). There doesn't seem to be any tests which contradict this finding and there are tests which have for example ssntp.AGENT role certificates as the trust anchor. If this is not supposed to happen then this change needs further refinement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,13 +25,13 @@ before_install:
 before_script:
    - sudo mkdir -p /etc/pki/ciao/
    - sudo mkdir -p /var/lib/ciao/logs/scheduler
-   - sudo -E $GOPATH/bin/ciao-cert -directory /etc/pki/ciao -host localhost -server -role scheduler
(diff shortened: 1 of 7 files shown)


albertomurillo:
+1 on the ansible part

rbradford:
Tested on SingleVM.

PR #854, opened 2016-11-22
Ensure volumes are marked as in-use

rbradford:
Ensure that the volume status is marked as "in-use" by setting the status when the attachment was created.

A previous version also updated as part of the instance stats command but that would override any detaching state (found via unit tests.)

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1510,8 +1510,25 @@ func (ds *Datastore) CreateStorageAttachment(instanceID string, blockID string, 
 	ds.attachLock.Unlock()
 
 	err := ds.db.createStorageAttachment(a)
+	if err != nil {
+		return types.StorageAttachment{}, fmt.Errorf("error creating storage attachment: %v", err)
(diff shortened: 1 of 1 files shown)


markdryan:
Work's fine for me.  Fixes an issue I encountered with the early storage tests.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1503,15 +1503,32 @@ func (ds *Datastore) CreateStorageAttachment(instanceID string, blockID string,
...
 
+	err := ds.db.createStorageAttachment(a)
+	if err != nil {
+		return types.StorageAttachment{}, fmt.Errorf("error creating storage attachment: %v", err)
+	}
+
+	// ensure that the volume is marked in use as we have created an attachment
+	bd, err := ds.GetBlockDevice(blockID)
+	if err != nil {
+		err := ds.db.deleteStorageAttachment(a.ID)
```
rbradford: @markdryan good catch  - fixed version uploaded!

markdryan:
LGTM

Issue #851, opened 2016-11-21
block_data not cleaned up when cnci fails to start


kaccardi:
if the CNCI fails to start, the block_data object in the sqlite3 database still indicates that the volume is "in-use"

rbradford:
I think I have a fix .. how do you test it?

kaccardi:
@rbradford you can force the CNCI fail to launch by improperly configuring the network node. For example, delete the /etc/ceph directory on the network node, and the launcher on the node will be unable to access the cnci volume.

tpepper:
I'm mucking in the same code with error paths and @rbradford's change here looks reasonable to me.

PR #855, opened 2016-11-22
ciao-controller: Make CNCI storage ephemeral

rbradford:
Refactor the instance launching code to make storage for the CNCI
ephemeral so that it will be cleaned up when the CNCI is deleted.

Fixes: #851

### DIFF
--- a/ciao-controller/instance.go
+++ b/ciao-controller/instance.go
@@ -87,18 +87,18 @@ func newInstance(ctl *controller, tenantID string, workload *types.Workload) (*i
 }
 
 func (i *instance) Add() error {
+	ds := i.ctl.ds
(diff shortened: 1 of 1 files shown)


rbradford:
Need verification from @kaccardi on how to reproduce issue seen in #851.

rbradford:
I'm pretty sure this is the right thing to do in general.

I'm not sure it fixes the bug though because of the slightly different way the CNCI workload is created -let me investigate further.

markdryan:
I haven't been able to test the error case that this PR fixes but I can confirm that it doesn't break anything.  The BAT tests are all passing on SingleVM, even the fancy new state of the art storage BAT tests, which I haven't committed yet.

PR #856, opened 2016-11-22
Use a memory datastore for datastore tests

### LINKED ISSUE #403: Improve datastore unit testing by creating fake persistent datastores
Change the way that the datastore unit tests are written to use a fake persistent datastores that can be instrumented to return success or failure. This will make it easier to test error paths in datastore.

rbradford:
Introduce a memory backed persistentStore that can be used as a backend to controller's datastore. Currently the implementation is complete enough to be used for the datastore test.

This partially fixes #403.

A follow-up change will enhance the backend to be usable for the controller test.


bryteise:
@rbradford Would it be possible to use the memory backend more generically (like for testing tracing?)

rbradford:
@bryteise Yes possibly...currently there is only as much filled out to satisfy the datastore test. I know more will be needed for the main controller test.

markdryan:
@kaccardi Could you take a look at this one as well?

rbradford:
@tpepper I don't think that's necessary as:

a.) datastore.go uses plenty of locking already
b.) go test -race shows no problems.

tpepper:
I would like to see more information in the commit for the persistentStore interface regarding who has the responsibility for data integrity assurances, the interface user or an author of code implementing this interface.  The interface enables swizzling of the backend.  The backends may or may not need locking to manager their internal consistency.  For the _most_ part ( @rbradford detailing which parts are and are not is what I'd really like to see) controller has a write through in-memory cache, with information written through to the backend at controller runtime and any persistent data read in from the backend at controller startup time, leading to a situation without much concurrency.

I don't yet believe the argument that there is concurrency safety here.  The "a)" statement of "datastore.go does a lot of locking" is proof, then the interface documentation should state that the concurrency is to be managed by the interface user, as is today done by datastore.go.  The "b)" statement of "it's -race clean" is not proof to me, because I believe -race only catches observed racing and does not prove that no races exist.  We cannot assert that the implementation is race free given the face that with the sleeps removed tests run (in all observed cases so far) without races reported, but that may be because the in-memory implementation of the interface is blazingly fast and rarely in practice has contention.  If we cannot assert that the implementation is race free because of datastore.go's preferred usage of the interface, then that leaves us in a place where we don't know if we have races and we haven't a plan on what is needed to protect against them.

I want to see a design and documentation that convinces me not that "most" issues are protect or that we don't have "much" concurrency, but that proves to me that we have no issues.

I'm ok with this taking the form of a new issue which documents known issues and protections and track the need to close the gap between those.  It is certainly nice that this implementation removes the sleeps from the unit test and has not yet exhibited runtime unit-test races.  That is undeniably an improvement on the current situation.

kaccardi:
@tpepper I filed an issue to track your documentation request #882 if you want to merge without those changes.

kaccardi:
@rbradford one final hesitation - why does coveralls show coverage dropping 3.9% with this change. Is that real?

kaccardi:
coveralls indicates that coverage drops significantly with this change. I think it's because our sqlite3_test is insufficient and it previously relied on datastore_test for coverage. I hate letting the coverage drop to this extent, but on the other hand, this is a more correct design for testing so I'd be ok to take the hit.

tpepper:
Github seems confused here:

```
Changes approved
All checks have passed
coverage/coveralls — Coverage increased (+0.02%) to 61.069%
```

ciao-controller/internal/datastore/memorydb.go
```diff
@@ -0,0 +1,263 @@
+package datastore
```
markdryan: We're missing a copyright notice on this file.

rbradford: Eek. Fixed.

tpepper:
@rbradford looks like this needs a simple rebase conflict fixup

rbradford:
@tpepper / @markdryan PR rebased and ready to roll.

kaccardi:
@rbradford thanks for adding all that extra sqlite3 coverage.

PR #857, opened 2016-11-22
Minor SingleVM fixes

rbradford:
Fix a couple of nits left over from the integration of the images as volumes work.

### DIFF
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -13,7 +13,4 @@ sudo pkill -F /tmp/dnsmasq.ciaovlan.pid
 sudo docker rm -v -f ceph-demo
 sudo rm /etc/ceph/*
 sudo rm -rf /var/lib/ciao/ciao-image
-sudo rm /var/lib/ciao/images/4e16e743-265a-4bf2-9fd1-57ada0b28904
-sudo rm /var/lib/ciao/images/df3768da-31f5-4ba6-82f0-127a1a705169
-sudo rm /var/lib/ciao/images/73a86d7e-93c0-480e-9c41-ab42f69b7799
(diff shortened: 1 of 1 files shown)


rbradford:
@mcastelino Please can you review as OWNER.

mcastelino:
LGTM.

rbradford:
@albertomurillo yes.

markdryan:
> Now that images are stored in ceph the CNCI image will be converted back
> on by the controller. Let's save some time and not convert it before
> uploading.

This PR actually seems to slow down single VM on my setup.  I'm guessing this is because the qcow file is compressed and so considerably smaller than the raw file (357MB compared to 857MB).  On my setup, it takes 4.25 seconds to convert the cnci image to qcow and upload it while it takes 8 seconds to upload the raw image.  This surprises me a little bit as the uploader of qcow images still needs to wait for the conversion to take place.  In singleVM at least it seems the task of uploading the image takes longer than the conversion.

albertomurillo:
+1

markdryan:
@albertomurillo What are you +1ing?  Do you also see the slow down in Single VM setup.sh when applying this patch?

albertomurillo:
The controller uses qemu-img convert to convert the image file to rbd block device.
There is no need to convert from qcow2 -> Raw -> RBD when it can be done directly qcow2 -> RBD

markdryan:
Right now the code is doing

raw -> qcow2 -> RBD

The patch changes this to

raw -> RBD

Which reduces an unnecessary conversion.  Unfortunately, this actually slows things down presumably as the compressed qcow2 images are quicker to upload than the raw images.

rbradford:
@markdryan dropped controversial change. :-)

Issue #662, opened 2016-10-12
ciao-cli image commands are very resource intensive


markdryan:
In particular the download and upload commands load the entire contents of the image into memory.  These images could be quite large, gigabytes, so we probably don't want to be reading these into memory.

imageDownloadCommand.run and uploadTenantImage are the biggest offenders here.  Note there are also unchecked error values returned from Read.

PR #859, opened 2016-11-22
Fix resource issues with ciao-cli image

rbradford:
@markdryan found some resource issues with ciao-cli image commands in #662 this PR addresses those by not loading the entire image into memory when uploading and downloading.

### DIFF
--- a/ciao-cli/image.go
+++ b/ciao-cli/image.go
@@ -17,7 +17,6 @@
 package main
 
 import (
-	"bytes"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


markdryan:
Looks good to me.  I was able to upload new images.  The upload was comparable too or perhaps slightly quicker than the old code.  I was going to suggest wrapping the file handle we're passing to images.Upload in a bufio.NewReader() but this doesn't work as images.Upload wants a ReadSeeker.  So looks good to me.

ciao-cli/image.go
```diff
@@ -362,19 +361,19 @@ func (cmd *imageDownloadCommand) run(args []string) error {
...
-		fatalf("Could not read [%s]\n", err)
-	}
-
-	if cmd.file == "" {
-		fmt.Printf("%s\n", b)
-	} else {
-		err := ioutil.WriteFile(cmd.file, b, 0644)
+	dest := os.Stdout
+	if cmd.file != "" {
+		dest, err = os.Create(cmd.file)
```
markdryan: We're not closing the file we're creating.  I think we need to close it and check the err from close to be sure all the data is correctly written.

rbradford: Fixed with a defer!

markdryan: we should really check the return value from close as we're writing to the file.

rbradford: okay. done.

Issue #864, opened 2016-11-23
ciao-cli uses incorrect identity API for unscoped tokens


kaccardi:
unscoped tokens can do very little, and when they do, they need to be using /auth endpoings. ciao-cli uses the identity endpoint for scoped tokens to list projects rather than the api for unscoped tokens.

PR #813, opened 2016-11-10
singlevm: Add keystone support

bwarden:
* Replaces embedded values with variables defined near the top of
setup.sh.
* Generates configuration.yaml automatically and removes static copy
from repo.
* Adds CIAO_* variables to ~/local/demo.sh in a bit more sustainable
fashion.
* Cleans up some long lines.

Signed-off-by: Brett T. Warden <brett.t.warden@intel.com>

PR #865, opened 2016-11-23
ciao-cli: use correct identity endpoint for unscoped token

kaccardi:
When using an unscoped token, you need to use a different
endpoint to list projects than when you have a scoped token.

Fixes: #864 


markdryan:
This patch works fine in conjunction with the Single VM Keystone changes #813 but does not work without them.  The BAT tests fail on Single VM when just this patch is merged.  So we need to delay merging until #813 gets merged.

Issue #875, opened 2016-11-28
Data volumes are not persisted


markdryan:
Ciao-controller does not seem to persist any data volumes.  This makes it impossible to test some of the use cases, for example, attaching a data volume at boot, as testing these use cases require you to create a volume, update the controller tables and restart controller. Unfortunately, when restarting controller the data volumes disappear.

Here's an example workflow.

```
$ ciao-cli volume list
Volume #1
	Name             [Storage for instance: b643dcd5-1a14-408e-8d27-4987f339dc3c]
	Size             [0 GB]
	UUID             [237a7a73-2285-454e-81b1-3a32b346c8c9]
	Status           [in-use]
	Description      []

$ ciao-cli volume add --name "test"
Created new volume: 015b74c1-04d0-44c0-9485-b486b711c954
markus@singlevm:~/go-fork/src/github.com/01org/ciao/testutil/singlevm$ ciao-cli volume list
Volume #1
	Name             [Storage for instance: b643dcd5-1a14-408e-8d27-4987f339dc3c]
	Size             [0 GB]
	UUID             [237a7a73-2285-454e-81b1-3a32b346c8c9]
	Status           [in-use]
	Description      []

Volume #2
	Name             [test]
	Size             [1 GB]
	UUID             [015b74c1-04d0-44c0-9485-b486b711c954]
	Status           [available]
	Description      []

$ sudo killall ciao-controller

$ cd ~/local/
$ ./run_controller.sh 
E1128 10:58:20.985761   25213 main.go:169] ========================
E1128 10:58:20.986141   25213 main.go:170] Identity URL: http://127.0.0.1:43613
E1128 10:58:20.986198   25213 main.go:171] Please
E1128 10:58:20.986222   25213 main.go:172] export CIAO_IDENTITY=http://127.0.0.1:43613
E1128 10:58:20.986276   25213 main.go:173] ========================

$ export CIAO_IDENTITY=http://127.0.0.1:43613
$ cd -
$ ciao-cli volume list
Volume #1
	Name             [Storage for instance: a5b5c9d0-8f5a-4908-b6a6-5496c3be7fc3]
	Size             [0 GB]
	UUID             [ba2b7b0e-6283-40f7-8cf8-8577551cf166]
	Status           [in-use]
	Description      []
```

markdryan:
User error.  run_controller.sh deletes the controller.db.

Issue #872, opened 2016-11-25
Attaching a volume to a stopped instance creates a zombie


markdryan:
Check out the following steps:

```
$ ciao-cli volume add -name zombie -size 1
Created new volume: 0ee6471f-0f68-4875-bce6-acade1c6771c

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 4c573b13-26b7-4f82-beea-1d2380c36709 active 172.16.0.2 198.51.100.82 33002
2 7181b76b-5c60-40b6-8272-73b4197067aa active 172.16.0.3 198.51.100.82 33003
3 94131758-7485-44c4-b883-d6a9f9ab2ea9 active 172.16.0.4 198.51.100.82 33004
4 4d980203-299f-460d-9905-262ac9c7ccf2 exited 172.16.0.5 198.51.100.82 33005

$ ciao-cli volume attach --instance 4d980203-299f-460d-9905-262ac9c7ccf2 --volume 0ee6471f-0f68-4875-bce6-acade1c6771c
Attached volume: 0ee6471f-0f68-4875-bce6-acade1c6771c

$ sleep 60

$ ciao-cli volume show --volume  0ee6471f-0f68-4875-bce6-acade1c6771c
	Name             [zombie]
	Size             [1 GB]
	UUID             [0ee6471f-0f68-4875-bce6-acade1c6771c]
	Status           [attaching]
```

kaccardi:
@markdryan I debugged this issue - the reason the status never updates to "in-use" here is because the instance stats for the exited instance only show the old attachments and are not updated to show the attachment that was added while the instance was exited. State change on volume attach relies completely on the stats since there is no "attached" event. If you had restarted the stopped instance, you would have seen the state change as the updated stats came in. So - this is either operator error in that you need to explicitly restart the instance before you can expect to show the attachment as fully attached (i.e. "in-use"), or you need to change launcher to send the stats with the updated attachment list.

kaccardi:
needs a retest with the change in #886

markdryan:
Agreed.   This is a launcher issue.  I've submitted a PR #887.   I've tested this PR in conjunction with #886 and the other PR submitted to fix the storage BAT tests.  All the tests pass.

Issue #871, opened 2016-11-25
Attaching Volume to running container should fail


markdryan:
It's not possible to attach a volume to a running container.  Volumes can only be attach to containers at creation time, i.e., there's no hotplugging of volumes to containers.  ciao-cli should therefore return an error.  Instead it attempts to attach the volume to a container.  It returns success but the volume is not attached.  It remains stuck in the attaching state for ever and cannot be deleted.   Zombie volume.

```
$ ciao-cli volume add -name zombie -size 1
Created new volume: 6817ab36-6b82-4088-8bd4-31be35c40ffb
$ ciao-cli instance add --workload ca957444-fa46-11e5-94f9-38607786d9ec
Created new instance: 7181b76b-5c60-40b6-8272-73b4197067aa
$ ciao-cli volume attach --instance 7181b76b-5c60-40b6-8272-73b4197067aa --volume 6817ab36-6b82-4088-8bd4-31be35c40ffb
Attached volume: 6817ab36-6b82-4088-8bd4-31be35c40ffb
$ sleep 60
$ ciao-cli volume show --volume 6817ab36-6b82-4088-8bd4-31be35c40ffb
	Name             [zombie]
	Size             [1 GB]
	UUID             [6817ab36-6b82-4088-8bd4-31be35c40ffb]
	Status           [attaching]
	Description      []
$ ciao-cli volume delete --volume 6817ab36-6b82-4088-8bd4-31be35c40ffb
F1125 14:27:50.350790   31705 main.go:403] ciao-cli FATAL: Expected HTTP response code [202 204] when accessing [DELETE https://singlevm:8776/v2/f452bbc7-5076-44d5-922c-3b9d2ce1503f/volumes/6817ab36-6b82-4088-8bd4-31be35c40ffb], but got 403 instead
Forbidden
```

This issue can also be reproduced by the unsubmitted TestAttachToRunningContainer BAT test.

markdryan:
I have a patch for this but have not been able to fully test because of #875

kaccardi:
@markdryan I reviewed the code in controller - it looks like this error path should cause the volume state to be reset to available. I also looked at the launcher code and it looks correct. @tpepper does the scheduler need to forward these errors explicitly? I don't see anything in the scheduler code to indicate that it does anything for these errors.

Issue #870, opened 2016-11-25
Detaching a bootable volume should fail


markdryan:
This is easy to reproduce.  Create a new VM instance, identify the volume of the instance's rootfs and detach it.  No error is returned from ciao-cli.  However, the volume is not detached.  It is stuck in the detaching state for ever.

```
$ ciao-cli instance add --workload 79034317-3beb-447e-987d-4e310a8cf410
Created new instance: bf4c43ea-0241-42ec-9c7c-96e9d94b07ac

$ ciao-cli instance show --instance bf4c43ea-0241-42ec-9c7c-96e9d94b07ac -f '{{range .OsExtendedVolumesVolumesAttached}}{{println .}}{{end}}'
48c11ee6-c508-4f58-8d99-24ce45b5bb73

$ ciao-cli volume detach --volume 48c11ee6-c508-4f58-8d99-24ce45b5bb73
Detached volume: 48c11ee6-c508-4f58-8d99-24ce45b5bb73

<wait 5 minutes>

$ ciao-cli volume show --volume 48c11ee6-c508-4f58-8d99-24ce45b5bb73
	Name             [Storage for instance: bf4c43ea-0241-42ec-9c7c-96e9d94b07ac]
	Size             [0 GB]
	UUID             [48c11ee6-c508-4f58-8d99-24ce45b5bb73]
	Status           [detaching]
	Description      []
```

The volume does however get deleted when the instance is deleted so at least it's not leaked.

This issue can also be reproduced by the unsubmitted TestDetachRootFS BAT test.

PR #876, opened 2016-11-28
Add Storage bat tests

### LINKED ISSUE #682: Create BAT tests for storage
We need to write some bat tests for storage.  In sort we should test the main storage use cases and some negative use cases, e.g.,
1.  Boot from volume.
2. Attach a non bootable volume to an instance at boot.
3. Attach a volume to an instance after it has been created.
4. Detach a volume.
5. Attach a volume to a container at boot.
6. Attach a volume to a container after it has been created.  This should fail.
7. Delete a volume
8. Delete an in-use volume.
9. Delete a boot from volume instance and check state of volume.

markdryan:
bat: Add storage bat tests
    
This PR adds the following storage BAT tests
    
- Test bootable volumes are created and deleted correctly
- Check bootable volumes of stopped instances are in-use
- Check that in-use volumes cannot be deleted
- Check that the rootfs volumes cannot be detached
- Check that volumes can be added, listed and deleted
- Check attaching volumes become available when deleting their instance.
- Check that a deleting a detaching volume fails
- Check that we can attach and detach volumes
- Check that we cannot attach to a running container
- Check that we cannot attach a volume to two instances
- Check we can attach a volume to a stopped instance
    
As we now have multiple BAT tests that create and delete instances in the same tenant the BAT tests should be run with test-cases, rather than go test, as the latter runs tests for separate packages concurrently, which is not what we currently want.

This PR also splits the bat.go file into multiple go files to make it easier to navigate.  Finally, it updates the ciao-cli command to take advantage of the new functions in the ciao-cli template language. 
    
Fixes #682

### DIFF
--- a/_release/bat/storage_bat/storage_bat_test.go
+++ b/_release/bat/storage_bat/storage_bat_test.go
@@ -1,0 +1,636 @@
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage_bat
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/01org/ciao/bat"
+)
+
+const standardTimeout = time.Second * 300
+
+// TODO: We're launching an instance from a known workload UUID that
+// that has no data volumes attached.
+
+func createSpecificInstance(ctx context.Context, t *testing.T, tenant, workloadID string,
+	mustBeActive bool) string {
+	wklds, err := bat.GetAllWorkloads(ctx, tenant)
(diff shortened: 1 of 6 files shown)


markdryan:
Not all these new BAT tests are currently passing.  This is due to the following bugs in controller:

- #875 (This doesn't cause a BAT test failure but needs to be fixed to manually test storage)
- #872
- #871
- #870
- #867
- #869  (This doesn't cause a BAT test failure)

_release/bat/storage_bat/storage_bat_test.go
```diff
@@ -0,0 +1,623 @@
...
+
+	err = bat.DeleteInstance(ctx, "", instanceID)
+	if err != nil {
+		t.Errorf("Failed to delete instance %s : %v", instanceID, err)
+	}
+	instanceID = ""
+
+	err = bat.WaitForVolumeStatus(ctx, "", volumeID, "available")
+	if err != nil {
+		t.Errorf("Timed out waiting for volume %s to become availabled : %v",
```
kaccardi: is "availabled" a typo, or a new verb describing the act of becoming available?

markdryan:
I update the commit to fix the spelling mistake.  I also changed the behaviour of the AttachedToRunningContainer test to reflect the expected behaviour and fixed some issues with the use of defer to clean up volumes in some of the tests.  All tests are now passing locally for me with

#885 , #886, #887

markdryan:
@tpepper I've rebased.  All tests are passing on master now.  We can merge.

markdryan:
Rebased.  Tests still pass.

@tpepper Could you review and merge?

PR #878, opened 2016-11-28
ciao-image: Fix open file issue on posix write ops

### LINKED ISSUE #661: ciao-image leaks file descriptors each time you upload a new file
We need a close in ciao-image/datatstore/posix.go:posix.Write

obedmr:
When writing a new image on posix backend, we open a new file
and drop all image data there, this opened file needs to be
closed before ending the function.

Fixes #661

### DIFF
--- a/ciao-image/datastore/posix.go
+++ b/ciao-image/datastore/posix.go
@@ -40,7 +40,15 @@ func (p *Posix) Write(ID string, body io.Reader) (int64, error) {
 
 	buf := make([]byte, 1<<16)
 
-	return io.CopyBuffer(image, body, buf)
+	length, err := io.CopyBuffer(image, body, buf)
+	defer func() {
+		err1 := image.Close()
+		if err == nil {
+			err = err1
+		}
+	}()
+
+	return length, err
 }
 
 // Delete removes an image from the posix filesystem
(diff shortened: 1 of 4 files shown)


obedmr:
@rbradford, you're right, my bad, I forgot to rename it. I'm changing it to something more descriptive. Taking https://golang.org/pkg/io/#CopyBuffer as base, I'm thingking in something like ``length`` or ``file_length``

rbradford:
@obedmr Is there any value in returning that? Does it get used anywhere?

obedmr:
@rbradford, you have a good point, as far I can see, that lenght value is not used anywhere. So, We should change this to only returning the error. It would imply a change on https://github.com/01org/ciao/blob/master/ciao-image/datastore/ceph.go#L65 What do you think?

rbradford:
@obedmr Yes. You should change that file too.

ciao-image/datastore/posix.go
```diff
@@ -27,20 +27,23 @@ type Posix struct {
...
 	if err != nil {
-		return 0, err
+		return err
 	}
 
 	buf := make([]byte, 1<<16)
 
-	return io.CopyBuffer(image, body, buf)
+	_, err = io.CopyBuffer(image, body, buf)
+	defer image.Close()
```
markdryan: As you are writing to image you need to check the return value from Close.  I usually do this by using a named return value, e.g.,

```
func fn() (err error) {
   ....

   defer func() {
     err1 := image.Close()
     if err == nil {
       err = err1
     }
    }()
}

```

obedmr:
@rbradford @markdryan done, thanks for the feedback.

PR #883, opened 2016-11-28
ciao-controller: Add boot flag to attachment

kaccardi:
Add a flag to indicate whether the attachment is for a boot volume.
If a user requests a detach of an attachment that is a boot device,
send a 403 to indicate this is a forbidden operation.

Fixes: #870


markdryan:
It does indeed fix #870.  Thanks.

PR #885, opened 2016-11-28
ciao-scheduler: add forwarding rules for attach/detach volume failure

kaccardi:
Modify the scheduler forwarding rules to forward attach/detach errors
to controller.

Fixes: #871


markdryan:
I confirm that this fixes #871.  I have updated the storage_bat tests to not expect an error from ciao-cli volume attach but to expect the status of the volume to become available.  The test now passes.

@tpepper If this is okay for you we can merge.

markdryan:
@tpepper  Could you review and merge?  This fixes one of the two remaining storage BAT issues.  The other is #887.

Issue #867, opened 2016-11-24
Deleting an instance can lead to permenantly attaching volumes


markdryan:
This is easy to reproduce and was discovered when writing BAT test.  The BAT test in question is
TestDeleteInstanceWhileAttaching.  It hasn't been merged yet but can be found on my fork.  The code is here:

https://github.com/markdryan/ciao/blob/storage_bat/_release/bat/storage_bat/storage_bat_test.go#L257

The test creates a new instance and a new volume.  It then attaches the volume to the instance but deletes the instance before the volume has finished attaching.  The volume remains in the attaching state for ever and cannot be deleted as can be seen from the logs.   Here's a log

```
=== RUN   TestDeleteInstanceWhileAttaching
--- FAIL: TestDeleteInstanceWhileAttaching (6.10s)
	storage_bat_test.go:272: Failed to delete volume b602e609-48fa-4283-9732-7521f506b83e : failed to launch ciao-cli [volume delete --volume b602e609-48fa-4283-9732-7521f506b83e] : exit status 255
		F1124 10:39:34.103966    3381 main.go:403] ciao-cli FATAL: Expected HTTP response code [202 204] when accessing [DELETE https://singlevm:8776/v2/f452bbc7-5076-44d5-922c-3b9d2ce1503f/volumes/b602e609-48fa-4283-9732-7521f506b83e], but got 403 instead
		Forbidden
		null
exit status 1
FAIL	github.com/01org/ciao/_release/bat/storage_bat	6.102s

$ ciao-cli volume show --volume b602e609-48fa-4283-9732-7521f506b83e
	Name             []
	Size             [1 GB]
	UUID             [b602e609-48fa-4283-9732-7521f506b83e]
	Status           [attaching]
	Description      []
```

The status of the volume should return to available after the instance has been deleted.  This is what happens if you wait for the volume to be attached and then delete the instance.

kaccardi:
The reason this fails is because of the way that controller currently creates attachments. It does not create the actual attachment until a stat comes in from launcher indicating that a volume is attached - However - it DOES update the state of the volume to be "attaching" before the actual attachment happens. So, if an instance is deleted prior to any stats being sent, then the attachment struct would not have been created and the block device would have never been updated. Controller needs to be modified to create an attachment at request time.

PR #886, opened 2016-11-29
ciao-controller: create attachment at request time

kaccardi:
Create an attachment struct when the attachment is first created, rather
than waiting for a stats update.

Fixes: #867


markdryan:
The patch needs to be rebased now #883 is merged.  I updated it locally to get it to compile and it does fix the issue.

kaccardi:
@markdryan I rebased as you requested.

PR #887, opened 2016-11-29
ciao-launcher: Always update stats when volumes change

### LINKED ISSUE #872: Attaching a volume to a stopped instance creates a zombie

markdryan:
Launcher was not sending updated volume information for exitted instances.  This
led to a situation where volumes attached to exiting instances remained permanently
stuck in the attaching state until the instance was restarted.

Fixes #872

### DIFF
--- a/ciao-launcher/instance_test.go
+++ b/ciao-launcher/instance_test.go
@@ -209,17 +209,56 @@ func (v *instanceTestState) verifyStatsUpdate(t *testing.T, cmd interface{}) {
 	}
 }
 
-func (v *instanceTestState) expectStatsUpdate(t *testing.T, ovsCh <-chan interface{}) bool {
+func (v *instanceTestState) getStatsUpdate(t *testing.T, ovsCh <-chan interface{}) *ovsStatsUpdateCmd {
 	var cmd interface{}
 	select {
 	case cmd = <-ovsCh:
 	case <-time.After(time.Second):
 		t.Error("Timed out waiting for ovsStatsUpdateCmd")
-		return false
+		return nil
 	}
 	stats, ok := cmd.(*ovsStatsUpdateCmd)
 	if !ok {
 		t.Error("Unexpected Command received on ovsCh")
+	}
+	return stats
+}
+
+func (v *instanceTestState) expectStatsUpdateWithVolumes(t *testing.T,
+	ovsCh <-chan interface{}, volumes []string) bool {
+
+	stats := v.getStatsUpdate(t, ovsCh)
+	if stats == nil {
+		return false
+	}
+
+	if len(volumes) != len(stats.volumes) {
+		t.Errorf("Unxpected number of volumes.  Expected %d found %d",
+			len(volumes), len(stats.volumes))
+	}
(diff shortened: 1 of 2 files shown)


markdryan:
@tpepper  Could you review and merge?  This fixes one of the two remaining storage BAT issues.  The other is #885.

ciao-launcher/instance_test.go
```diff
@@ -209,18 +209,57 @@ func (v *instanceTestState) verifyStatsUpdate(t *testing.T, cmd interface{}) {
...
 	}
 	stats, ok := cmd.(*ovsStatsUpdateCmd)
 	if !ok {
 		t.Error("Unexpected Command received on ovsCh")
 	}
+	return stats
+}
+
+func (v *instanceTestState) expectStatsUpdateWithVolumes(t *testing.T,
+	ovsCh <-chan interface{}, volumes []string) bool {
```
tpepper: Is there an intended usage for the bool return coming?  Looks like all callers do "_ := expectStatsUpdateWithVolumes()".  I'm not sure the code would be notably more succinct with an error return only and no  *testing.T input, but it feels extraneous as is.

markdryan: The bool allows us to detect an error and quit the test case early, shutting down cleanly before we exit.  This is the model used in the similar expectStatsUpdate method and so I duplicated the behaviour when adding the expectStatsUpdateWithVolumes.  However, none of the storage tests do actually quit early if there is an error.  They keep executing and logging any errors they encounter, so as you point out, the return value for expectStatsUpdateWithVolumes isn't currently used.  It may be in the future though, and to be consistent with the expectStatsUpdate method, I'd prefer to keep it.

tpepper: Seems reasonable to me.

PR #877, opened 2016-11-28
ciao-controller unit test improvements

rbradford:
This PR contains a few changes to reduce the flakiness and errors seen from the ciao-controller unit test.

The main change removes the sleeps from the test cases in controller_test.go replacing them with a mechanism that wraps ciao-contoller's client with code that uses channels to signal that commands/errors/events have been received.

PR #888, opened 2016-11-29
testutil: channels are a reference type

tpepper:
Mark Ryan noted in https://github.com/01org/ciao/pull/877 that
channels are reference type.  That means there's no need for the
reference/deference/address-of operators in the result channels in
testutil.


rbradford:
@tpepper I think you mean channels, right?

tpepper:
Read my mind not my commit messages?  Ammending...

PR #891, opened 2016-11-29
ciao-image: Set status to 'Killed' when upload fails

### LINKED ISSUE #663: ciao-image: Failure to upload an image can lead to an invalid state and corrupted image
Take a look at the code in ImageCache.Upload image.  If there is an error writing the uploaded file, we return without reseting the state of the image.  The state will be set to saving, which isn't correct.   It should probably switch back to generic.

Also, in the case where we do have a failure we don't clean up the file, at least we don't in posix.go.  This means that a half written corrupted file will be present in the image service.

obedmr:
In case of failure on image rawDs writting from upload operation,
it will automatically set image status to Killed.

Fixes #663

### DIFF
--- a/ciao-image/datastore/store.go
+++ b/ciao-image/datastore/store.go
@@ -130,16 +130,21 @@ func (s *ImageStore) DeleteImage(ID string) error {
 	s.ImageMap.Lock()
 	defer s.ImageMap.Unlock()
 
-	err := s.rawDs.Delete(ID)
-	if err != nil {
-		return err
-	}
-	err = s.metaDs.Delete(ID)
+	img, err := s.metaDs.Get(ID)
 	if err != nil {
 		return image.ErrNoImage
 	}
 
-	return nil
+	if img.State == Active {
+		err = s.rawDs.Delete(ID)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = s.metaDs.Delete(ID)
+
+	return err
 }
 
 // UploadImage will read an image, save it and update the image cache.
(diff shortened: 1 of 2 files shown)


markdryan:
@obedmr This one needs to be rebased.

obedmr:
@markdryan, it's rebased now.

ciao-image/datastore/store.go
```diff
@@ -159,20 +159,20 @@ func (s *ImageStore) UploadImage(ID string, body io.Reader) error {
 	img.State = Saving
 
 	if s.rawDs != nil {
-		err := s.rawDs.Write(ID, body)
+		err = s.rawDs.Write(ID, body)
 		if err != nil {
-			return err
+			img.State = Killed
+			err = s.rawDs.Delete(ID)
```
markdryan: I don't think the Delete is needed as Write should clean up if there is an error.  Also, as Write has already cleaned up, we're almost guaranteed that Delete will fail and will overwrite the original error from Write.  I suggest we just get rid of this line.

obedmr: @markdryan, agree, good catch, it's applied now. thanks

markdryan:
Maybe we could reword the commit title as well, from

ciao-image: Avoid issues when corrupted image upload 

to

ciao-image: Set status to 'Killed' when upload fails.

markdryan:
@obedmr Could you update the commit title as well?

markdryan:
@obedmr I think we need one small change here.  We should not try to execute

```
	err := s.rawDs.Delete(ID)
	if err != nil {
		return err
	}
```

when deleting an image in the Killed State.  It's currently not possible to delete such an image as the call to err := s.rawDs.Delete(ID) fails as there are no rbd images to delete.  So we could either not call s.rawDs.Delete(ID) when the image state is killed or we could call it an ignore the error, if one occurs, if the image state is killed.

obedmr:
@markdryan, I'm confused, I deleted that line.

obedmr:
@markdryan, agree, I added some changes on ``Delete`` function in order to get status from image, it must have an ``Active`` ``Status`` in order to perform the ``rawDs.Delete``

PR #895, opened 2016-11-29
testutil: Add temporary test cert file generation

bryteise:
In order to run tests, cert files may need to be available outside of
the /etc/pki/ciao folder. This patch makes use of ssntp/certs to
generate certificates for the scheduler, controller, agent, cnci agent,
and net agent roles in a testutil function.

Test components can use testutil.MakeTestCerts() and if no error is
returned be reasonbly confident a valid set of certs can be found in the
paths returned by testutil.RoleToTestCertPath(role) which returns the ca
file path, the cert path for the given role and a potential error. Test
components should cleanup certs with RemoveTestCerts().

For example:

import "github.com/01org/ciao/testutil"

func TestMain(m *testing.M) {
	if err := testutil.MakeTestCerts(); err != nil {
        	// Inform of error/run tests not using certs/etc
        	os.Exit(1)
        }
	result := m.Run()
        testutil.RemoveTestCerts()
        os.Exit(result)
}

---
This change supercedes #827 after discussion about how to best handle test certificates.

### DIFF
--- a/testutil/ssntp_test_certs.go
+++ b/testutil/ssntp_test_certs.go
@@ -16,7 +16,17 @@
 
 package testutil
 
-import "github.com/01org/ciao/ssntp"
+import (
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/01org/ciao/ssntp"
+	"github.com/01org/ciao/ssntp/certs"
+)
 
 // TestCACert is a snake oil Certificate Authority for test automation.
 const TestCACert = `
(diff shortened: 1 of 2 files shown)


testutil/ssntp_test_certs.go
```diff
@@ -486,3 +495,123 @@ func RoleToTestCert(role ssntp.Role) string {
...
+		}
+	}
+
+	return nil
+}
+
+// MakeTestCerts will create test certificate files for all roles
+// that return valid paths from RoleToTestCertPath
+func MakeTestCerts() error {
+	err := makeTestCert(ssntp.SCHEDULER)
```
markdryan: I think it would be nicer if MakeTestCerts called RemoveTestCerts if it encounters an error.  This way clients don't need to do any clean up if the function fails which is probably what they would expect.  Currently, the example presented in the commit message leaks certs if there is an error.

testutil/ssntp_test_certs.go
```diff
@@ -486,3 +495,123 @@ func RoleToTestCert(role ssntp.Role) string {
...
+	template, err := certs.CreateCertTemplate(role, "test", "test@test.test", []string{"localhost"}, []string{"127.0.0.1"})
+	if err != nil {
+		return fmt.Errorf("Unable to create cert template: %v", err)
+	}
+
+	writer, err := os.Create(path)
+	if err != nil {
+		return fmt.Errorf("Unable to create cert file: %v", err)
+	}
+	defer writer.Close()
```
markdryan: We need to check the return value from Close which can be a bit tricky when using defer.

markdryan:
@bryteise Just a few little comments but otherwise looks good to me.  Need approval from @tpepper to merge.

testutil/ssntp_test_certs.go
bryteise: So I want to note that I changed this but it is really cumbersome and fragile needing to have all the cleanup done at every return point. Hopefully I'm missing a go idiom here but without exceptions or goto or automatic cleanup of some sort I don't see a particularly better way to write it. @tpepper I'd love to see thoughts on this part when you review.

testutil/ssntp_test_certs.go
```diff
@@ -486,3 +495,140 @@ func RoleToTestCert(role ssntp.Role) string {
...
+	}
+
+	writer, err := os.Create(path)
+	if err != nil {
+		return fmt.Errorf("Unable to create cert file: %v", err)
+	}
+	if role == ssntp.SCHEDULER {
+		caWriter, err := os.Create(caPath)
+		if err != nil {
+			writer.Close()
```
rbradford: Use a defer with a named error to handle these cases. https://blog.golang.org/defer-panic-and-recover

like https://github.com/01org/ciao/blob/master/ciao-image/datastore/posix.go#L44

bryteise: That's exactly the page I overlooked since it mentioned panic and recover =(. Thanks a bunch for pointing me back at it.

bryteise:
Okay this seems like it is right for review now @tpepper, thanks all for the helpful suggestions!

Issue #373, opened 2016-07-20
Modify compute API to support storage volumes


kaccardi:
Implement the Open Stack compute API changes required to support volume attachments.

kaccardi:
Change existing create server apis to include support for the block-device-mapping-v2 parameter to boot from volume
Change existing list server apis to include support for listing attached volumes
Add support for os-volume_attachements apis (4 of them)

kaccardi:
not done yet - working on moving compute API to new openstack package first.

tpepper:
Marking as unassigned as it's blocked.  When unblocked whoever is free can take this on.

kaccardi:
The block-device-mapping-v2 parameter is specified in the compute api spec as part of the create server api.
http://developer.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server

kaccardi:
this work item is unblocked now.

tpepper:
Discussing with @kaccardi, the four os-volume_attachments APIs (POST, GET, GET {attachment_id}, DELETE) are duplicated versus APIs in Cinder which are already implemented elsewhere in Ciao, so these are probably lower priority / optional.

tpepper:
With the os-volume_attachments APIs on the low priority list, that leaves:

* block-device-mapping-v2 implementation
* list servers implementation

as the priority work in this ticket

tpepper:
As far as I can see, the list servers implementation is currently in order.  Which means the focus here is the block-device-mapping-v2.

tpepper:
Addressed by https://github.com/01org/ciao/pull/896

Issue #387, opened 2016-07-20
Add volume support to ciao-cli


kaccardi:

kaccardi:
This support includes modifying any existing calls to Open Stack compute APIs, adding new calls to Open Stack compute APIs, or any Block API calls that need to be made.

kaccardi:
partially complete. Support for new volume functionality added, however, modification of existing calls to compute api is not done.

tpepper:
When doing instance add, create volume dynamically at the same time.

tpepper:
Addressed by https://github.com/01org/ciao/pull/896

Issue #723, opened 2016-10-20
ciao-cli lacks any unit tests


kaccardi:
ciao-cli doesn't have any unit tests to speak of. Add unit tests and bring coverage up to >70%.

PR #896, opened 2016-11-30
extended volume support

tpepper:
This PR includes changes driven by the desire to support the OpenStack compute API's server creation optional block-device-mapping-v2.  It adds the ability to specify multiple "-volume" arguments on the ciao-cli at instance creation time.  These volumes would typically constitute ancillary data disks beyond the boot volume/image in a formal workload definition, but could also be bootable volumes.  Since workload resources are by definition per-workload, and "ciao-cli instance add" is about instances, this gives the ability to run distinct work instances from the same workload definition base, where the instances of the workload vary in their attached volumes.

Fixes Issue #373 and #387


tpepper:
@markdryan in particular I'd be curious your thoughts on where in launcher I should be looking for insertion points for creating a local volume, accounting its full size against the local Ready frame disk free amount, optionally doing mkswap and swap-on in cloud-init.

markdryan:
@tpepper I need a bit more information about the plans and use cases for local disks and swap devices.

ciao-controller/openstack_compute.go
```diff
@@ -74,6 +77,234 @@ func instanceToServer(ctl *controller, instance *types.Instance) (compute.Server
 	return server, nil
 }
 
+func validateBlockDeviceMappingSourceType(srcType string) (err error) {
+	validSourceTypes := []string{
```
kaccardi: a minor style preference, for return value of error here where you never actually use the variable named "err" in the function, I would prefer to leave return value unnamed. it is a trivail thing of course, but it does prevent an unneeded allocation and initialization of "err" since you aren't using it, and naming error "err" doesn't actually provide any additional readability information.

ciao-controller/instance.go
```diff
@@ -93,8 +93,7 @@ func (i *instance) Add() error {
 	} else {
```
kaccardi: I agree that there's no need to worry about ssntp protocol bump.

ciao-controller/openstack_compute.go
```diff
@@ -373,6 +373,7 @@ func (c *controller) CreateServer(tenant string, server compute.CreateServerRequ
 	if err != nil {
 		return server, err
 	}
+	volumes := abstractBlockDevices(blockDeviceMappings)
```
kaccardi: I notice that this would make this particular commit not compile if we had to bisect. maybe rearrange if possible to improve bisectability.

ciao-controller/instance.go
```diff
@@ -193,70 +214,27 @@ func getStorage(c *controller, wl *types.Workload, tenant string, instanceID str
...
-			TenantID:    tenant,
-			Name:        fmt.Sprintf("Storage for instance: %s", instanceID),
-		}
-
-		err = c.ds.AddBlockDevice(data)
-		if err != nil {
-			c.DeleteBlockDevice(device.ID)
-			return payloads.StorageResource{}, err
-		}
+		s.Persistent = true
```
kaccardi: when you modify s you are actually modifying the cached workload since this is by reference, thus preserving your changes. for the next user of this workload. Is that ok?

tpepper:
Personally I question the need for swap, but I expect that some pet machines/workloads are set up to expect it and even rely on it for a safety net relative to poorly provisioned memory amounts, or for error recovery/diagnosis by recording a crash dump to swap.  Given that it seems reasonable that the OpenStack API would support it.  I see no reason to require it be backed by a volume service volume, versus by a local volume.

For local ephemeral disk, this is a standard industry provided performance up-sell feature, which seems reasonable to support.  Even if we don't support it initially, we're liable to have demand for it.

With respect to implementation, I'm imagining a launcher initiated flow of fallocate in /var/lib/ciao/ a file of the requested size, which would immediately/automatically cause a decrease in reported disk available in Ready frames, map the drive file in qemu's launch options, and then optionally when swap flag is set insert a mkswap/swapon /dev/vdXX into the cloud-init boot command list.

The "swap" flag is an indication to format as swap according to the API as opposed to an indication to simply attach as swap.  This means an mkswap can happen in cloud-init for both local and volume service backed swap devices.  There's arguably a use-case for persistent swap too, but I expect a user would just treat that as a normal volume and do their own swapon in their pet VM, because the "swap" flag would mean format it and that would blow away anything they'd intentionally left in their swap device.

ciao-cli/instance.go
```diff
@@ -69,11 +71,251 @@ var instanceCommand = &command{
...
+	msg += "\t-volume uuid=${UUID},swap\n"
+	msg += "\t-volume uuid=${UUID},ephemeral[,boot_index=N]\n"
+	msg += "\t-volume size=${SIZE}\n"
+	msg += "\t-volume ephemeral,size=${SIZE}\n"
+	msg += "\t-volume local,ephemeral,size=${SIZE}\n"
+	msg += "\t-volume swap,size=${SIZE}\n"
+	msg += "\t-volume local,swap,size=${SIZE}\n"
+	msg += "\n"
+	msg += "Multiple -volume arguments may be specified per workload instance.\n\n"
+	fmt.Fprintf(os.Stderr, msg)
```
kaccardi: you could use the raw string literal here.

tpepper:
Rebased after External IP merge as it had a tiny conflict in one patch.

ciao-controller/openstack_compute.go
tpepper: We're conflicted stylistically through out the code on this topic.  I tend to leave them unnamed, I think?  But I also tend to copy/paste.

I've also spotted some code that does stuff like a dozen instances of exactly "return varA, varB, err", with the var's being inconsistently set/unset and err being something or nil.  I like an explicit "return nil, nil, err" or "return varA, varB, nil" distinguishing quickly visually success and error paths, but we don't consistently do that either.

I'm somewhat surprised we don't get dinged by 'go vet' on this.

@markdryan do you have any idiomatic guidance in this area?

ciao-controller/instance.go
tpepper: That's a very good question.  types.StorageResource's Persistent flag comment is unclear on how the field is used.  If it is set to false, would the storage resource be deleted after completion of an instance of the workload?  If so, given this is in the case for an Image (ie: not a volume), it would seem dangerous to have the image not be persistent and automatically get deleted when a workload depends on it.

markdryan:
@tpepper 

> We're conflicted stylistically through out the code on this topic. I tend to leave them unnamed, I think? But I also tend to copy/paste.

The guidance is here

https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

I tend to use named parameters to override the return value in a defer, when closing a file for example, which is permitted by the guidelines.

tpepper:
Rebased.

ciao-controller/instance.go
```diff
@@ -153,12 +164,32 @@ func (c *config) GetResources() map[string]int {
...
+		Description: s.Tag,
+	}
+
+	err := c.ds.AddBlockDevice(data)
+	if err != nil {
+		c.DeleteBlockDevice(device.ID)
+		return payloads.StorageResource{}, err
+	}
+
+	return payloads.StorageResource{ID: data.ID, Bootable: s.Bootable, Ephemeral: s.Persistent}, nil
```
rbradford: `Ephemeral: s.Persistent`

looks wrong. Is there a logical not missing somewhere?

ciao-controller/instance.go
```diff
@@ -191,80 +222,51 @@ func getStorage(c *controller, wl *types.Workload, tenant string, instanceID str
...
 
-		return payloads.StorageResources{ID: data.ID, Bootable: false}, nil
+	return payloads.StorageResource{}, errors.New("Unsupported workload storage variant in getStorage()")
+}
 
-	}
+func controllerStorageResourceFromPayload(volume payloads.StorageResource) (s types.StorageResource) {
+	s.ID = volume.ID
+	s.Bootable = volume.Bootable
+	s.Persistent = volume.Ephemeral
```
rbradford: Again here..persistence is the apposite ephemerality.

I think we should just standardise on one name.

ciao-controller/instance.go
```diff
@@ -153,12 +164,32 @@ func (c *config) GetResources() map[string]int {
 	return resources
 }
 
-func getStorage(c *controller, wl *types.Workload, tenant string, instanceID string) (payloads.StorageResources, error) {
-	s := wl.Storage
+func addBlockDevice(c *controller, tenant string, instanceID string, device storage.BlockDevice, s types.StorageResource) (payloads.StorageResource, error) {
```
rbradford: Arrgh..the confusion from having a types.StorageResource and a payloads.Storage resource that are very similar.

ciao-controller/instance.go
tpepper: Yes indeed.  I did that a couple times and I thought I'd gone through and fixed them. Clearly not.

tpepper: For simplicity, I fixed the bugs in my existing patches.

ciao-controller/instance.go
tpepper: I've added a patch to the stack which renames Persistent to Ephemeral.

ciao-controller/controller_test.go
```diff
@@ -1122,7 +1122,7 @@ func TestGetStorageForVolume(t *testing.T) {
 		t.Errorf("bootable flag not correct")
 	}
 
-	if pl.Ephemeral != false {
+	if pl.Ephemeral != true {
```
tpepper: @rbradford why is this one automatically changing its ephemerality?  Is this because it's an autocreated overlay volume created from another volume?

tpepper:
I've removed the "WIP".  The code here is nearly functionally complete.  I can specify one or many volumes on the ciao-cli, they're created if needed, and attached on boot.

There is the open question from @markdryan regarding supporting swap and local storage or not.  @kaccardi what are you thoughts?  See also https://github.com/01org/ciao/pull/896#issuecomment-263955057 for some existing discussion.

I still need to add some unit tests for my new code, although coveralls says I my changes didn't cross the threshold boundary to be blocking without additional test coverage.  This appears to be because the two areas in which I added the most code are ciao-cli and ciao-storage, both of which are 100% uncovered and are not even included in coveralls' statistics.  If I add tests for my code additions and add those two packages to coveralls, the overall coverage will go down dramatically, which will require me to write a loooot of tests for existing ciao-cli resolving issue #723 and ciao-storage resolving the newly opened issue #940.  I'm open to the team's suggestions on whether those should block this PR or not.

kaccardi:
@tpepper in my opinion we should defer supporting swap and local storage for now. It's possible that someone will want these things, but I think we should approach incrementally and tackle later when our basic functionality is a little bit more robust.

payloads/start.go
```diff
@@ -86,17 +86,35 @@ const (
 	Docker = "docker"
 )
 
-// StorageResources represents a requested storage resource for a workload.
-type StorageResources struct {
+// StorageResource represents a requested storage resource for a workload.
```
kaccardi: I agree with @rbradford that we should name this something else to avoid confusion in controller between payloads.StorageResource and types.StorageResource.

tpepper:
I modified startWorkload to reduce its inputs by one, having added one.  I don't believe there is an existing struct that makes sense to use for startWorkload() so I'd need to add a struct definition, and with the data not existing in an instantiated object, we have to create an instance of the new struct and we only use it just at the point of passing data into the startWorkload() function.  We'd move from 9 single long lines of startWorkload() with five arguments, to 9 sets of, eg:

Code that is currently:
```
_, err = c.startWorkload(workloadID, tenantID, 1, "", noVolumes)
```

Would become:
```
w := types.SomethingWorkload{
        workloadID: workloadID
        tenantID: tenantID
        instances: 1
        traceLabel: ""
        volumes: noVolumes
}
_, err = c.startWorkload(w)
```

I'm not convinced that is better.

tpepper:
I would prefer to defer changing struct names.  There are a dozen semi-overlapping structs related to storage because ciao-storage, ciao-image, payloads, launcher, scheduler, controller, workload definnition, openstack compatibility interface, etc. all have slightly different parts of storage they care about and image and volume storage are only semi-intertwined.  Renaming one of these structs does little to reduce the complexity and confusion.

I think we could reduce this down to a single storage resource struct, some of the fields of which are ignored by some of the parts of the stack, but which covers both volume and image needs as those ultimately are all the same abstract object...an "rbd volume"...in our integrated ceph-is-good-enough approach.

kaccardi:
@tpepper I prefer the modification to use the request struct. The reason is that as we move more items into startWorkload in the future, we can just add onto the struct and have a default value that is set - then you don't have to modify every call to startWorkload just because you decided you needed a new argument. That said, if you don't want to do it I can certainly do it myself later, so I won't block merge.

tpepper:
If the change to startWorkload() is categorically going to happen, it may as well happen here as requested.

tpepper:
I've gone back to WIP as I'm dabbling at covering my code additions in test and I trigger the coveralls barrier for excessive reduction in coverage by enabling test on ciao-storage and ciao-cli (previously 100% uncovered from coveralls perspective).

ciao-launcher/payload.go
```diff
@@ -168,11 +170,13 @@ func parseStartPayload(data []byte) (*vmConfig, *payloadError) {
 	vnicIP := strings.TrimSpace(net.PrivateIP)
 	sshPort := computeSSHPort(networkNode, vnicIP)
 	var volumes []volumeConfig
-	if start.Storage.ID != "" {
-		volumes = append(volumes, volumeConfig{
-			UUID:     start.Storage.ID,
-			Bootable: start.Storage.Bootable,
-		})
+	for _, storage := range start.Storage {
```
markdryan: We should return an error if storage.ID == "".  This check was needed in the past as we had a string and not an array so we needed to compare to the empty string to see whether a volume was specified.  But now we have a slice, the fact that there is an element in the slice means that someone has specified a volume.

ciao-launcher/vmconfig.go
```diff
@@ -96,6 +96,14 @@ func (cfg *vmConfig) findVolume(UUID string) *volumeConfig {
 	}
 	return nil
 }
+func (cfg *vmConfig) haveBootableVolume() bool {
```
markdryan: The bootable volume must be the first volume in the volume list.  We cannot really sort the volumes to ensure this as doing so would change the names that the volumes are given in the guest.  Maybe there's some nice way of assigning names in volumes going forward, but this is currently the way things are.  So this function should be

```
func (cfg *vmConfig) haveBootableVolume() bool {
  return len(cfg.Volumes) > 0 && cfg.Volumes[0].Bootable
}
```

markdryan: As things stand right now, if you specify two volumes in the workload template of which the second is bootable but the first is not, I don't think the instance will actually boot.  This is a launcher restriction and we need to enforce it here.

ciao-cli/instance_test.go
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2016 Intel Corporation
```
markdryan: I think we could arrange for these tests to be run in travis without affecting our coverage statistics.  We could just try adding a new invocation of test-cases that leaves off the coverage:

test-cases -v -timeout 9  github.com/01org/ciao/ciao-cli

ciao-launcher/vmconfig.go
tpepper: It does work as is.  But yes the ordering varies depending on how things are defined.  I can define things in a way that there is or is not a scratch volume on the instance add command line.  If the scratch is first it comes up as /dev/vda in the guest and the image is /dev/vdb and the guest still boots.  Without the scratch volume the image is /dev/vda.  Or I can do no image in the workload_storage.csv and mix around the ordering of -volume on the cli and move the disks around arbitrarily.  To me this mimics real life disk discovery in a computer.  I can see a rationale for enforcing a specific ordering and that seems to be what boot_index is about.  Launcher would need to look at the list of volumes any optional boot_indexes and hand the volumes to qemu in the index order to get them to come up with names in that order.

ciao-launcher/payload.go
tpepper: storage.ID == "" would be the case for a "local" volume, which launcher would auto-create/destroy, but we've discussed not supporting that.  I'll add a check for that and open an issue for tracking eventual support.

tpepper: I added the error return and opened issue #972.

ciao-launcher/vmconfig.go
mcastelino: @tpepper @markdryan I actually tried it an it seems to work
`ciao-cli instance add -workload=e35ed972-c46c-4aad-a1e7-ef103ae079a2 -volume uuid=b3d391e9-309d-4271-b76e-1767c44877b8,boot_index=0`
Here the volume was a previously auto cloned volume from another instance.

The only thing that bothers me with this is that we end up with an extra boot volume (the one based off of the image).
This is a bit confusing as, when you boot from a volume, the image definition is immaterial. The bootable volume is the image.

See this
qemu-system-x86_64 
-drive file=rbd:rbd/b3d391e9-309d-4271-b76e-1767c44877b8:id=ciao,if=none,id=drive_b3d391e9-309d-4271-b76e-1767c44877b8,format=raw 
-device virtio-blk-pci,scsi=off,bus=pci.0,addr=0x3,id=device_b3d391e9-309d-4271-b76e-1767c44877b8,drive=drive_b3d391e9-309d-4271-b76e-1767c44877b8 

-drive file=rbd:rbd/2c1bf8a6-6bb6-4201-b463-acaa683d4bdf:id=ciao,if=none,id=drive_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf,format=raw 
-device virtio-blk-pci,scsi=off,bus=pci.0,addr=0x4,id=device_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf,drive=drive_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf 

-drive file=/var/lib/ciao/instances/29ea470d-96f6-4689-afd3-ec120d012c4b/seed.iso,if=virtio,media=cdrom

PR #904, opened 2016-11-30
ansible: Upload images to ciao-image

### LINKED ISSUE #841: Ansible needs to upload image files
The ansible scripts need to upload the images that are used in the workload templates via "ciao-cli image new".

This is required as part of moving the images from the filesystem to ceph: #811

albertomurillo:
First commit Fix #841
The rest are improvements

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/defaults/main.yml
+++ b/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/defaults/main.yml
@@ -83,3 +83,18 @@ fedora_cloud_image_url: https://download.fedoraproject.org/pub/fedora/linux/rele
 
 # Fedora cloud image file name
 fedora_cloud_image: "{{ fedora_cloud_image_url.split('/') | last }}"
+
+# URL for the latest ciao networking image
+cnci_image_url: https://download.clearlinux.org/demos/ciao/clear-8260-ciao-networking.img.xz
+
+# Ciao networking image file name
+cnci_image: "{{ cnci_image_url[0:-3].split('/') | last }}"
+
+# URL for the latest clearlinux cloud image
+clear_cloud_image_url: https://download.clearlinux.org/releases/11960/clear/clear-11960-cloud.img.xz
+
+# Clearlinux cloud image file name
+clear_cloud_image: "{{ clear_cloud_image_url.split('/') | last | regex_replace('(.*).xz', '\\1') }}"
+
+# URL for EFI Firmware required by CNCI Image
+ovmf_url: https://download.clearlinux.org/image/OVMF.fd
--- a/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/defaults/main.yml
+++ b/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/defaults/main.yml
@@ -77,3 +77,9 @@ ceph_id: admin
 
 # Path to ceph user keyring
 secret_path: /etc/ceph/ceph.client.admin.keyring
+
+# URL for the latest fedora cloud image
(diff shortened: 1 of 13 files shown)


kaccardi:
@albertomurillo is there a specific github issue this PR is addressing? If so, would you mind editing the description to add a reference to it?

albertomurillo:
The first commit fixes 841

albertomurillo:
This is what singlevm does

albertomurillo:
Also any enhancement to the handling of the certificate should be tracked at another issue. The certificates are already added to the trust store on the controller node.

7e7103c adds the certificate to the trust store on the deployment node when the deployment
node and the controller node are different (which should always be) but we were running ansible
in the controller node due to lack of HW.

rbradford:
@albertomurillo CIAO_CA_CERT_FILE environment's sole purpose is to avoid having to install certificates systemwide for ciao-cli to be able to access a cluster.

If you *now* have to install then that functionality has broken.

If you have tried only using that environment variable then i'm happy to remove my objection if you open a bug saying that this environment variable is not sufficient to access a cluster.

albertomurillo:
@rbradford Thanks for the hint. I'll investigate further and fix this PR, or open another issue otherwise.

albertomurillo:
Commit to install certificates on the global trust store on the deployment node has been dropped.
upload_image.yml now uses CIAO_CA_CERT_FILE instead

rbradford:
@albertomurillo  scratch my last comment - I misread. This is good as it is.

tpepper:
Looks like commit ID's changed with revisions and I can't tell for sure which commit ordering was desired or if @albertomurillo achieved that.  @rbradford can you confirm your desired change was done?

albertomurillo:
I agree the name is not fully descriptive but it is indeed the keystone certificate required to make request to keystone via ssl.

https://github.com/01org/ciao/blob/master/testutil/singlevm/setup.sh#L145
https://github.com/01org/ciao/blob/master/testutil/singlevm/setup.sh#L80

rbradford:
@tpepper i'm cool with this ordering.

rbradford:
@tpepper The cert is the https self-signed certificate (and so it also the CA) for both keystone and controller (https).

If you were in environment where keystone and controller had different certs (they were on different hosts for example) but with a common, non-statndard CA, e.g. a corporate PKI one, then this variable also covers that. Hence the CA naming.

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/defaults/main.yml
```diff
@@ -77,3 +77,24 @@ ceph_id: admin
...
+fedora_cloud_image: "{{ fedora_cloud_image_url.split('/') | last }}"
+
+# URL for the latest ciao networking image
+cnci_image_url: https://download.clearlinux.org/demos/ciao/clear-8260-ciao-networking.img.xz
+
+# Ciao networking image file name
+cnci_image: "{{ cnci_image_url[0:-3].split('/') | last }}"
+
+# URL for the latest clearlinux cloud image
+clear_cloud_image_url: https://download.clearlinux.org/releases/11960/clear/clear-11960-cloud.img.xz
```
obedmr: Are we planning to update this on every ClearLinux Release, because when a new ClearLinux release is out, this link will be broken.

obedmr: oh no, sorry, I missread the full link, it's getting from ``...releases/<release>``

obedmr: That link will not get broken, the question would be, how often we'll be updating this URL?

albertomurillo: We need a new issue to implement a feature to download the latest image

PR #906, opened 2016-11-30
ciao-image: Fix raw datastore error handling on ciao-image service

### LINKED ISSUE #905: ciao-storage, ciao-image: RBD commands hang and then hang ciao-image service calls
In order to reproduce this issue you need:

1. A CIAO cluster running with Ceph as storage backend
2. Create 1 image with ciao-cli
```
ciao-cli image add -name TEST_IMAGE -file TEST_IMAGE.img -id df3768da-31f5-4ba6-82f0-127a1a705169
```
3. Stop Ceph cluster
4. Try to run an image delete
```
ciao-cli image delete -image df3768da-31f5-4ba6-82f0-127a1a705169
```

You will notice that last command will hang forever, not sure how long, I waited for more than 1 minute. If you try to run any other ``ciao-cli image`` subcommand, they will be hanged because there's a lock on ``ciao-image`` service from last command.

obedmr:
This commit fixes the loss of ceph error when an image upload operation
is performed. Now, Ceph errors are properly returned and Metadata errors
are separately handled.

Fixes #905

### DIFF
--- a/ciao-image/datastore/datastore_test.go
+++ b/ciao-image/datastore/datastore_test.go
@@ -15,15 +15,23 @@
 package datastore
 
 import (
+	"os"
+	"path"
 	"strings"
 	"testing"
 
 	"github.com/01org/ciao/database"
 )
 
+var mountPoint = "/tmp"
+var metaDsTables = []string{"images"}
+var dbDir = "/tmp"
+var dbFile = "ciao-image.db"
+var testImageID = "12345678-1234-5678-1234-567812345678"
+
 func testCreateAndGet(t *testing.T, d RawDataStore, m MetaDataStore) {
 	i := Image{
-		ID:    "validID",
+		ID:    testImageID,
 		State: Created,
 	}
 
@@ -49,7 +57,7 @@ func testCreateAndGet(t *testing.T, d RawDataStore, m MetaDataStore) {
 
 func testGetAll(t *testing.T, d RawDataStore, m MetaDataStore) {
 	i := Image{
-		ID:    "validID",
+		ID:    testImageID,
 		State: Created,
 	}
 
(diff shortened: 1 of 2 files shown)


obedmr:
Based on our IRC discussion, below some points:

- There's a default timeout ``300 seconds``
- It seems like that timeout is for connecting with the Ceph cluster, once connected, file uploads may depend on size.

Considering ``300 seconds`` as a reasonable timeout, this PR will make sense anymore. Is that ok for you @markdryan  and @kaccardi ?

markdryan:
@obedmr We probably want to fix this in the ceph configuration as discussed on IRC.  I'd still like to perform one experiment though.  Set the rados_mon_op_timeout to a really low value say one second and then perform an rbd task that takes longer than one second, such as an import of a large file of a qemu-img convert.  I'd like to make sure that this tasks don't time out.

obedmr:
@markdryan sure, I'll test/play with ``rados_mon_op_time`` configuration and will let you know the results

obedmr:
@markdryan after some research, I found what's the option for timeout with ``rbd`` commands
```
[client]
client mount timeout = 1
```
I set it to 1 second, then I tried with a big image and it worked. That 1 sec timeout is just for the auth/connection, once it got the authentication confirmation, it starts sending the file without a problem.

This timeout was really useful when a Ceph cluster is down, I didn't have to wait the 300 seconds to receive a timeout message. It failed faster and let me know the issue.

Now, we can leave it in the ceph.conf hands, for the case of images creation, it will create the image in metadata store and when the ceph commands run they will fail and image's status will be marked as killed (which means an error on image upload occured).

I modified the commit for a proper raw datastore errors handling, we'll mark image's status as killed and then will return the raw datastore error in order to return a Server Error to the HTTP client.

obedmr:
@markdryan, thanks for the feedback. I made the rewording, thanks.

PR #908, opened 2016-12-01
Single VM: Fix Single VM

markdryan:
Single VM was broken by #850 as it was not updated to reflect the new
controller paths introduced by that PR.  This patch modifies Single VM
to use these paths.  The BAT tests pass in Single VM with this patch applied.


markdryan:
@mcastelino I'm just going to merge this as I can't really do any work while Single VM is broken.  Maybe you could take a look later just to make sure I didn't mess up.

mcastelino:
@markdryan this seems to be still broken. The ssh keys are not being setup correctly...

PR #909, opened 2016-12-01
Persistent store cleanup

rbradford:
Make a few cleanups to persistentStore after #850: 

a.) delete some unnecessary directory creations.
b.) treat the URI member as a URI


obedmr:
LGTM, those paths are created by ciao deployment, so, they should be there before running the ``ciao-controller`` service.  Sorry for the ``spurious`` fmt.Println

Issue #822, opened 2016-11-14
document required host/role firewall rules


tpepper:
If deployed on hosts with restrictive firewall rules, Ciao is unable to operate.  The Ciao deployment documentation needs to express what ports/protocols are needed per Ciao server host/role type (eg: identity, storage, controller, scheduler, compute node, network node, etc.).

@mcastelino has mentioned that at least the networking node is likely going to need its iptables fully delegated to Ciao.

This also impacts singlemachine/singleVM.

Especially on Fedora the default firewall rules are liable to be restrictive.  But also on a well run cluster, IT security is likely to desire strong controls with only well established ports/protocols exposed.

mcastelino:
@tpepper to expand on this, on CN and NN's the Firewall rules have to be setup such that
- All tenant expressed per workload restrictions are enforced
- All Data Center required firewall rules are enforced
In case of conflicts between Data Center firewall rules and Tenant firewall rules, the data center rules should take precedence. It will be hard to have two agents managing the firewall rules in the case of CN and NN.

On the control nodes, we can support allowing the data center operator to setup the system with any firewall settings and ciao can add ciao specific exceptions.

obedmr:
@erick0z could you take a look on this.

PR #927, opened 2016-12-02
[RFC] ansible: Add Firewall documentation

erick0z:
User/Operator must be aware of the firewall configuration
when deploying a Ciao cluster. This commit adds a reference
of the firewall configuration that must be done in order to have
an operational Ciao cluster.

Fixes https://github.com/01org/ciao/issues/822

With this settings, running an all-fedora-24 Ciao cluster, I'm able to launch instances, reach them via SSH ~~but not able to get instances reach internet(WIP, maybe proxies)~~ and able to reach the Internet within the instances.


mcastelino:
@erick0z You can only reach the internet from workloads if the CNCI is on a fully route able network from which it can itself reach the internet. If the Ciao network is in an isolated network. Can the CNCI in your setup directly reach the internet?

erick0z:
@mcastelino Thanks! I'm currently able to reach the internet from the workload instances(...proxy issues). I will recreate the environment to make sure the rules are valid, because I was adding test rules.

Issue #919, opened 2016-12-02
osprepare should not import glog


markdryan:
Osprepare includes a logging abstraction that allows clients to specify how and where they would like osprepare's logs written.  It also includes an implementation of this abstraction based on glog.  The problem is that as soon as you import glog it adds the glog flags to your program.  So clients of osprepare autoamitcally get the glog flags added to their usage messages even if their implementation of osprepares logging abstraction doesn't use glog, which is irritating.

Osprepare's logging interface and abstraction need to be moved to a separate package.  So that they can be used by both osprepare and other packages, e.g, qemu.

tpepper:
@bryteise was this something you were going to look at?

bryteise:
Oh no I hadn't looked at that yet. I'll take a look at it though.

bryteise:
@markdryan So are you looking for osprepare to shed its component specific logging for a generic toplevel (ciao-logger say) logging interface that osprepare would switch to and other components would start to use? Then osprepare would use that interface to add the current glog interface implementation?

markdryan:
@bryteise Yes.  Pretty much.  The idea being that packages which function as libraries (osprepare, ssntp, qemu) wouldn't include glog at all.  They could then be used by programs which don't want or need to use glog.  Right now programs we write that use osprepare inherit the glog flags even if they never use glog which is undesirable.

We may need to put the abstract logging interface and the glog implementation in separate packages so osprepare doesn't import a package that imports glog.  This may mean that it's not possible for osprepare to provide a default glog implementation of its logs.

bryteise:
@markdryan Okay. I agree there would be no default logging implementation outside of the null logger in the new implementation.

PR #930, opened 2016-12-05
Refactor logging interface used in osprepare

bryteise:
Instead of having osprepare attempt to provide all the logging interface
components, add a new ciao module for logging that provides by default
only a no log implementation.

Also add a separate module that provides an interface implementation
that uses glog.

This change is done to avoid any interface implementor getting glog
flags added to their messages (since previously osprepare always
required glog).

Addresses #919.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -37,6 +37,7 @@ import (
 	datastore "github.com/01org/ciao/ciao-controller/internal/datastore"
 	image "github.com/01org/ciao/ciao-image/client"
 	storage "github.com/01org/ciao/ciao-storage"
+	"github.com/01org/ciao/clogger/gloginterface"
 	"github.com/01org/ciao/openstack/block"
 	"github.com/01org/ciao/openstack/compute"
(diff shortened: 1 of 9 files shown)


markdryan:
@bryteise Works fine for me but

1. Could you sign off your commit?
2. Add a README.md file to clogger?

@tpepper What do you think of the high level directory name.  Is clogger okay for you or would you prefer ciao-logger?  I think clogger is probably okay for me.

bryteise:
So I originally had ciao-logger but I didn't really like package name different from folder name and didn't see any use of dashes for non main packages. I also didn't know what to call the package for that case either X(.

tpepper:
I'm ok with clogger.  Would also be fine with "logger".  I've never been a fan of github.org/01org/ciao/ciao-{name} as it feels redundant having the second ciao.

bryteise:
@tpepper I went with clogger since logger is such a nice variable name that I'd hate to take over.

bryteise:
Anyway README.md and and signoff added.

PR #935, opened 2016-12-06
bat: Specify where to run bat tests from

albertomurillo:
Bat test requires access to a ciao cluster and for storage
tests it requires qemu-img and access to the ceph cluster.

This change updates _relase/bat/README.md to

- Instruct the user to run the tests from the controller
- Fix typo in 'wildcard'
- Fix hyperlink to singlevm

### DIFF
--- a/_release/bat/README.md
+++ b/_release/bat/README.md
@@ -23,7 +23,14 @@ existing test case runner, test-cases.
 The BAT tests require a running ciao cluster to execute.  This can be a
 full ciao cluster running on hundreds of nodes or a Single VM ciao cluster
 running on a single machine.  For more information about Single VM see
-(here)[https://github.com/01org/ciao/wiki/Single-Machine-Development-Environment].
+[here](https://github.com/01org/ciao/wiki/Single-Machine-Development-Environment).
+
+The BAT tests have some dependencies. The device on which they are run must have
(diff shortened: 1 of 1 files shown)


albertomurillo:
@kaccardi Can you review this one?

_release/bat/README.md
```diff
@@ -20,10 +20,10 @@ existing test case runner, test-cases.
 
 ## Set up
 
-The BAT tests require a running ciao cluster to execute.  This can be a
+The BAT tests should run from the controller node of a ciao cluster.  This can be a
```
markdryan: As you say in the commit message the real issue is not that the BAT tests need to be run from the controller node.  Rather it's that the controller node has all the pieces in place for the tests to run successfully.

I'd leave the existing text the way it is an add an extra paragraph after the Single VM URL.

> The BAT tests have some dependencies.  The device on which they are run must have qemu-img installed and must also have access to the ceph cluster.  The controller node in a ciao cluster fulfills both of these requirements so it is often easiest to run the BAT tests from this node.  There is only one node in Single VM clusters, so when using Single VM, simply run the BAT tests from the device on which you ran setup.sh.

tpepper: Looks like @albertomurillo 's f663ef4 modification addresses this additional paragraph request so I'm going to merge.

Issue #684, opened 2016-10-14
ciao-cli image list on SingleVM has some empty values


rbradford:
Taken from a run on SingleVM. I would expect values such as size/checksum/status at the least to be filled in.

```
$ ciao-cli image list
Image #1
    Name             [Fedorda Cloud Base 24-1.2]
    Size             [0 bytes]
    UUID             [73a86d7e-93c0-480e-9c41-ab42f69b7799]
    Status           [queued]
    Owner            []
    Disk format      []
    Container format []
    Minimal disk     [0 GB]
    Minimal memory   [0 MB]
    Checksum         []
    Visibility       [public]
    Protected        [false]
    CreatedDate      [2016-10-14T15:58:10.241691884+01:00]
    LastUpdate       []
    Tags             [[]]
    File             [/v2/images/73a86d7e-93c0-480e-9c41-ab42f69b7799/file]
    Schema           [/v2/schemas/image]

```

This might be related to #632 but there might also need to be changes elsewhere.

tpepper:
dumpImage(i *images.Image) in ciao-cli/image.go needs reviewed.  Some of the fields have no meaning in Ciao, versus in OpenStack.  The lines of code that currently fmt.Printf()'ing those fields simply need removed.

Relates also to issue #681 which enables the user to get data into the fields in the first place.

rbradford:
I think we should only show the fields where we have useful data. We can expose 
more if we start to use those fields

Giving the following:
```
     Name             [Fedorda Cloud Base 24-1.2]
     Size                [<Fill in from rbd>]
     UUID              [73a86d7e-93c0-480e-9c41-ab42f69b7799]
     Status            [queued]
     CreatedDate      [2016-10-14T15:58:10.241691884+01:00]
```

Which seems a bit more reasonable.

Issue #681, opened 2016-10-14
ciao-image ignores most of the parameters passed to by ciao-cli image add


markdryan:
If you look at the help for ciao-cli image add you will see that it takes a whole pile of parameters.  However, the majority of these are simply ignored by the image service.  We should do something with them.

markdryan:
Of all the fields in the image.CreateImageRequest

```
Name            string
ID              string
Visibility      Visibility
Tags            []string
ContainerFormat ContainerFormat
DiskFormat      DiskFormat
MinDisk         int
MinRAM          int
Protected       bool
Properties      interface{}
```

Only Name and ID are used by ciao-image.

rbradford:
What parameters are we going to remove?

obedmr:
@rbradford if we want compatibility with OpenStack image service (Glance), I'd recommend to keep all of them and set some default values for those we're not using yet. Those unused parameter could be useful in the future. What do you think @rbradford  @kaccardi @markdryan

rbradford:
We can still maintain compatibility with openstack tools but we don't need to expose them in our tools.

obedmr:
@rbradford, which ones you're proposing to remove?

obedmr:
Assigning to @rbradford who is actually doing the PR.

PR #936, opened 2016-12-06
ciao-cli image cleanup

rbradford:
This PR addresses #684 and #681 by removing unused functionality from ciao-cli image.

### DIFF
--- a/ciao-cli/image.go
+++ b/ciao-cli/image.go
@@ -36,17 +36,8 @@ const imageTemplateDesc = `struct {
 	SizeBytes        int      // Size of image in bytes
 	ID               string   // Image UUID
 	Status           string   // Image status.  Can be queued or active
-	Owner            string   // Tenant that owns the image
-	ContainerFormat  string   // Format of the container, e.g., ovf
-	DiskFormat       string   // Format of the image, e.g., qcow2
-	MinDiskGigabytes int      // Minimum amount of disk space required to boot image
-	MinRAMMegabytes  int      // Minimum amount of RAM required to boot image
-	Checksum         string   // Checksum of image data
-	Protected        bool     // Indicates whether or not an image can be deleted
-	Visibility       string   // Indicates whether the image is public or private
 	CreatedDate      string   // Image creation date
 	LastUpdate       string   // Timestamp of last update
-	Tags             []string // List of image tags
 	File             string   // Image path
 	Schema           string   // Path to json schema
 }`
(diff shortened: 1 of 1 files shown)


ciao-cli/image.go
```diff
@@ -36,17 +35,8 @@ const imageTemplateDesc = `struct {
...
 	ID               string   // Image UUID
 	Status           string   // Image status.  Can be queued or active
-	Owner            string   // Tenant that owns the image
-	ContainerFormat  string   // Format of the container, e.g., ovf
-	DiskFormat       string   // Format of the image, e.g., qcow2
-	MinDiskGigabytes int      // Minimum amount of disk space required to boot image
-	MinRAMMegabytes  int      // Minimum amount of RAM required to boot image
-	Checksum         string   // Checksum of image data
-	Protected        bool     // Indicates whether or not an image can be deleted
-	Visibility       string   // Indicates whether the image is public or private
```
obedmr: Visibility is useful to have information about which images are private (self created images) and public (all-tenants available) images.

ciao-cli/image.go
```diff
@@ -97,18 +87,11 @@ var imageCommand = &command{
 }
 
 type imageAddCommand struct {
-	Flag            flag.FlagSet
-	name            string
-	id              string
-	containerFormat imageContainerFormat
-	diskFormat      imageDiskFormat
-	minDiskSize     int
```
obedmr: When workloads CRUD capabilities are available, it is really useful to have Disk and Memory minimal values for Workloads launching.

ciao-cli/image.go
rbradford: When the functionality is added to handle public vs private then this can be trivially re-added. However right now it serves no purpose

ciao-cli/image.go
rbradford: That should be on the workloads not the image.

ciao-cli/image.go
obedmr: Ok, got it, good point. agree with that.

ciao-cli/image.go
tpepper: Image implies read-only.  Once uploaded, we know the actual size of the image and done.  An image minimum size seems unuseful to me.

obedmr: it's not only disk size, what about memory, you need to specifiy which is the minimal memory a OS image needs for running. Then, workloads can start from that minimal required ram.

obedmr: And it's not image size, it's the required disk for the OS to run. A VM with only 100M of available disk would not be really useful in real worlds scenarios.

obedmr:
@markdryan, yes, I have not started that. But it shouldn't take that long. I'd expect 1 week as maximum. Let me start Today with this and will send my PR with the progress I have in order to start receiving feedback about it.

markdryan:
@obedmr So there's a reasonable chance that your meta data improvements won't get merged in this Sprint.  In this case I agree with @rbradford.  Let's remove the unused stuff for now and add it back in as needed.  As soon as the BAT tests are fixed we can merge.

obedmr:
@markdryan yeah, that's ok, agree too. There's still an open discussion about which fields we should keep at the end. So, let's just print what is reasonable now.

ciao-cli/image.go
kaccardi: @obedmr I certainly can see an argument for including this as you mentioned.  However, in ciao, launcher will bump the memory and disk size automatically if the workload doesn't specify a large enough memory/disk for the image. So right now controller doesn't use the image mem/disk size at all, but maybe @markdryan might wish it did.

Issue #566, opened 2016-09-15
Return correct error response from upload image


kaccardi:
Properly handle ErrImageSaving return value from upload image - return appropriate 4xx response as indicated by image api spec.

PR #937, opened 2016-12-06
ciao-image: Add regex pattern for validating UUID

### LINKED ISSUE #563: Add input validation for uuids: image service
Modify the gorilla mux routes to require uuid4() for any uuid input (using regex in path). For example:
r.Path("/articles/{category}/{id:[0-9]+}").
Handler(ArticleHandler)

obedmr:
It implements a regex validator for UUIDs in ciao-image service HTTP requests. It will respond with a ``Not Found`` (404) HTTP Error for all those requests that contain a malformed UUID.

It also fixes an issue with images getting from meta datastore. Errors handling needs to return a valid error when we have diferent than ``Not Found`` errors.

Fixes #563 and #566

### DIFF
--- a/openstack/image/api.go
+++ b/openstack/image/api.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/01org/ciao/ssntp/uuid"
 	"github.com/gorilla/mux"
 )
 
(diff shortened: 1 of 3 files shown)


openstack/image/api.go
```diff
@@ -409,10 +413,10 @@ func Routes(config APIConfig) *mux.Router {
 	// API versions
 	r.Handle("/", APIHandler{context, listAPIVersions}).Methods("GET")
 	r.Handle("/v2/images", APIHandler{context, createImage}).Methods("POST")
-	r.Handle("/v2/images/{image_id}/file", APIHandler{context, uploadImage}).Methods("PUT")
+	r.Handle("/v2/images/"+UUIDRegex+"}/file", APIHandler{context, uploadImage}).Methods("PUT")
 	r.Handle("/v2/images", APIHandler{context, listImages}).Methods("GET")
-	r.Handle("/v2/images/{image_id}", APIHandler{context, getImage}).Methods("GET")
-	r.Handle("/v2/images/{image_id}", APIHandler{context, deleteImage}).Methods("DELETE")
+	r.Handle("/v2/images/"+UUIDRegex+"}", APIHandler{context, getImage}).Methods("GET")
```
markdryan: These first two don't look correct to me.  Shouldn't there be a "{image_id :"?

openstack/image/api.go
```diff
@@ -393,6 +396,7 @@ func deleteImage(context *Context, w http.ResponseWriter, r *http.Request) (APIR
 	imageID := vars["image_id"]
 
 	_, err := context.DeleteImage(imageID)
+	fmt.Println(err)
```
markdryan: We don't need this print statement.

obedmr: @markdryan, yeah, you're right, I need to check for println before submitting the patch. Thanks fot the catch.

openstack/image/api.go
obedmr: @markdryan, yes, you're right, /me fixing it.

obedmr:
@kaccardi, sure, change applied.

markdryan:
Agree with @tpepper Let's put the regexp in ssntp/uuid.go

obedmr:
@tpepper @markdryan @kaccardi  changes applied. Thanks for the feedback.

Issue #642, opened 2016-10-06
ciao-storage: ceph: improve the error reporting when creating volumes


kaccardi:
Use more helpful error messages when volume create fails.

markdryan:
We need to capture the output of the commands used to create ceph volumes and either log or report them or both.

PR #939, opened 2016-12-06
ciao-storage: Improve error logging from rbd commands

rbradford:
Report the output from rbd commands in the log if the command errors
out (non-zero exit code.)

Fixes: #642

### DIFF
--- a/ciao-storage/ceph.go
+++ b/ciao-storage/ceph.go
@@ -55,9 +55,9 @@ func (d CephDriver) CreateBlockDevice(volumeUUID string, imagePath string, size 
 		cmd = exec.Command("rbd", "--id", d.ID, "--image-feature", "layering", "create", "--size", strconv.Itoa(size)+"G", volumeUUID)
 	}
 
-	err := cmd.Run()
+	out, err := cmd.CombinedOutput()
 	if err != nil {
-		return BlockDevice{}, fmt.Errorf("Error when running: %v: %v", cmd.Args, err)
+		return BlockDevice{}, fmt.Errorf("Error when running: %v: %v: %s", cmd.Args, err, out)
 	}
 
 	return BlockDevice{ID: volumeUUID}, nil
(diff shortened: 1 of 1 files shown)


markdryan:
@tpepper This is generally the case but not universally true.  The exec functions like Output and CombinedOutput can return non-nil slices for their first return value even if their second return value is a non-nil error.  There are other examples in the standard library such as io.Reader.Read.

rbradford:
@tpepper I did check this by forcing some failure paths .e.g. by trying to delete the CNCI image which rbd wouldn't let you do because it's actively in use.

In those cases I got an !nil err as well as output.

ciao-storage/ceph.go
```diff
@@ -166,16 +170,22 @@ func (d CephDriver) MapVolumeToNode(volumeUUID string) (string, error) {
...
+	}
+	return nil
 }
 
 // GetVolumeMapping returns a map of volumeUUID to mapped devices.
 func (d CephDriver) GetVolumeMapping() (map[string][]string, error) {
 	args := append(d.getCredentials(), "showmapped", "--format", "json")
 	cmd := exec.Command("rbd", args...)
-	data, err := cmd.Output()
+	data, err := cmd.CombinedOutput()
```
markdryan: This is changing the behaviour of the function.  The old version was reading the first line of Stdout.  The new code now reads the first line of combined Stdout and Stderr which might not be the same thing.  Probably best to continue to use Output, but to retrieve the error information by

err.(*ExitError).Stderr

rbradford: Ah @markdryan  that's a nice solution. I'd have hoped that rbd did the right right of not writing random stuff to stderr with a zero exit code. But this fix is the safest.

ciao-storage/ceph.go
```diff
@@ -175,7 +185,7 @@ func (d CephDriver) GetVolumeMapping() (map[string][]string, error) {
 	cmd := exec.Command("rbd", args...)
 	data, err := cmd.Output()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Error when running: %v: %v: %s", cmd.Args, err, err.(*exec.ExitError).Stderr)
```
markdryan: @rbradford Sorry this is my fault, but reading the docs again the error is not guaranteed to be of type ExitError.  If it's not the cast will panic.

> Any returned error will usually be of type *ExitError. If c.Stderr was nil, Output populates ExitError.Stderr. 

So we should probably write
```
var output []byte
if err, ok := err.(*exec.ExitError); ok {
  output = err.Stderr
}
```

to be safe.

PR #941, opened 2016-12-07
networking: libsnnet: Export bridge API

sboeuf:
Methods from bridge.go can be reused by some other projects such as
virtcontainers, that's the reason why we export them to make them
accessible.


mcastelino:
LGTM

Verified with ciao-down on FC24

@tpepper good to merge once travis CI agrees.

Issue #959, opened 2016-12-08
Image sizes are not populated from rbd


rbradford:
Need to query the size from rbd and make sure that the datastore is populated with the appropriate sizes and then send that to the client.

rbradford:
(Updated to only refer to images, volume size mostly works.)

PR #960, opened 2016-12-08
Populate the image size based off of ceph

rbradford:
Query the image size from rbd and populate that into the datastore. From the datastore populate the openstack image API datastructures.

Fixes: #959


rbradford:
@kaccardi I think you're the owner for this one.

obedmr:
@rbradford I like it, LGTM.

PR #961, opened 2016-12-08
Fix glog dependencies

sboeuf:
glog is a very useful library used by many projects. Because libsnnet package (and database implicitly) is intended to be used from external packages, we cannot rely on a vendored version of glog. This PR allows the calling package define its own clogger, and pass it to imported packages. In case nothing is defined, nothing will be printed because the default implementation of clogger is CiaoNullLogger, generating nothing.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -38,6 +38,7 @@ import (
 	image "github.com/01org/ciao/ciao-image/client"
 	storage "github.com/01org/ciao/ciao-storage"
 	"github.com/01org/ciao/clogger/gloginterface"
+	"github.com/01org/ciao/database"
 	"github.com/01org/ciao/openstack/block"
 	"github.com/01org/ciao/openstack/compute"
 	osIdentity "github.com/01org/ciao/openstack/identity"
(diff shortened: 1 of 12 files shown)


markdryan:
We'll need to initialise database.Logger in

github.com/01org/ciao/ciao-controller

and 

libsnnet.Logger in

github.com/01org/ciao/ciao-launcher
github.com/01org/ciao/networking/ciao-cnci-agent
github.com/01org/ciao/networking/libsnnet/tests/cncicli
github.com/01org/ciao/networking/libsnnet/tests/cncli

sboeuf:
Yes I am adding those commits

markdryan:
@obedmr , @mcastelino could you take a look at this one as well?

mcastelino:
@sboeuf I think you missed changing the ciao-cnci-agent which uses libsnnet.

sboeuf:
Okay I have added missing initializations related to ciao/database and ciao/networking/libsnnet. Let's see if travis complains about something...

mcastelino:
LGTM.

Also I looked that the networking code. It does not really use logging execept in the docker plugin.
This is something we can easily eliminate, which will make it a self contained library independent of ciao.

sboeuf:
@mcastelino yes we are doing all that stuff because of docker plugin, and it would be much better to remove the dependency from there.

sboeuf:
LGTM

PR #966, opened 2016-12-12
networking: libsnnet: Make API more explicit and get existing VNIC device by name

sboeuf:
libsnnet API lets us get an existing VNIC device only from its alias. Because this
API is used externally by other project such as virtcontainers, there are cases where
no alias is set for an existing interface, and we want to be able to retrieve a VNIC
object from this existing interface.

This PR duplicates the existing GetDevice() to provide a new GetDeviceByName, relying
on netlink LinkByName function. It builds a VNIC object according to the information found
for the related link name.

The API SetHardwareAddr() was implicitely creating a Vnic peer related
to the current Vnic, and it was setting the hardware address for this peer.
It was a very specific case for Ciao but we want more explicit things now
that libsnnet is used externally from other projects.

This PR simplifies SetHardwareAddr() function and adds the complexity
in the Ciao's calling code instead.

Also, it returns a default peer name in case the link name does not fit the
expected prefixes. It is needed because in case you set manually the link
name, you will get a creation error. Indeed, the peer name will be "", and
it will be impossible to create a link with that name.

### DIFF
--- a/networking/libsnnet/vnic_test.go
+++ b/networking/libsnnet/vnic_test.go
@@ -184,6 +184,52 @@ func TestVnicContainer_GetDevice(t *testing.T) {
 	performVnicOps(true, assert, vnic)
 }
 
+//Test ability to attach to an existing VNIC
+//
+//Tests the ability to attach to an existing
+//VNIC and perform all VNIC operations on it
+//
+//Test is expected to pass
+func TestVnic_GetDeviceByName(t *testing.T) {
+	assert := assert.New(t)
+	vnic1, _ := NewVnic("testvnic")
+	vnic1.LinkName = "testiface"
+
+	assert.Nil(vnic1.Create())
+	vnic, _ := NewVnic("testvnic")
+
+	assert.Nil(vnic.GetDeviceByName("testiface"))
+	assert.NotEqual(vnic.InterfaceName(), "")
+	assert.Equal(vnic.InterfaceName(), vnic1.InterfaceName())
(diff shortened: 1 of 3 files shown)


networking/libsnnet/vnic_test.go
```diff
@@ -184,6 +184,52 @@ func TestVnicContainer_GetDevice(t *testing.T) {
...
+	vnic1, _ := NewVnic("testvnic")
+	vnic1.LinkName = "testiface"
+
+	assert.Nil(vnic1.Create())
+	vnic, _ := NewVnic("testvnic")
+
+	assert.Nil(vnic.GetDeviceByName("testiface"))
+	assert.NotEqual(vnic.InterfaceName(), "")
+	assert.Equal(vnic.InterfaceName(), vnic1.InterfaceName())
+	assert.NotEqual(vnic1.PeerName(), "")
```
mcastelino: This may be incorrect... why are you checking peerName for a VM Vnic?

sboeuf: You're right. But I guess TestVnic_GetDevice() is also incorrect then.

mcastelino: It works because of https://github.com/01org/ciao/blob/master/networking/libsnnet/vnic.go#L66
I forgot about that. So this is ok

rbradford:
@sboeuf Please look at the travis builds - looks like you introduce a panic triggered by the unit tests.

sboeuf:
@rbradford yes I have noticed those issues, I will investigate them today.

Issue #968, opened 2016-12-13
ansible: upload images with different uuid


erick0z:
Ansible does upload the images but with an incorrect uuid. And after a second run of the ansible playbooks, the conditional is expected to fail because the uuid is not being found an then it proceeds to upload again the images. The issue is related to the white spaces in the ```image.name``` variables used in the ansible task ```roles/ciao-controller/tasks/main.yml```. The conflicting names are:
  - "Clear Cloud"
  - "Fedora 24 Cloud"

When creating the images from ansible, the image names are not being correctly registered and leading to a different uuid assignment.

PR #969, opened 2016-12-13
ansible: Fix image uuids

erick0z:
Creating images from ansible leads to invalid uuid assigment
due to wrong names.

fixes https://github.com/01org/ciao/issues/968


rbradford:
@albertomurillo can you review?

erick0z:
@albertomurillo You're right, I though it was more convenient to not have spaces but it seems there is no convention in Openstack for the image naming.

PR #977, opened 2016-12-15
Storage: fix for failing BAT test

tpepper:
My commit c389aed9b22c360a522630a6d76ca71c8424acda broke BAT by removing some implicit behavior.  In the past code could overriding the ephemerality of a storage volume declared in workload_storage.csv and make it always ephemeral.  This PR allows the existing BAT test, which expected controller auto-created volumes were ephemeral, by marking those volumes as ephemeral instead of persistent in the workload_storage.csv.

There is also a patch here to move single VM from golang 1.7.3 to 1.7.4 as I noticed we're still on the older release during my BAT run in a VM.


tpepper:
@mcastelino this passes unit and bat tests

tpepper:
Re-fixed.  This _now_ passes unit test and BAT.

mcastelino:
@tpepper I like this change. Also it keeps the system clean. When we move away from the csv file based workload definition, I assume we either need to add explicit snapshot capability or be able to specify this at instance launch time. @kaccardi's workload changes may already comprehend this.

Issue #974, opened 2016-12-14
ciao-launcher storage - Image backed volume should not be created when an instance is added using a bootable volume


mcastelino:
See this case

`ciao-cli instance add -workload=e35ed972-c46c-4aad-a1e7-ef103ae079a2 -volume uuid=b3d391e9-309d-4271-b76e-1767c44877b8,boot_index=0`

Here the volume was a previously auto cloned volume from another instance.

The thing that bothers me with this is that we end up with an extra boot volume (the one based off of the image).

This is a bit confusing as, when you boot from a volume, the image definition is immaterial. The bootable volume is the image.

See this, we end up with an extra volume 
```
qemu-system-x86_64
-drive file=rbd:rbd/b3d391e9-309d-4271-b76e-1767c44877b8:id=ciao,if=none,id=drive_b3d391e9-309d-4271-b76e-1767c44877b8,format=raw
-device virtio-blk-pci,scsi=off,bus=pci.0,addr=0x3,id=device_b3d391e9-309d-4271-b76e-1767c44877b8,drive=drive_b3d391e9-309d-4271-b76e-1767c44877b8

-drive file=rbd:rbd/2c1bf8a6-6bb6-4201-b463-acaa683d4bdf:id=ciao,if=none,id=drive_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf,format=raw
-device virtio-blk-pci,scsi=off,bus=pci.0,addr=0x4,id=device_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf,drive=drive_2c1bf8a6-6bb6-4201-b463-acaa683d4bdf

-drive file=/var/lib/ciao/instances/29ea470d-96f6-4689-afd3-ec120d012c4b/seed.iso,if=virtio,media=cdrom
```

rbradford:
Looking at the code (and seeing the above behaviour) indicates that instance storage is additive with workload storage.

To me an instance specific storage specification should (completely) replace the workload storage specification.

This would be mean that if an instance needs two volumes one instance specific and one shared then the user would be responsible for copying that volume id from the workload into the instance command line.

A small compromise for the simple logic that specifying instance storage replaces workload storage.

Any dissent? @tpepper / @kaccardi ?

PR #979, opened 2017-01-03
ciao-controller: Make instance storage override workload storage

rbradford:
The specification of any per-instance storage should override the
workload storage specification.

With this change if any volumes are specified at instance creation the
workload storage information is not used.

Fixes: #974

### DIFF
--- a/ciao-controller/instance.go
+++ b/ciao-controller/instance.go
@@ -359,8 +359,8 @@ func newConfig(ctl *controller, wl *types.Workload, instanceID string, tenantID 
 		userData.Hostname = "cnci-" + tenantID
 	}
 
-	// handle storage resources in workload definition
-	if wl.Storage != nil {
+	// if no instance volumes specified, use those in workload
+	if len(volumes) == 0 && wl.Storage != nil {
 		workloadStorage, err := getStorage(ctl, *wl.Storage, tenantID, instanceID)
 		if err != nil {
 			return config, err


ciao-controller/instance.go
```diff
@@ -359,8 +359,8 @@ func newConfig(ctl *controller, wl *types.Workload, instanceID string, tenantID
 		userData.Hostname = "cnci-" + tenantID
 	}
 
-	// handle storage resources in workload definition
-	if wl.Storage != nil {
+	// if no instance volumes specified, use those in workload
+	if len(volumes) == 0 && wl.Storage != nil {
```
tpepper: We don't have a cohesive design intent on this agreed and documented.

If wl.storage is primarily about specifying a boot image, I think it should have precedence.  If a bootable volume is user specified, I can see that overriding the workload specified one.

rbradford: Right, but what if the workload has two volumes, or two volumes are specified for the instance then it's confusing.

I feel like the behaviour needs to be predictable and explainable - which this is: instance volume specification always overrides workload specification.

tpepper: I feel the workload specifies the predictable, explainable, common parts of a workload.  And then the cli storage extension allows for an additive action of user/instance specific ancillary storage.

rbradford: Okay. So what should happen in this case where the user tries to override the boot device? 

Should we just reject this at instance creation time with a check (banning boot_index=0) ?

Or are you proposing we replace the boot device in the workload with the new one?

tpepper: Good question.  Right now the workload doesn't include a boot_index.  So it's hard to reconcile with the cli's volume option that does include that.

Also I see the OpenStack API boot index argument as an option, especially when coupled with qemu/kvm, that is non-mandatory.  I'm not sure if/how we could really truly enforce boot ordering.  We can order the disks as we present them to qemu, but the actual VM disk enumeration and boot order is going to depend on qemu internals, the emulated bios, the VM's bootloader code and config, the kernel, disk labels, etc.

I don't think we have sufficiently specified supported use cases to declare an expected behavior here.  But:

The workload definition is our top level fundamental definition of the abstract thing that's going to run.  Where the abstract thing has an OS that is booted, IMHO that should be specified in the workload definition.  Were this not the case, it feels like it devolves to: does it make sense to have a workload defined as "Ubuntu 16.10 VM with 1vcpu and 16GB RAM", and then launch an instance of that workload with say a Fedora Cloud Base 24 iso attached, and hope for that iso to be the boot device.  I think no, but you could implement it.

It would help me more to understand the specific case you're trying to run:
* what's the workload definition in detail
* what's the volume list in the instance add cli invocation
* what's the expected outcome in your opinion for this context

rbradford: In #974 @mcastelino expected that specifying a volume with the boot_index=0 would make it use that instead of the workload.

I'm happy with additive (or instead of) but it just needs to be made obvious to the user what's going on.

Also it's probably a good idea to move this discussion to #974

rbradford:
Closing as it looks like the behaviour is working as expected. In #974 we can thrash out how to document volumes provided on the command line are additive.

CC: @mcastelino

PR #985, opened 2017-01-05
Update sqlite3

markdryan:
This PR updates the vendored version of go-sqlite3 to 6f2749a.  The underlying sqlite3 c library is updated to 3.15.1.  The latest version of sqlite3 does not compile with gobeta1 so this PR also updates the version of Go 1.8 used by travis to Go 1.8beta2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ services:
 
 go:
     - 1.7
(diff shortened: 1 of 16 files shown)


markdryan:
BAT tests pass on Single VM with this patch.

Issue #934, opened 2016-12-06
Remove canned workloads from ciao


kaccardi:
ciao-controller prepopulates the datastore with some initial workloads. We should not do this anymore once we have the ability to upload custom workloads, and instead allow singleVM scripts or ansible deployment scripts to add them.

Issue #510, opened 2016-09-01
Support per tenant and public workloads


kaccardi:
workloads should be able to be per tenant, as they contain cloud init yaml code that is not shareable. We should also allow users to specify a workload be public and shareable.

Issue #912, opened 2016-12-01
Ciao has no control over the CNCI lifecycle


erick0z:
Right now, Ciao has no control over the CNCI other than spawning it. The CNCI runs forever since the first VM is launched. It would be good to shut down the CNCI if there is no VMs running and the state of the tenant is set to idle after "x" time.

tpepper:
Part of the lifecycle management is also defining what is the cnci (initial base image), and maintaining that over time (new base image, restart cnci).  With the .csv workload definitions, today I think you'd need to edit the csv file, kill the cnci, restart controller and trigger a cnci relaunch somehow, in order to use a new cnci image.  It would be better if these were handled like any other system workload.

rbradford:
Fixed by #1404

PR #991, opened 2017-01-06
ciao-controller: add tenant separated workloads and public workloads

kaccardi:
This commit allows users to have private workloads and makes
all admin created workloads public. It also updates singlevm
start up script to create a couple public workloads.

This PR deletes all the canned workloads and csv files.

Fixes: #510 
Fixes: #934 
Fixes: #947
Fixes: #948


albertomurillo:
Hi @kaccardi Can you rebase this branch?
I will fork this branch to work on the ansible scripts to remove the tables and add a section to
create custom workloads via the cli.

Also, I just saw the examples under ciao-cli and I think it would be better to have a 
`demo_user.yml` cloud init file and an example for fedora and clearlinux which both of them can reuse the same `demo_user.yml` cloud init file.

mcastelino:
@kaccardi can an admin tenant create a non public workload?

albertomurillo:
@kaccardi a `ciao-cli workload show -workload <id>` is missing.

kaccardi:
@albertomurillo we previously agreed to keep workloads with "create" and move the rest of the cli to use create at some point. The show feature is not part of this PR - it's a later stage of development.

albertomurillo:
This works for me, I Still have some workarounds to fix on ansible roles to make it create the example workloads without issues.

kaccardi:
@mcastelino no - the admin can only create pubic workloads.

kaccardi:
I confirmed the existing set of bat tests will work as long as you have setup your cluster with some public workloads, such as what is done in setup.sh

testutil/singlevm/setup.sh
```diff
@@ -154,14 +155,17 @@ function createWorkloads() {
 	EOF
 	) > "$ciao_bin"/workload_examples/fedora_vm.yaml
 
+	# get the clear image id
```
mcastelino: @kaccardi with https://github.com/01org/ciao/pull/1004 you will have suffix matching, which will make this simpler

ciao-controller/workload.go
```diff
@@ -122,6 +137,24 @@ func (c *controller) CreateWorkload(req types.Workload) (types.Workload, error)
 		return req, err
 	}
 
+	// check to see if this is a new tenant. If so, we need to add
+	// them to the datastore. We do not however want to launch a
+	// CNCI yet since this might be a request to upload a CNCI
+	// workload. Instead, we'll add the new tenant directly to the
+	// datastore. On first launch request, if the tenant doesn't yet
+	// have a cnci, it will get launched for them then.
+	tenant, err := c.ds.GetTenant(req.TenantID)
```
rbradford: I also spotted that "ciao-cli volume list" will start the CNCI. Is there another codepath we need to protect against there too?

ciao-controller/api/api.go
```diff
@@ -440,12 +440,29 @@ func addWorkload(c *Context, w http.ResponseWriter, r *http.Request) (Response,
...
 
+	// we allow admin to create public workloads for any tenant. However,
+	// users scoped to a particular tenant may only create workloads
+	// for their own tenant.
+	vars := mux.Vars(r)
+	tenantID, ok := vars["tenant"]
+	if ok {
+		req.TenantID = tenantID
+	} else {
+		req.TenantID = "public"
```
rbradford: Ah, I see you comment on it below. Are we opening up any security concerns with this pseudo-tenant "public" - do you also forsee using the same thing for images too..?

ciao-controller/instance.go
```diff
@@ -46,7 +46,7 @@ type instance struct {
 	startTime time.Time
 }
 
-func isCNCIWorkload(workload *types.Workload) bool {
+func isCNCIWorkload(workload types.Workload) bool {
```
tpepper: I think it's idiomatic to not use pointers in a case like this.  It's probably not really a copy under the covers, unless the function code modifies the argument's contents.  Trust in the go memory manager.

kaccardi: I changed it exactly because of what @tpepper said.

ciao-controller/controller_test.go
```diff
@@ -1150,6 +1209,12 @@ func TestGetStorageForVolume(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	//wl := types.Workload{
+	//	ID:      "validID",
+	//	ImageID: uuid.Generate().String(),
+	//	Storage: &s,
+	//}
+
```
kaccardi: no - I had found that code in there and thought I didn't need it, so I commented it out to see if tests still worked without it, then forgot about it.

ciao-controller/controller_test.go
```diff
@@ -1201,6 +1266,12 @@ func TestGetStorageForImage(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	//wl := types.Workload{
+	//	ID:      "validID",
+	//	ImageID: filepath.Base(tmpfile.Name()),
+	//	Storage: &s,
+	//}
+
```
kaccardi: ditto.

ciao-controller/api/api.go
kaccardi: the identity code we use checks that your token is scoped to the tenantID - so that prevents normal users from messing with "public". I guess if you had admin rights you might be able to come in using tenantID "public" on apis where we haven't added the mux regex matching for the uuid4 tenantID yet (the openstack ones I believe) - but I think if you have admin rights you can pretty much do whatever you want to the entire cluster anyway.

kaccardi: I don't know how @obedmr is planning on storing public images.

ciao-controller/instance.go
markdryan: I'm not sure the compiler will avoid the copy here.  The spec says the the function will receive a copy of the argument and doesn't mention that optimizations are allowed.  I guess we could check the assembly.  But in general I think it's reasonable to pass values by type if the parameter is large.  In this case it's 160 bytes.  The best justification I can find is here:

https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

The 5th bullet point states that

```If the receiver is a large struct or array, a pointer receiver is more efficient. How large is large? Assume it's equivalent to passing all its elements as arguments to the method. If that feels too large, it's also too large for the receiver.
```

A receiver here is just an argument so I'm extrapolating to arguments in general.  The fact that it's faster to use pointer rather than value receivers implies that for receivers at least the compiler doesn't do any optimisation.

ciao-controller/workload.go
kaccardi: so - I thought about this - right now anytime confirmTenant() is used (in many of the volume calls) if there isn't a CNCI workload ID, you'll get an error returned and the call will fail. I personally think this is ok - it means the admin hasn't configured the cluster properly. Even though in theory we could of course make it possible to access volume services without the CNCI being configured properly, I'd rather it cause an error that forces them to finish setting up the cluster. the workloads api is special of course.

ciao-controller/controller_test.go
kaccardi: @rbradford I pushed a commit which removes this code as a separate patch - I can squash it in if you want me to.

ciao-controller/workload.go
kaccardi: btw - for now this will never be a problem since the CNCI workload is actually generated by controller. In the future we might have per tenant CNCI workloads.

ciao-controller/instance.go
kaccardi: @markdryan ok, thanks for clarifying. I did have another reason as well - and that was that I was trying to reduce the likelyhood that a receiver would modify the struct. It felt safer to send a copy.

kaccardi: @markdryan another thing that would be a great clarification for me - in the go class we took, they told us to avoid passing by pointer if possible due to it complicating garbage collection and go's escape analysis. At what point do we need to worry about that?

ciao-controller/controller_test.go
kaccardi: I went ahead and squashed it in since I had to rebase anyway.

ciao-controller/api/api.go
obedmr: @kaccardi for image service, I'm planning to use the ``visibility`` metadata field for specifiying if the image is public, internal or private. A similar approach as the used at: https://wiki.openstack.org/wiki/Glance-v2-community-image-visibility-design

ciao-controller/instance.go
markdryan: > I did have another reason as well - and that was that I was trying to reduce the likelyhood that a receiver would modify the struct. It felt safer to send a copy.

Unfortunately, this is really hard to achieve in go even when passing by value, unless the parameter is of a primitive type or is an array or struct containing only simply types.  The reason is the contents of maps and slices are modifiable by a function even if they are passed to that function by value, either directly or part of a struct which is itself passed by value.  For example, it's still possible to modify the caller's copy of the workload parameter inside this function even though it is passed by value

workload.Default[0].Type = payloads.VCPUs

This change will be visible outside isCNCIWorkload.

You could always change the parameter of this function to take a []payloads.RequestedResource.  That wouldn't prevent the function from modifying the slice but would bypass the value vs pointer discussion.

markdryan: > they told us to avoid passing by pointer if possible due to it complicating garbage collection and go's escape analysis. At what point do we need to worry about that?

I haven't seen this advice given elsewhere and it sounds a little bit like premature optimisation to me.  I think we only need to worry about this if we see that our services are doing too much GC.  This is probably something we should profile at some point.  The easiest way to get an idea is to prefix your executable with

GODEBUG=gctrace=1

Interestingly you can see which objects escape a function by passing using 

go build -gcflags="-m"

testutil/singlevm/setup.sh
```diff
@@ -105,6 +105,127 @@ function ctrl_c() {
...
+	defaults:
+	  vcpus: 2
+	  mem_mb: 512
+	  disk_mb: 80
+	cloud_init: "container-test.yaml"
+	EOF
+	) > "$ciao_bin"/workload_examples/ubuntu_latest.yaml
+
+	# store the new workloads into ciao
+	cd "$ciao_bin"/workload_examples
```
rbradford: Use pushd and popd rather than this?

testutil/singlevm/setup.sh
```diff
@@ -375,9 +458,17 @@ qemu-img convert -f raw -O qcow2 "$ciao_cnci_image" "$ciao_cnci_image".qcow
...
+if [ $download -eq 1 ]
+then
+	LATEST=$(curl https://download.clearlinux.org/latest)
+else
+	# replace this will a function that looks in ~local cache
+	# for the last version of clear to use.
+	LATEST="12620"
+fi
+
+if [ $download -eq 1 ] && [ ! -f clear-"${LATEST}"-cloud.img ]
```
rbradford: This will only download now if downloads are forced on which is not the default.

testutil/singlevm/setup.sh
```diff
@@ -394,7 +485,7 @@ fi
 
 #Fedora, needed for BAT tests
 cd "$ciao_bin"
-if [ $download -eq 1 ] || [ ! -f $fedora_cloud_image ]
+if [ $download -eq 1 ] && [ ! -f $fedora_cloud_image ]
```
rbradford: ditto.

testutil/singlevm/setup.sh
rbradford: See line 229 above: https://github.com/01org/ciao/pull/991/commits/b5c676e20fbfd206e9c26edcdbb1d56f1b6d3034#diff-04ed742f8328787431b9975d1866ffd2R229

testutil/singlevm/verify.sh
```diff
@@ -281,32 +283,6 @@ checkEventStatus $event_counter "Unmapped"
 
 "$ciao_gobin"/ciao-cli external-ip list
 
-#Test for External IP Failures
-
-#Map failure
```
markdryan: Why did we remove these tests?

testutil/singlevm/setup.sh
```diff
@@ -105,6 +105,127 @@ function ctrl_c() {
...
+	echo "Generating Fedora VM test workload"
+	(
+	cat <<-EOF
+	description: "Fedora test VM"
+	vm_type: qemu
+	fw_type: legacy
+	image_id: "$id"
+	defaults:
+	    vcpus: 2
+	    mem_mb: 512
```
markdryan: I don't think these values match what we used to have in SingleVM.  If I read the resource file correctly, the fedora workload had 128MB or ram and 80GB.  These differences might be significant for ciao-down users.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -2280,3 +2257,61 @@ func (ds *Datastore) UnMapExternalIP(address string) error {
...
+`
+	cpus := payloads.RequestedResource{
+		Type:      payloads.VCPUs,
+		Value:     4,
+		Mandatory: false,
+	}
+
+	mem := payloads.RequestedResource{
+		Type:      payloads.MemMB,
+		Value:     128,
```
markdryan: Currently, we have two separate workload_resources.csv files, one for SingleVM and one for normal clusters.  The singleVM version of workload_resources.csv gives the CNCI 128MB ( we want to keep it small  for SingleVM ) but the controller copy of the file, the one used in real clusters, gives it 2048MB disk and 2048MB RAM.  Is this commit using 128MB for all CNCI instances regardless of whether they are deployed on SingleVM or a real cluster?  If so I don't think it's correct.  We'll probably run into memory issues on real deployments.  I think we need to be able to specify different CNCI workloads.

@mcastelino Can you comment on this?

testutil/singlevm/verify.sh
kaccardi: because there is no longer any way to set ssh keys for the cnci vm.

ciao-controller/internal/datastore/datastore.go
kaccardi: I could do this 2 ways - one way is to add a flag available only to the privileged user to be able to create CNCI workloads. The disadvantage of this approach is that if the admin messes up the CNCI workload definition, his cluster is hosed. Another option is to make parts of the cnci configurable via a separate api (such as /cncis) and have only certain values be configurable (like mem or disk for example). this has the disadvantage of needing yet another interface, but reduces the likelyhood of screwing up the cnci definition.

kaccardi: Adding an api one way or another to configure the cnci has been my plan all along btw, I just thought I'd do it in another PR. But I can do it in this one, no problem. it would also resolve the issue of not being able to pass ssh keys to the cnci for debug (and those map failure tests in verify.sh)

mcastelino: @kaccardi the 128MB CNCI cannot handle full 10 Gbps worth of traffic, which is our minimal cluster requirement. So as @markdryan suggests that is a parameter we would like to control. However that should be derived based on the size of the tenant (vs an explicit CNCI configuration IMO). I know we do not have a tenant size hint today.

tpepper: We'd probably want at the same time the ability to start/stop/recycle the CNCI, so the hosed can attempt a reconfigure and hopefully get unhosed.

testutil/singlevm/verify.sh
mcastelino: @kaccardi @markdryan this was a corner case that  I wanted to validate. However till we have CNCI ssh key provisioning support, we can disable the same. Without this case one of the events (which in a sane ciao setup, should never occur will not get validated).

ciao-controller/internal/datastore/datastore.go
mcastelino: @kaccardi the tenant actually has a field called InstanceLimit. So that can be used a trigger I assume. Today it is set by default to unlimited in the case of SingleVM. But in the ciao-cli I do not seem to have an option to specify the same.

mcastelino: @tpepper the stop,increase limit,restart (which may need re-scheduling to an appropriate NN) will be needed if we want to support scaling of a tenant/project.

kaccardi: @mcastelino currently limit setting in not implemented in ciao. we'd also need to have a way to deal with the unlimited case.

markdryan: I'd say we need to find a way of giving the CNCI the appropriate amount of memory for both SingleVM and real clusters as part of this PR.  If we commit without a solution to this we'll either break Single VM (for people with less powerful computers) or we break ciao on real clusters.

#912 is important but I would say doesn't need to be done in this PR as it is an existing problem.

kaccardi: @markdryan @mcastelino I created issue #1024 and #1023 . I'll add support for determining the CNCI memory resource based on instance limit as mcastelino has requested. I'll add it to this PR as mark has requested. we'll need to add some steps to singleVM to limit the number of instances allowed for the test tenant.

kaccardi: and #1022 to track creation of the algorithm for determining the cnci resource.

rbradford:
After finding a little bug with the existing workload creation stuff and with such a large change I think the BAT tests that exercise the workload functionality added in this PR should be included in this PR.

ciao-controller/internal/datastore/datastore.go
kaccardi: @markdryan how do you feel about repurposing the "singlemachine" parameter to add cnci configuration for singleVM? The defaults can be for a "real" cluster (i.e. what we used to use) for now. We can work on determining the CNCI memory resources needed programmatically in a different PR - I'm worried about this one getting quite large.

kaccardi: the other choice is that I wait for the quotas & limits work to be done and forget the singlevm config for resources.

markdryan: >  how do you feel about repurposing the "singlemachine" parameter to add cnci configuration for singleVM? 

I don't really understand what is meant by the "singlemachine" parameter.  I do understand your concern about the PR getting too big though.   So if you have a way, even a hacky way, to ensure both single vm and normal ciao clusters work as before after this PR is committed, that's good enough for me.

testutil/singlevm/verify.sh
kaccardi: @markdryan @mcastelino I added a new commit that allows the user to set the CNCI size in the configuration.yaml file for now. It's optional, and if it's not set, controller defaults will be used. We can remove this later once we have a programmatic way of determining the CNCI size. Take a look and tell me what you think.

mcastelino:
@kaccardi The CNCI changes look good. I suggested add a config var for admin password also.

Can we add the unmap failure test back again? That will actually end up testing the admin ssh key provisioning also.

kaccardi:
@mcastelino I got the unmap failures back in and verify.sh is working now.

kaccardi:
@rbradford @markdryan I wrote a couple bat tests for the "create" functionality. Can you take a look and let me know what you think? also, @albertomurillo has sent a PR to this branch for the deployment changes. Should I merge it into this PR, or should we review separately? I'd recommend merging in this PR since this PR will break deployment otherwise.

_release/bat/workload_bat/workload_bat.go
```diff
@@ -0,0 +1,18 @@
+//
+// Copyright (c) 2016 Intel Corporation
```
markdryan: A bit pedantic, but should this be 2017?

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -0,0 +1,134 @@
...
+	} else {
+		ID, err = bat.CreateWorkload(ctx, tenant, opt, vmCloudInit)
+	}
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// now get all the workloads
+	wls, err := bat.GetAllWorkloads(ctx, "")
```
markdryan: Why don't we just call GetWorkload here?

bat/workload.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+// this function should output the yaml file for the workload definition
+func createWorkloadDefinition(filename string, opt WorkloadOptions) error {
+	// take the cloud config file and output it to a file.
+	f, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
markdryan: When closing writable files it's best to check the return from close.  I usually do this by using a named returned value, e.g,

func createWorkloadDefinition(filename string, opt WorkloadOptions) (err error) {

and then the defer becomes

defer func() {
   closeErr := f.Close()
   if err == nil {
      err = closeErr
   }
}

bat/workload.go
```diff
@@ -0,0 +1,196 @@
...
+
+	y, err := yaml.Marshal(opt)
```
markdryan: Of course you could just use ioutil.WriteFile and then you wouldn't need to worry about creating or closing at all.

bat/workload.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	defer f.Close()
+
+	_, err = f.WriteString(config)
+	return err
+}
+
+func createWorkload(ctx context.Context, tenant string, opt WorkloadOptions, config string, public bool) (string, error) {
+	// write out the cloud config.
+	err := createCloudConfig("config.yaml", config)
```
markdryan: It might be better to write these files to somewhere under /tmp to avoid cluttering the working directory (in the case of a crash for example).  There are functions in ioutil for creating temporary files and directories.

bat/workload.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return workloads, nil
+}
+
+// GetWorkload will return a specific workload referenced by name.
+// An error will be returned if either no workloads exist in the cluster,
+// or if the specific workload does not exist. It inherits all error
+// conditions of GetAllWorkloads.
+func GetWorkload(ctx context.Context, tenant string, name string) (Workload, error) {
```
markdryan: A ciao-cli workload show would be nice.

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, w := range wls {
+		if w.ID == ID {
+			if w.Name != opt.Description || w.CPUs != opt.Defaults.VCPUs || w.Mem != opt.Defaults.MemMB {
+				t.Fatalf("Workload not defined correctly")
+			}
+			return
+		}
+	}
+
+	t.Fatalf("New workload did not appear in workload list")
```
markdryan: Shouldn't we delete the workload?

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// TestCreateTenantWorkload will create a new private workload, then attempt to
```
markdryan: The documentation for the test cases is supposed to have three paragraphs, the first is the test title (this is what appears in the tap reports), the second explains what the test does and the third explains what should happen.

markdryan:
@kaccardi The tests look good.  I've just made a few small comments.

_release/bat/workload_bat/workload_bat_test.go
markdryan: I guess this is the main blocker for me.  Is there any way to cleanup after running the tests?

rbradford:
@kaccardi It sounds good to me for you to cherry-pick his changes onto your branch.

bat/workload.go
kaccardi: @markdryan workload show will be done in a different PR.

_release/bat/workload_bat/workload_bat_test.go
kaccardi: @markdryan we don't have workload delete support yet, so no. Delete is going to be a lot more code, so I'd really prefer to submit it in a different PR. We could simply not merge the BAT test portion till we have delete support though.

rbradford:
@kaccardi I think the BAT tests look pretty good acknowledging @markdryan's feedback and yours on his.

_release/bat/workload_bat/workload_bat_test.go
markdryan: Fair enough.  Let's just commit the tests but skip them for now.

kaccardi:
@rbradford I think I got everything addressed and also included the deployment patch. possibly it's good to go?

tpepper:
The last item in the PR message's TODO list is done now too with the most recently added patch, or?  If so maybe just edit and remove the whole TODO list?

kaccardi:
@tpepper actually I haven't done the last item - that's updating the release cluster scripts. Doesn't require further changes to this PR, and I'm going to do it immediately after this gets merged. I'll delete it from this list to avoid confusion - it was there as a reminder to myself.

rbradford:
@kaccardi This works well on SingleVM (well..i'm sure you tested it there anyway :-)

On a real cluster I had some problems related to #1005 where anything using the Ciao HTTP API didn't work as the controller was being accessed by FQDN but urls returned to the client to use to access that endpoint only used os.Hostname() giving TLS errors.

With the attached changes I have been able to run ciao-cli workload create against an ansible cluster and intantiate that workload to an instance.

Unfortunately BAT has issues running against an ansible cluster due to admin having multiple tenants that it can access. That's not related to your change though so I don't see any need to hold up merging this if you're happy with my additional changes.

kaccardi:
@rbradford thanks for getting this sorted out, while I'm disappointed we have to add yet another configuration option this seems like a simple options for now.

PR #811, opened 2016-11-10
Store images in ceph

rbradford:
Unify all storage on ceph - including volumes and images. As part of this the major changes needed are:

1. Add a new backend to the image service to upload the images into rbd images in ceph.
2. Enhance the ceph storage backend to snapshot and protect the rbd images used to store ciao images and expose the rbd clone functionality that lets you create an copy-on-write volume backed by the image to boot.
3. Modify ciao-controller to use the new functionality when creating storage based workloads.
4. Modify workload templates to use storage.

PR #997, opened 2017-01-06
BAT enhancements for storage

rbradford:
This PR adds some new BAT tests for storage that test the different source types whilst also testing the new size reporting and calculation.

This identified some issues also fixed by this PR.

@kaccardi to review ciao-controller change as owner
@markdryan to review bat changes as owner

### DIFF
--- a/_release/bat/storage_bat/storage_bat_test.go
+++ b/_release/bat/storage_bat/storage_bat_test.go
@@ -646,20 +646,18 @@ func TestCreateVolumeFromImage(t *testing.T) {
 	ctx, cancelFunc := context.WithTimeout(context.Background(), standardTimeout)
 	defer cancelFunc()
 
-	images, err := bat.GetImages(ctx, "")
-	if err != nil {
-		t.Fatal(err)
+	options := bat.ImageOptions{
+		Name: "test-image",
 	}
 
-	var imageUUID string
-	var image *bat.Image
-	for imageUUID, image = range images {
-		break
+	image, err := bat.AddRandomImage(ctx, "", 10, &options)
+	if err != nil {
+		t.Fatalf("Unable to add image %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


_release/bat/storage_bat/storage_bat_test.go
```diff
@@ -634,3 +635,88 @@ func TestAttachToStoppedInstance(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	var imageUUID string
+	var image *bat.Image
+	for imageUUID, image = range images {
+		break
+	}
+
+	volumeUUID, err := bat.AddVolume(ctx, "", imageUUID, "image", &bat.VolumeOptions{})
```
markdryan: If there are no images this will probably fail which would be a bit misleading.  Might be best to skip the rest of the test if imageUUID == "" or len(images) == 0

markdryan:
Works fine for me although I did have one comment about one of the tests.  Thanks for the go vet fixes.  I'm working on a PR that enables static anaylsis on the BAT tests, and fixes a few other small issues.

_release/bat/storage_bat/storage_bat_test.go
rbradford: Updated version of the patch now creates a test image to use.

rbradford:
> What's the rationale for everything being a snapshot? Is this needed because each user needs to be able to write, and you can only have one writer?

We always use a snapshot for images. This fix should really have been part of #811 but we didn't have any test coverage of "ciao-cl volume add -source_type image -source <image uuid>" until this BAT test so it slipped through. Without this change that command fails as ciao-conttoller tries to find a file on disk.

Using snapshots gives us CoW behaviour of images allowing the underlying blocks to be shared across different volumes. This both saves space but also improves the cachability as more blocks are common across volumes.

Issue #982, opened 2017-01-04
Let ciao packages create the ciao user


albertomurillo:
The ciao user should not be created from the ansible scripts.

ClearLinux provides the ciao user on its systemd configuration.
Ciao package for fedora creates the ciao user.
Ciao package for ubuntu is NOT creating the ciao user.

Once the ubuntu package creates the ciao user, remove "Create ciao user" task from ciao-common/tasks/main.yml

Issue #899, opened 2016-11-30
ciao-deploy: Install dependencies


albertomurillo:
osprepare installs missing dependencies when the service starts but there is no way osprepare to yell when all the dependencies are installed and the service is ready to be used.

ciao_wait.yml include a task to check on the nodes that all the binaries have been installed.

It would be better to let ansible install the dependencies in the ciao nodes before starting the service. this way the services are ready to be used and any error installing the dependencies can be handled in the ansible playbooks.

erick0z:
@albertomurillo the new CIAO Packages(ciao-controller, ciao-network, ciao-compute) provided by OBS now install the dependencies needed by Ciao except for Docker.

albertomurillo:
ciao.spec file for fedora and ciao.dsc file for ubuntu now installs the required dependencies
Source code available at https://build.opensuse.org/package/show/home:clearlinux:preview:ciao/ciao

This issue can be closed now

PR #1000, opened 2017-01-09
Deployment CleanUp

albertomurillo:
This PR Fixes two issues:

https://github.com/01org/ciao/issues/982
https://github.com/01org/ciao/issues/899


_DeploymentAndDistroPackaging/ansible/cleanup/ciao.yml
```diff
@@ -39,12 +39,13 @@
...
 
     - name: Stop and disable services
-      service: name={{ item }} state=stopped enabled=no
+      systemd: name={{ item }} state=stopped enabled=no
       with_items:
         - docker-keystone.service
         - docker-ciao-webui.service
         - ciao-controller.service
         - ciao-scheduler.service
+      ignore_errors: yes
```
tpepper: What's the rationale for this addition?

albertomurillo: if `ciao-scheduler.service` exists, then ansible makes sure the service is disabled and stopped.

If the file does not exists. Ansible fails to make sure the service is disabled and stopped.

`ignore_errors: yes` let ansible continue with the other tasks as the service is in the desired state (disabled and stopped) when the service does not exist.

erick0z:
@albertomurillo +1, It is working as expected.

Issue #945, opened 2016-12-07
Add support for multiple volumes per workload.


kaccardi:
the types.Workload structure only allows a single storage resource per Workload. This should be modified to allow a workload to specify multiple storage resources (i.e. an array).

PR #1003, opened 2017-01-10
ciao-controller: Support multiple storage resources per workload

rbradford:
A prerequisite of converting a running instance to a workload is being
able to support multiple types.StorageResource per workload.  This
change enables that.

Creating a workload from fedora_cloud_disk.yaml and then creating an
instance from that now shows two volumes created.

Fixes: #945


rbradford:
@kaccardi please can you take a look at this change

PR #1004, opened 2017-01-10
ciao-cli: Add support for generic string matching in templates

mcastelino:
ciao-cli: Add support for substring matching

Add support for new filter capabilities. This allows the selection
of list elements based on substring, prefix, suffix matching etc
which will be useful for test automation.

### DIFF
--- a/ciao-cli/template.go
+++ b/ciao-cli/template.go
@@ -27,17 +27,27 @@ import (
 )
 
 const templateFunctionHelp = `
-ciao-cli adds three new functions to Go's template language
+ciao-cli adds some new functions to Go's template language
 
 - tojson outputs the specified object in json format, e.g., {{tojson .}}
 - filter operates on an slice or array of structures.  It allows the caller
   to filter the input array based on the value of a single field.
   The function returns a slice containing only the objects that satisfy the
   filter, e.g.
-    
+
   ciao-cli image list -f '{{$x := filter . "Protected" "true"}}{{len $x}}'
-   
+
   outputs the number of protected images maintained by the image service.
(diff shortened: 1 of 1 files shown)


ciao-cli/template.go
```diff
@@ -61,6 +70,40 @@ func findField(fieldPath []string, v reflect.Value) reflect.Value {
 	return f
 }
 
+func filterBySubstr(obj interface{}, field, val string) (retval interface{}) {
```
markdryan: This function is almost identical to filterByField.  Perhaps we need to make a new function, fieldField that takes a function as a parameter, e.g.,

```
func filterField(obj interface{}, field, val string, cmp func(string,string) bool)(retval interface{}) {
```

filterBySubstr then becomes

```
func filterBySubstr(obj interface{}, field, val string) (retval interface{}) {
     return filterField(obj, field, val, strings.Contains)
}
```

filterByField

```
func filterByField(obj interface{}, field, val string) (retval interface{}) {
     return filterField(obj, field, val, func(a, b string) bool {
         return a == b
    })
}

```

You could then have a whole file of filter functions, with now extra effort, e.g,

filter
filterSubstr (might be better to call is filterContains to match the std lib function called)
filterHasPrefix
filterHasSuffix
filterRegexp
filterFolded

mcastelino: @markdryan makes perfect sense. I will also add a few more generic filters based on what the strings suport today and with the same nomenclature. That may help in other areas also, specially event processing.

mcastelino: @markdryan updated as per review comments. Please re-review

PR #1017, opened 2017-01-16
Improve error reporting in controller

rbradford:
Where we are able to return an error, wrap the error so that there is some useful context available for debugging.

Where we are not able to return an error (or on paths with multiple potential errors) ensure that the errors that do occur are logged in a consistent manner.

Also modify the prototype of GetStorageAttachments not to return an error as it cannot do so.

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -350,7 +350,7 @@ func (ds *Datastore) AddTenantChan(c chan bool, tenantID string) {
 func (ds *Datastore) AddLimit(tenantID string, resourceID int, limit int) error {
 	err := ds.db.addLimit(tenantID, resourceID, limit)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "error adding limit to database")
 	}
 
 	// update cache
@@ -370,14 +370,14 @@ func (ds *Datastore) AddLimit(tenantID string, resourceID int, limit int) error 
 
 	ds.tenantsLock.Unlock()
 
-	return err
+	return nil
 }
 
 func newHardwareAddr() (net.HardwareAddr, error) {
 	buf := make([]byte, 6)
 	_, err := rand.Read(buf)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error reading random data")
 	}
 
 	// vnic creation seems to require not just the
@@ -401,26 +401,24 @@ func newHardwareAddr() (net.HardwareAddr, error) {
 func (ds *Datastore) AddTenant(id string) (*types.Tenant, error) {
 	hw, err := newHardwareAddr()
 	if err != nil {
-		glog.V(2).Info("error creating mac address", err)
-		return nil, err
+		return nil, errors.Wrap(err, "error creating MAC address")
 	}
 
 	err = ds.db.addTenant(id, hw.String())
 	if err != nil {
-		glog.V(2).Info("error adding tenant", err)
-		return nil, err
+		return nil, errors.Wrapf(err, "error adding tenant (%v) to database", id)
 	}
 
 	t, err := ds.getTenant(id)
 	if err != nil || t == nil {
-		glog.V(2).Info(err, " unable to get tenant: ", id)
+		return nil, err
 	}
 
 	ds.tenantsLock.Lock()
 	ds.tenants[id] = t
 	ds.tenantsLock.Unlock()
 
-	return &t.Tenant, err
+	return &t.Tenant, nil
 }
 
 func (ds *Datastore) getTenant(id string) (*tenant, error) {
@@ -433,7 +431,8 @@ func (ds *Datastore) getTenant(id string) (*tenant, error) {
 		return t, nil
 	}
 
-	return ds.db.getTenant(id)
+	t, err := ds.db.getTenant(id)
+	return t, errors.Wrapf(err, "error getting tenant (%v) from database", id)
 }
 
 // GetTenant returns details about a tenant referenced by the uuid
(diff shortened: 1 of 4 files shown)


rbradford:
@kaccardi to review

ciao-controller/internal/datastore/datastore.go
```diff
@@ -401,26 +401,24 @@ func newHardwareAddr() (net.HardwareAddr, error) {
...
 	if err != nil {
-		glog.V(2).Info("error adding tenant", err)
-		return nil, err
+		return nil, errors.Wrapf(err, "error adding tenant (%v) to database", id)
 	}
 
 	t, err := ds.getTenant(id)
 	if err != nil || t == nil {
-		glog.V(2).Info(err, " unable to get tenant: ", id)
+		return nil, err
```
tpepper: Is this one intentionally bare?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -643,7 +644,6 @@ func (ds *Datastore) GetAllTenants() ([]*types.Tenant, error) {
 	// numbers of tenants. I'd rather reuse the code
 	// than make this more efficient at this point.
 	ts, err := ds.getTenants()
-
 	if err != nil {
 		return nil, err
```
tpepper: Intentionally bare?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1107,14 +1104,17 @@ func (ds *Datastore) deleteInstance(instanceID string) (string, error) {
...
 
-	err = ds.ReleaseTenantIP(i.TenantID, i.IPAddress)
-	if err != nil {
-		glog.V(2).Info("deleteInstance: ", err)
+	if tmpErr := ds.ReleaseTenantIP(i.TenantID, i.IPAddress); tmpErr != nil {
+		glog.Warningf("error releasing IP for instance (%v): %v", i.ID, err)
+		if err == nil {
+			err = errors.Wrapf(err, "error releasing IP for instance (%v)", i.ID)
+		}
 	}
```
tpepper: Are the conditional wraps here required?  I understood from https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully examples that you could just wrap nil and the right thing happened.

And is keeping the glog useful?  I understood the idiom with errors.Wrap() to be one where you no longer do that.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1154,21 +1154,23 @@ func (ds *Datastore) HandleStats(stat payloads.Stat) error {
...
 	for index := range trace.Frames {
 		i := trace.Frames[index]
 
-		err := ds.db.addFrameStat(i)
-		if err != nil {
-			glog.Warning(err)
+		if tmpErr := ds.db.addFrameStat(i); tmpErr != nil {
+			if err != nil {
+				err = errors.Wrapf(tmpErr, "error adding stats to database")
+			}
```
tpepper: Another one where I thought you could just wrap unconditionally.

ciao-controller/internal/datastore/datastore.go
markdryan: Good point.  I checked the code and Wrap(nil, "My message") returns nil.

ciao-controller/internal/datastore/datastore.go
rbradford: Yes intentionally base. As ds.getTenant() wraps and we can't add any useful context over that one.

ciao-controller/internal/datastore/datastore.go
rbradford: In this case I want to continue to do further cleanups and not return an error immediately. So therefore i'm logging so that both potentially errors will be seen even if only the first one is returned.

Hence why I save the error result into tmpError and then update err with the one that will be actually returned.

ciao-controller/internal/datastore/datastore.go
rbradford: Here I want to continue around the loop even if any one errors out. The code is not consistent on this but i'm simply trying to adhere to the previous behaviour as closely as possible. A later cleanup might decide it's fine to abort halfway through the loop.

There is a bug here though, it should say if err == nil not err != nil.

ciao-controller/internal/datastore/datastore.go
rbradford: Yes, as we have nothing useful to add over the message in ds.getTenants()

ciao-controller/internal/datastore/datastore.go
rbradford: And now fixed.

ciao-controller/internal/datastore/datastore.go
tpepper: I was thinking in this function you could unconditionally do a series of errors.Wrapf(err, "text") and you'd end up with the cumulative set of wrapped errors if multiple, or the single one if just one, or none if none, and then return which ever it is in the single err at the end.

rbradford: I think that would be weird as you normally do that for a call chain, not multiple errors in the same function.

ciao-controller/internal/datastore/datastore.go
tpepper: Ok makes sense.  ds.getTenants() either does something that can't fail or calls ds.db.getTenants(), where the latter would do an errors.Wrap().

ciao-controller/internal/datastore/datastore.go
tpepper: The officially expected usage paradigm is unclear.  But from reading the documentation and watching the video, I see the wrap function as adding information to a stack of events where the stack is multiple errors.  Whether those are from one function or a series of functions doesn't seem different to me in the abstract.

Issue #990, opened 2017-01-05
launcher needs to use GiB


rbradford:
@markdryan mentioned that in launcher reports storage values as GB (10^9) vs GiB (2^30).

For consistency with our Open Stack frontend code and our internals we should track in GiB.

PR #1018, opened 2017-01-16
ciao-launcher: Use MiB in the launcher storage code

rbradford:
This makes us more consistent with the rest of ciao and general industry
expectations.

Fixes: #990

### DIFF
--- a/ciao-launcher/docker.go
+++ b/ciao-launcher/docker.go
@@ -495,7 +495,7 @@ func (d *docker) computeInstanceDiskspace() int {
 		return -1
 	}
 
-	return int(*con.SizeRootFs / 1000000)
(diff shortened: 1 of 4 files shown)


rbradford:
@markdryan to review

tpepper:
Lgtm, but awaiting Mark

markdryan:
The launcher changes look good but should we also change the names of payloads.Ready.DiskTotalMB and payloads.Stat.DiskTotalMB?  Should we also change the text in the WebUI from MB to MiB?

rbradford:
@markdryan Will do - but i'll put those into a separate PR as they're quite widespread and should really change the database schemas.

markdryan:
@rbradford Sounds good to me.

Issue #1009, opened 2017-01-12
storage_bat.TestDeleteInstanceWhileAttaching is flaky


rbradford:
```
Logs for failed tests
---------------------
Logs for storage_bat.TestDeleteInstanceWhileAttaching
	storage_bat_test.go:315: Failed to delete volume 36d61091-9063-4fea-8360-80047b5a8511 : failed to launch ciao-cli [volume delete --volume 36d61091-9063-4fea-8360-80047b5a8511] : exit status 255
		F0112 14:28:33.580644   18020 main.go:468] ciao-cli FATAL: Expected HTTP response code [202 204] when accessing [DELETE https://singlevm:8776/v2/af6e94d5fe7c456f94ad886b5e677dae/volumes/36d61091-9063-4fea-8360-80047b5a8511], but got 500 instead
		Internal Server Error
		null
		goroutine 1 [running]:
		github.com/01org/ciao/vendor/github.com/golang/glog.stacks(0x983400, 0xc400000000, 0x110, 0x290)
			/home/rob/src/go/src/github.com/01org/ciao/vendor/github.com/golang/glog/glog.go:769 +0xa5
		github.com/01org/ciao/vendor/github.com/golang/glog.(*loggingT).output(0x967080, 0xc400000003, 0xc420320300, 0x941257, 0x7, 0x1d4, 0x0)
			/home/rob/src/go/src/github.com/01org/ciao/vendor/github.com/golang/glog/glog.go:720 +0x357
		github.com/01org/ciao/vendor/github.com/golang/glog.(*loggingT).printDepth(0x967080, 0xc400000003, 0x1, 0xc420141e88, 0x1, 0x1)
			/home/rob/src/go/src/github.com/01org/ciao/vendor/github.com/golang/glog/glog.go:646 +0x126
		github.com/01org/ciao/vendor/github.com/golang/glog.FatalDepth(0x1, 0xc420141e88, 0x1, 0x1)
			/home/rob/src/go/src/github.com/01org/ciao/vendor/github.com/golang/glog/glog.go:1134 +0x5d
		main.fatalf(0xc420314ee0, 0xd4, 0x0, 0x0, 0x0)
			/home/rob/src/go/src/github.com/01org/ciao/ciao-cli/main.go:132 +0x126
		main.main()
			/home/rob/src/go/src/github.com/01org/ciao/ciao-cli/main.go:468 +0x14d

```

rbradford:
@kaccardi I believe you might have seen this issue before?

[...]

PR #1019, opened 2017-01-17
[NOT FOR MERGING] ciao-launcher: Only unmap explicitly mapped volumes and improve logging

rbradford:
The cfg.Volumes slice contains all the volumes used by the instance -
both those that have been attached after boot and those specified on the
qemu command line.

The volumes specified on the command line will be unmapped by qemu when
it is terminated and so those do not need to be explicitly unmapped.

This patch stores a boolean in the volumeConfig struct indicating if the
volume should be unmapped.

### DIFF
--- a/ciao-launcher/attachvolume_instance.go
+++ b/ciao-launcher/attachvolume_instance.go
@@ -81,7 +81,7 @@ func processAttachVolume(storageDriver storage.BlockDriver, monitorCh chan inter
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


rbradford:
Debugging #1009 showed up this issue with unmapping already unmapped volumes.

rbradford:
@markdryan ready for review now with the alternative approach we discussed.

markdryan:
> currently trying to verify this but looking at the code this commit will break containers

I was able to verify this.  The volume is still mapped on the compute node after the container to which it is attached is deleted.

PR #1020, opened 2017-01-17
ciao-launcher: Wait for virtualizer to terminate

rbradford:
After requesting that the virtualizer (e.g. qemu) to stop wait on the
channel monitoring the domain socket used to communicate with the
virtualizer.  This allows us to identify when the virtualizer has
finished and the next step of the cleanup can begin.

This addresses #1009 by not trying to unmap volumes that are still in
use by the running qemu process.

Fixes: #1009


rbradford:
This fix was @markdryan's idea.

Issue #988, opened 2017-01-05
remove ImageID from workload definition


kaccardi:
Image ID should be removed and instead the storage section of the workload definition should specify the source type.

PR #1028, opened 2017-01-18
Fix #1027 - workload yaml parsing

### LINKED ISSUE #1027: Cannot create workloads with image source
Modifying the .yaml files to specify an image source for the storage does not work.

rbradford:
Fix the parsing of workload storage specified in the .yaml file. When then transition from yaml to json happened the disks struct was omitted resulting the yaml files being incorrectly parsed.

This PR fixes the tags on the struct and updates the examples to explicitly use the source type.

A further PR will remove the image_id functionality as per #988


tpepper:
Looks good to me.

Issue #1026, opened 2017-01-18
Workload verification needs improving


rbradford:
Needs to check that there is at least one bootable volume for VMs

Needs to check that if there is a source id that there is a source type.

PR #1030, opened 2017-01-18
Enhance workload storage verification

rbradford:
Enhance the verification code used to verify storage specification in the workload definition at both a ciao-cli and ciao-controller level.

As part of this PR ciao-controller now expects and explicit source type which ciao-cli will now provide.

Fixes: #1026

### DIFF
--- a/ciao-cli/workload.go
+++ b/ciao-cli/workload.go
@@ -200,6 +200,10 @@ func optToReq(opt workloadOptions, req *types.Workload) error {
 			Ephemeral: disk.Ephemeral,
 		}
 
+		if disk.Source != nil && disk.Source.Type == "" {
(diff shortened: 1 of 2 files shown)


tpepper:
LGTM aside from the gocyclo complaint.

kaccardi:
@rbradford does the README.md in ciao-cli/examples need to be updated to reflect this requirement, or is this change transparent to the user?

rbradford:
@kaccardi not with *this* PR. I have another branch that deletes the image_id functionality. This just exercises what was already there and fixes it.

rbradford:
@kaccardi yes this change is transparent to the user. It will in fact improve the user experience considerably with better error reporting.

PR #1032, opened 2017-01-19
Networking: cnci-agent: Persist network state across reboots/migration

mcastelino:
Add support to the CNCI agent to save the state of the CNCI
network setup to the boot disk. The network state is recovered
by replaying the commands. This allows for potential migration
to a new implemenation of the command if the cnci agent is
upgraded.

This allows the CNCI agent to recover its state if the CNCI
is rebooted or migrated to a different network node.

Includes a test case where the CNCI is rebooted by sshing into it
as a stand in for migration, and the network setup is re validated.

Also added a commit to use the errors package.

### DIFF
--- a/networking/ciao-cnci-agent/client.go
+++ b/networking/ciao-cnci-agent/client.go
@@ -92,6 +92,7 @@ func (s *ssntpConn) setStatus(status bool) {
 
 type agentClient struct {
 	ssntpConn
+	db    *cnciDatabase
 	cmdCh chan *cmdWrapper
 	netCh chan struct{} //Used to signal physical network changes
 }
@@ -190,6 +191,7 @@ func processCommand(client *ssntpConn, cmd *cmdWrapper) {
 			c := &netCmd.TenantRemoved
 			glog.Infof("Processing: CiaoEventTenantRemoved %v", c)
 			err := delRemoteSubnet(c)
+
 			if err != nil {
 				glog.Errorf("Error Processing: CiaoEventTenantRemoved %v", err)
 			}
@@ -261,6 +263,8 @@ func (client *agentClient) CommandNotify(cmd ssntp.Command, frame *ssntp.Frame) 
 				return
 			}
 			glog.Infof("EVENT: ssntp.AssignPublicIP %v", assignIP)
+
+			dbProcessCommand(client.db, &assignIP)
 			client.cmdCh <- &cmdWrapper{&assignIP}
 		}(payload)
 
@@ -275,6 +279,8 @@ func (client *agentClient) CommandNotify(cmd ssntp.Command, frame *ssntp.Frame) 
 				return
 			}
 			glog.Infof("EVENT: ssntp.ReleasePublicIP %s", releaseIP)
+
+			dbProcessCommand(client.db, &releaseIP)
 			client.cmdCh <- &cmdWrapper{&releaseIP}
 		}(payload)
 
@@ -299,6 +305,7 @@ func (client *agentClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
 			}
 			glog.Infof("EVENT: ssntp.TenantAdded %s", tenantAdded)
 
+			dbProcessCommand(client.db, &tenantAdded)
 			client.cmdCh <- &cmdWrapper{&tenantAdded}
 		}(payload)
 
@@ -313,6 +320,8 @@ func (client *agentClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
 				return
 			}
 			glog.Infof("EVENT: ssntp.TenantRemoved %s", tenantRemoved)
+
+			dbProcessCommand(client.db, &tenantRemoved)
 			client.cmdCh <- &cmdWrapper{&tenantRemoved}
 		}(payload)
 
@@ -321,7 +330,7 @@ func (client *agentClient) EventNotify(event ssntp.Event, frame *ssntp.Frame) {
 	}
 }
 
-func connectToServer(doneCh chan struct{}, statusCh chan struct{}) {
+func connectToServer(db *cnciDatabase, doneCh chan struct{}, statusCh chan struct{}) {
 
 	defer func() {
 		statusCh <- struct{}{}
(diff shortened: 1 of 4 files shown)


mcastelino:
@markdryan This PR adds basic support required in networking for CNCI migration.

This does not address the CNCI upgrade scenario. To support this database should ideally be stored in a data volume. The simplest way would be to have /var/lib/ciao on the CNCI be mounted on a data volume.

networking/ciao-cnci-agent/database.go
```diff
@@ -17,36 +17,196 @@
...
+}
+
+func dbInit() (*cnciDatabase, error) {
+	db := &cnciDatabase{}
+	db.DbProvider = database.NewBoltDBProvider()
+	db.SubnetMap.m = make(map[string]*payloads.TenantAddedEvent)
+	db.PublicIPMap.m = make(map[string]*payloads.PublicIPCommand)
+
+	if err := db.DbInit(dbCfg.DataDir, dbCfg.DbFile); err != nil {
+		return nil, err
```
rbradford: How about using the new errors work:

e.g. return nil, errors.Wrap(err, "error initialising database")

networking/ciao-cnci-agent/database.go
```diff
@@ -17,36 +17,196 @@
...
+	}
+	if err := db.DbTableRebuild(&db.SubnetMap); err != nil {
+		return nil, err
```
rbradford: return nil, errors.Wrap(err, "error rebuilding database tables for subnet map")

networking/ciao-cnci-agent/database.go
```diff
@@ -17,36 +17,196 @@
...
+	}
+	if err := db.DbTableRebuild(&db.PublicIPMap); err != nil {
+		return nil, err
```
rbradford: return nil, errors.Wrap(err, "error rebuilding database tables for public IP map")

networking/ciao-cnci-agent/database.go
```diff
@@ -17,36 +17,196 @@
 package main
 
 import (
+	"fmt"
+	"sync"
+	"time"
+
```
rbradford: For the errors stuff below you'll need to import "github.com/pkg/errors"

networking/ciao-cnci-agent/client.go
```diff
@@ -420,6 +429,39 @@ func discoverUUID() (string, error) {
...
+	defer db.SubnetMap.Unlock()
+	db.PublicIPMap.Lock()
+	defer db.PublicIPMap.Unlock()
+
+	for key, subnet := range db.SubnetMap.m {
+		glog.Infof("Key: %v Subnet: %v", key, subnet)
+		err := addRemoteSubnet(subnet)
+		if err != nil {
+			lastError = err
+			glog.Errorf("rebuildNetworkState: %v", err)
```
rbradford: Would be good to wrap and drop the glog here - and then make sure it's printed at the calling point.

mcastelino:
@rbradford the cnci agent is small enough that I can move the entire code base to use the errors package. Let me do that in one more commit as part of this PR.

networking/ciao-cnci-agent/client.go
mcastelino: @rbradford Here I try to make forward progress even on error. This means that the CNCI will try to rebuild itself the best it can even if some of the rebuild may fail. Hence I only return the lastError in this function. In the caller of this function, I am considering making forward progress even if the rebuild fails.

networking/ciao-cnci-agent/client.go
```diff
@@ -420,6 +429,39 @@ func discoverUUID() (string, error) {
...
 }
 
+//Rebuild network state from database
+func rebuildNetworkState(db *cnciDatabase) error {
+	var lastError error
+	if db == nil {
+		return nil
+	}
+
+	db.SubnetMap.Lock()
```
markdryan: I guess the locks are really needed here but there's no problem keeping them.

markdryan:
@mcastelino What's the best way to test this on Single VM?

mcastelino:
@markdryan We need the add back the support to ssh into the CNCI to be able to test this in single VM. I initially tested this manually. Another options is to use sshpass which I am not a big fan of.
https://linux.die.net/man/1/sshpass

However in the case of ciao-down having and using sshpass may not be too bad?

@kaccardi Can we add support in controller in the short term to support ssh key provisioning only in the case of single VM. We already have the flag right?

networking/ciao-cnci-agent/client.go
mcastelino: @markdryan locks are not really needed here. But I was just being paranoid.

kaccardi:
@mcastelino since the workloads PR has not been merged yet, the ssh support has not yet been removed. Yes, I can modify the controller to read the cnci ssh key off the disk in the case of singlevm at start up time. We do still have the "singlemachine" parameter - it's unused. I can also use it for cnci configuration until we get the API for limits done and the algorithm for calculating the size of the cnci.

mcastelino:
@rbradford I moved the entire codebase of the CNCI Agent to use errors. The stack trace being informative resulted in some simplification of logging on pure error paths as the trace will provide enough context.

mcastelino:
@kaccardi in that case I will add a test case for CNCI state restore testing in single VM using a manual reboot of the CNCI.

mcastelino:
@markdryan Added a test case to automate this test on Single VM. Rebooting the CNCI VM after sshing into it. In the future this can be replaced by a stop. kill launcher, start it with a new UUID and start CNCI to approximate the migration.

networking/ciao-cnci-agent/client.go
rbradford: @mcastelino Ack, in which case this is great! If the error doesn't need any more context then there is no need for a wrap.

markdryan:
@mcastelino Actually, I realized I could test this without ssh.  On single VM launcher is build with the debug tag, so I was able to connect to the cnci with netcat.  I connected with netcat, and did a 

ip link | grep alias
    alias br_172.16.0.0+24
    alias gre_172.16.0.0+24##198.51.100.1

I then rebooted the CNCI instance and executed the same commands

ip link | grep alias
    alias br_172.16.0.0+24
    alias gre_172.16.0.0+24##198.51.100.1

So this all looks good.

markdryan:
verify.sh is passing for me as well.

mcastelino:
@markdryan I had forgotten about the netcat option to reboot. I wonder if logging in via netcat can be automated. It will open up lots of new test previously non testable scenarios for us.

PR #1037, opened 2017-01-19
openstack: Reduce number of Keystone requests

obedmr:
This commit reduces 2 Keystone service calls per request.
Basically, it will initialize token-based client information at the
moment a request comes to the service and will add it to the main
handler to be re-used on all the identity corresponding custo functions.

At the end, it boosts the performance with the removal of non-required API calls.

Below is an example of Keystone service logs with a simple instance listing call ``ciao-cli instance list``.

```
# Previous Code
➜  docker-ciao git:(master) ✗ docker-compose exec  ciao-keystone tail -f /var/log/nginx-access.log -n 0
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "POST /v3/auth/tokens HTTP/2.0" 201 719 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "POST /v3/auth/tokens HTTP/2.0" 201 2638 "-" "gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2635 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2634 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2635 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:26:28 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2635 "-""gophercloud/1.0.0"

# New Code

➜  docker-ciao git:(master) ✗
➜  docker-ciao git:(master) ✗ docker-compose exec  ciao-keystone tail -f /var/log/nginx-access.log -n 0
172.18.0.1 - - [19/Jan/2017:18:27:03 +0000] "POST /v3/auth/tokens HTTP/2.0" 201 718 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:27:03 +0000] "POST /v3/auth/tokens HTTP/2.0" 201 2638 "-" "gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:27:03 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2635 "-""gophercloud/1.0.0"
172.18.0.1 - - [19/Jan/2017:18:27:03 +0000] "GET /v3/auth/tokens HTTP/2.0" 200 2635 "-""gophercloud/1.0.0"
```

### DIFF
--- a/openstack/identity/identity.go
+++ b/openstack/identity/identity.go
@@ -146,11 +146,9 @@ func (r getResult) extractRoles() (*Roles, error) {
 // validateServices
 // Validates that a given user belonging to a tenant
 // can access a service specified by its type and name.
-func validateService(client *gophercloud.ServiceClient, token string, tenantID string, serviceType string, serviceName string) bool {
-	r := v3tokens.Get(client, token)
-	result := getResult{r}
+func validateService(tokenResult getResult, tenantID string, serviceType string, serviceName string) bool {
 
-	p, err := result.extractProject()
+	p, err := tokenResult.extractProject()
 	if err != nil {
 		return false
 	}
@@ -160,7 +158,7 @@ func validateService(client *gophercloud.ServiceClient, token string, tenantID s
 		return false
 	}
 
-	services, err := result.extractServices()
+	services, err := tokenResult.extractServices()
 	if err != nil {
 		return false
 	}
@@ -180,10 +178,8 @@ func validateService(client *gophercloud.ServiceClient, token string, tenantID s
 	return false
 }
 
-func validateProjectRole(client *gophercloud.ServiceClient, token string, project string, role string) bool {
-	r := v3tokens.Get(client, token)
-	result := getResult{r}
-	p, err := result.extractProject()
+func validateProjectRole(tokenResult getResult, project string, role string) bool {
+	p, err := tokenResult.extractProject()
 	if err != nil {
 		return false
 	}
@@ -192,7 +188,7 @@ func validateProjectRole(client *gophercloud.ServiceClient, token string, projec
 		return false
 	}
 
-	roles, err := result.extractRoles()
+	roles, err := tokenResult.extractRoles()
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


openstack/identity/identity.go
```diff
@@ -283,16 +277,24 @@ type ValidAdmin struct {
...
+	tokenGetResult getResult
 }
 
 // ServeHTTP satisfies the http handler interface.
 // It will check to make sure that the api caller is validated with
 // keystone before allowing the next handler in the chain to be called.
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+
+	token := r.Header["X-Auth-Token"]
+	if token != nil {
```
markdryan: This should be

if len(token) > 0

There's no guarantee that a non nil slice contains any elements.

openstack/identity/identity.go
```diff
@@ -283,16 +277,24 @@ type ValidAdmin struct {
...
+		res := v3tokens.Get(h.Client, token[0])
+		h.tokenGetResult = getResult{res}
```
markdryan: Isn't this racy?  Isn't each invocation of ServeHTTP called in a different goroutine?

markdryan: To be honest I don't really understand how all this identity stuff works but I'm concerned about what would happen if we have multiple clients.  Do they send different X-Auth-Token values in their request?  If so, and we have one handler serving these multiple clients, aren't we creating a security problem here, different clients could use each other's tokens.  This would be the case even if we fixed the race above using mutexes.  If you want to cache the results of v3tokens.Get you'll need to do it in a local variable and pass it to other functions via a parameter.

openstack/identity/identity.go
obedmr: @markdryan, yeah, you're right, good catch, thanks, applied.

openstack/identity/identity.go
```diff
@@ -237,15 +233,19 @@ func (h Handler) validateToken(r *http.Request) bool {
 		return false
 	}
 
+	if len(token) > 0 {
```
markdryan: This doesn't look right.  Does it even compile?

obedmr: @markdryan  I removed that part of the code.

obedmr:
@markdryan, you're right, this code needed some re-thinking, I send that getResult variable by parameters to those functions that needed it indead of putting it in the Handler structure. Now, it should be safer.

markdryan:
> you're right, this code needed some re-thinking, I send that getResult variable by parameters to those functions that needed it indead of putting it in the Handler structure. Now, it should be safer.

Great.  That will address my concerns.

openstack/identity/identity.go
obedmr: @markdryan, interesting, good catch, I missed that thing. /me fixing it.

markdryan:
@kaccardi Is this okay for you?  It passes BAT and verify.sh.

markdryan:
@kaccardi Have you had a chance to look at this one yet?  We need your approval to merge.

kaccardi:
@markdryan looks good to me. @obedmr thanks for the improvement.

markdryan:
@obedmr The code is fine and I've tested with BAT and verify.sh.  However, I don't think the commit message matches the code anymore.

> Basically, it will initialize token-based client information at the
moment a request comes to the service and will add it to the main
handler to be re-used on all the identity corresponding custo functions.

It doesn't work like this anymore does it?

As soon as the commit message is updated I'll merge.

obedmr:
@markdryan updated, thanks

PR #1045, opened 2017-01-20
ciao-controller: fix bug in workload validation

kaccardi:
Fix a bug in the workload validation that improperly returned
BadRequest when Storage.ID == "" and Storage.SourceType != Empty.
What we really wanted to do was check if Storage.SourceID was "".


markdryan:
verify.sh and BAT pass fine, although I'm not sure they exercise this code.

PR #1049, opened 2017-01-23
controller: Fix race condition in confirmTenant

### LINKED ISSUE #1007: ciao-webui breaks ciao
I came across this issue when trying to integrate the webui into ciao-down.  I guess the underlying issue must be in ciao (not the webui), but I can't figure out what's going wrong.  It's easy to reproduce

git checkout -b markdryan-singlevm-webui master
git pull https://github.com/markdryan/ciao.git singlevm-webui

cd $GOPATH/src/github.com/01org/ciao/testutil/ciao-down
go build
./ciao-down prepare

wait for 10 minutes

./ciao-down connect

Follow the instructions for running setup.sh

Start a new terminal

Follow the instructions for starting the webui

From a browser on your host, enter https://localhost:3000

log in as csr.  The password is hello

Create an instance.  The instance will not be created.

Return to the terminal from which you ran setup.sh and run the BAT tests.  Many tests will fail.

And here's the weird piece.

Shutdown the single VM (using cleanup.sh) and the web-ui
Run setup.sh again
Run the BAT tests.  They'll pass
Run the webui and create an instance.  This will work now.

Looking at the logs controller is getting confused and is creating multiple CNCIs.  Not sure why.

markdryan:
This commit ensures that confirmTenant is only successfully called once
for each tenant during any invocation of controller.  The problem is
that confirmTenant is not re-entrant and yet was called by multiple
go routines simulataneously.  This mean that a ciao cluster could be
broken by two commands issued simultaneously on a new tenant.

The tricky thing about fixing this bug is that we cannot really protect
the whole function with a mutex.  The function is potentially very slow
as it optionally needs to wait for a CNCI to launch.  Using a critical
section for the entire function would result in commands for existing
tenants being blocked while a CNCI was launched for a new tenant, which
wouldn't be great.

The fix works by introducing tenant specific channels that provide
synchronisation points for individual tenants.  These channels are
stored in a global map.  This map requires a lock, but the lock only
needs to be held for a short period of time, the time taken to
read or write something to the map.  Thus the act of launching a
CNCI on a tenant blocks only commands on that tenant and not commands
destined for other tenants.

The fix assumes that tenants do not get deleted, and if they do, that
their UUIDs do not get re-used.

Fixes #1007

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -94,7 +94,7 @@ func (c *controller) deleteInstance(instanceID string) error {
 	return nil
 }
 
-func (c *controller) confirmTenant(tenantID string) error {
+func (c *controller) confirmTenantRaw(tenantID string) error {
 	tenant, err := c.ds.GetTenant(tenantID)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


markdryan:
@kaccardi I'm assuming here that we don't actually need to call confirmTenant for each invocation of a command.  We just need to be sure that it has already been successfully called for this tenant at some point.

rbradford:
@markdryan this looks good. Just wondering if we can reuse Datastore.cnciAddedChans as that should already be signalling on the slow step. But that might require a bigger refactor to make it works...

markdryan:
Just updated the commit to correct some grammatical errors with the comments and the commit message.

markdryan:
I'm going to merge this as is as it fixes a major issue with the web-ui.  We can look at reworking the fix to improve re-use at a later stage if needed.

PR #1059, opened 2017-01-25
Updates to ciao-down

markdryan:
There are two updates:

1. Fix a couple of race conditions that prevent the prepare command from removing an instance in case of error or cancellation.
2. Add a new option, -runcmd, that can be used to provide an additional set of instructions for cloud-init to run when setting up the vm.  These instructions augment the setup work that ciao-down does to setup ciao itself.

### DIFF
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -117,8 +117,14 @@ func downloadProgress(p progress) {
 }
 
 func prepare(ctx context.Context, errCh chan error) {
+	var err error
+
+	defer func() {
+		errCh <- err
+	}()
+
 	if !hostSupportsNestedKVM() {
-		errCh <- fmt.Errorf("nested KVM is not enabled.  Please enable and try again")
+		err = fmt.Errorf("nested KVM is not enabled.  Please enable and try again")
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


markdryan:
@mcastelino Could you review and merge this one if you're okay with it?

mcastelino:
LGTM

markdryan:
@mcastelino I don't see the comments.  I'll merge this now but feel free to enter the comments in another issue.

mcastelino:
@markdryan I missed submitting the comments.. they were sitting in draft

testutil/ciao-down/ciao_down.go
```diff
@@ -77,24 +77,25 @@ func checkDirectory(dir string) error {
 	return nil
 }
 
-func prepareFlags() (memGB int, CPUs int, debug bool, uiPath string, err error) {
+func prepareFlags() (memGB int, CPUs int, debug bool, uiPath, runCmd string, err error) {
 	fs := flag.NewFlagSet("prepare", flag.ExitOnError)
 	vmFlags(fs, &memGB, &CPUs)
 	fs.BoolVar(&debug, "debug", false, "Enables debug mode")
 	fs.StringVar(&uiPath, "ui-path", "", "Host path of cloned ciao-webui repo")
+	fs.StringVar(&runCmd, "runcmd", "", "Path to a file containing additional commands to execute when preparing the VM")
```
markdryan: Yes.  I'll do this.  I also need to document and improve the language used to create these cloud-init files.  RIght now it's a bit odd.  The other thing we need is control over the ports that are opened up.  My guess is you don't want ports 3000 and 5000 opened up on CC.  I'll enter some issues for these.

testutil/ciao-down/ciao_down.go
```diff
@@ -117,28 +117,32 @@ func downloadProgress(p progress) {
 }
 
 func prepare(ctx context.Context, errCh chan error) {
+	var err error
+
+	defer func() {
+		errCh <- err
```
markdryan: Currently, we only send error down the channel so there's no need to close it.  The reason we use this model is so we can multiplex the result of the command with the CTRL-C signal.

PR #1070, opened 2017-01-26
Singlevm archlinux support

dorileo:
A few bits to support singlevm(bare metal) on archlinux host systems.

### DIFF
--- a/testutil/singlevm/setup.sh
+++ b/testutil/singlevm/setup.sh
@@ -332,11 +332,24 @@ if [ x"$cacert_prog_ubuntu" != x ] && [ -x "$cacert_prog_ubuntu" ]; then
     # Do it a second time with nothing new to make it clean out the old
     sudo "$cacert_prog_ubuntu" --fresh
 elif [ x"$cacert_prog_fedora" != x ] && [ -x "$cacert_prog_fedora" ]; then
-    cacert_dir=/etc/pki/ca-trust/source/anchors/
+    cacert_dir_fedora=/etc/pki/ca-trust/source/anchors/
+    cacert_dir_archlinux=/etc/ca-certificates/trust-source/anchors
+    cacert_dir=""
+
+    if [ -d "$cacert_dir_fedora" ]; then
+	cacert_dir=$cacert_dir_fedora
+    elif [ -d "$cacert_dir_archlinux" ]; then
+	cacert_dir=$cacert_dir_archlinux
+    fi
+    
     if [ -d "$cacert_dir" ]; then
         sudo install -m 0644 -t "$cacert_dir" "$keystone_cert"
         sudo install -m 0644 -t "$cacert_dir" "$CIAO_CA_CERT_FILE"
         sudo "$cacert_prog_fedora" extract
+    else
+	echo "Unable to add keystone's CA certificate to your system's trusted \
+             store!"
+	exit 1
     fi
 else
     echo "Unable to add keystone's CA certificate to your system's trusted \


testutil/singlevm/setup.sh
```diff
@@ -332,11 +332,23 @@ if [ x"$cacert_prog_ubuntu" != x ] && [ -x "$cacert_prog_ubuntu" ]; then
     # Do it a second time with nothing new to make it clean out the old
     sudo "$cacert_prog_ubuntu" --fresh
 elif [ x"$cacert_prog_fedora" != x ] && [ -x "$cacert_prog_fedora" ]; then
-    cacert_dir=/etc/pki/ca-trust/source/anchors/
+    cacert_dir_fedora=/etc/pki/ca-trust/source/anchors/
```
mcastelino: so archlinux gets detected as fedora and then your demux it. Looks good.

ciao-image/datastore/store.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/01org/ciao/clogger/gloginterface"
 	"github.com/01org/ciao/database"
 	"github.com/01org/ciao/openstack/image"
+	"github.com/golang/glog"
 )
```
obedmr: For logging, I think we should start using the  Ciao [clogger](https://github.com/01org/ciao/tree/master/clogger).  You could take the database subproject as an example https://github.com/01org/ciao/tree/master/database

testutil/singlevm/setup.sh
```diff
@@ -332,11 +332,23 @@ if [ x"$cacert_prog_ubuntu" != x ] && [ -x "$cacert_prog_ubuntu" ]; then
...
+    cacert_dir_archlinux=/etc/ca-certificates/trust-source/anchors
+    cacert_dir=""
+
+    if [ -d "$cacert_dir_fedora" ]; then
+	cacert_dir=$cacert_dir_fedora
+    elif [ -d "$cacert_dir_archlinux" ]; then
+	cacert_dir=$cacert_dir_archlinux
```
obedmr: The Archlinux validation lgtm

testutil/singlevm/setup.sh
```diff
@@ -332,11 +332,23 @@ if [ x"$cacert_prog_ubuntu" != x ] && [ -x "$cacert_prog_ubuntu" ]; then
...
+    fi
+    
     if [ -d "$cacert_dir" ]; then
         sudo install -m 0644 -t "$cacert_dir" "$keystone_cert"
         sudo install -m 0644 -t "$cacert_dir" "$CIAO_CA_CERT_FILE"
         sudo "$cacert_prog_fedora" extract
+    else
+	echo "Unable to add keystone's CA certificate to your system's trusted \
+             store!"
```
markdryan: Don't we need an exit 1 here?

dorileo: @markdryan sure

dorileo: done

ciao-image/datastore/store.go
```diff
@@ -178,11 +178,14 @@ func (s *ImageStore) UploadImage(ID string, body io.Reader) error {
 	if s.rawDs != nil {
 		err = s.rawDs.Write(ID, body)
 		if err != nil {
+			database.Logger.Errorf("Could not write image: %v", err)
+			// TODO shouldn't we return err here?
```
markdryan: @dorileo well spotted but actually, I think it's the block of code below that should not have a return.  By returning it means the the image is not persisted.  The image will exist with state set to killed but then if you reboot the controller I think the image will dissapear, and the rbd image will be leaked.

ciao-image/datastore/store.go
```diff
@@ -178,11 +178,14 @@ func (s *ImageStore) UploadImage(ID string, body io.Reader) error {
...
 			img.State = Killed
 		}
 
 		img.Size, err = s.rawDs.GetImageSize(ID)
 		if err != nil {
+			database.Logger.Errorf("Could not get image size: %v", err)
 			img.State = Killed
 			return err
```
markdryan: I think this code should be 

```
if err != nil {
  database.Logger.Errorf("Could not get image size: %v", err)
  img.State = Killed
  _ s.rawDS.Delete(ID)
}
```

@obedmr, @rbradford Could you comment on this?

rbradford: Yes we need to delete it from the ceph if we've marked it as Killed here otherwise we'll leak the rbd image.

dorileo: Hi @markdryan and @rbradford I can add a patch to fix that on the series. thanks.

ciao-image/datastore/store.go
dorileo: Ok, the code comment has achieved its goal, removed on new version. Thanks.

markdryan:
@obedmr  Just need your approval to merge.

rbradford:
This might be pedantic but I really think the two changes to ciao-image should be in a separate PR.

dorileo:
Series broken as requested by @rbradford, the ciao-image patches were included in the PR #1119.

Issue #1065, opened 2017-01-26
singlevm: single machine mode fails with tls issues


kaccardi:
You can no longer run testutil/singlevm/setup.sh directly on your dev machine (single machine mode) without ciao-down due to tls issues.

kaccardi:
E0125 11:38:30.867236   94702 authenticate.go:43] ciao-cli ERROR: Could not get AuthenticatedClient Post https://<valid hostname deleted by me>:5000/v3/auth/tokens: tls: oversized record received with length 24932
E0125 11:38:30.869251   94702 identity.go:126] ciao-cli ERROR: Failed to create an AuthenticatedClient
E0125 11:38:30.872775   94702 main.go:235] ciao-cli ERROR: Could not send HTTP request Get https://<valid hostname deleted by me>:5000/v3/users//projects: tls: oversized record received with length 24932
F0125 11:38:30.872799   94702 main.go:429] ciao-cli FATAL: Get https://<valid hostname deleted by me>:5000/v3/users//projects: tls: oversized record received with length 24932

kaccardi:
I removed the valid hostname from these urls manually - they previously did contain the correct hostname.

tpepper:
In ```Get https://:5000/v3/users//projects``` should there be data between "users/" and "/projects"?

tpepper:
Ignore that prior comment.  I believe it's a separate https://github.com/01org/ciao/issues/1068

markdryan:
@kaccardi This should be fixed now.  Could you verify and close?

PR #1071, opened 2017-01-26
Networking: Add explicit allow rules for Ciao Traffic

### LINKED ISSUE #1064: ciao-down fails ssh connectivity tests
verify.sh fails with a fresh ciao-down image. Unclear as to why.

mcastelino:
Latest version of docker (1.13+) adds a default deny rule
for all forwarded traffic if docker enables ip forwarding.

If docker already finds ip forwarding enabled on the system
it will leave the firewall setup untouched and only append
rules at the very top.

The ciao traffic that traverses the bridges gets impacted by
this as the bridge traffic is also subject to iptables rules
as the following rules are enabled

bridge-nf-call-arptables
bridge-nf-call-iptables
bridge-nf-call-ip6tables

To fix this 
- Add explicit allow rules for Single VM Ciao Traffic
- Add explicit allow rules for Ciao traffic on compute nodes.

This also fixes the TLS issues observed in the bare metal single machine setup

Fixes #1064 #1065 

### DIFF
--- a/networking/libsnnet/cn.go
+++ b/networking/libsnnet/cn.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/coreos/go-iptables/iptables"
 	"github.com/vishvananda/netlink"
 )
 
@@ -178,6 +179,8 @@ type ComputeNode struct {
 	//simultaneously certain netlink calls suffer higher latencies
 	APITimeout time.Duration
 
+	*iptables.IPTables
+
 	*cnTopology
 	apiThrottleSem chan int
 }
@@ -291,6 +294,12 @@ func (cn *ComputeNode) Init() error {
 	}
 
 	cn.cnTopology = newCnTopology()
+
+	ipt, err := iptables.New()
+	if err != nil {
+		return fmt.Errorf("Unable to setup iptables %v", err)
+	}
+	cn.IPTables = ipt
 
 	return nil
 }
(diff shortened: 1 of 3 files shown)


mrkz:
@mcastelino should this also be addressed for the ciao deployments using docker 1.13+ ?

mcastelino:
@mrkz I am working on a fix inside our code to add explicit firewall rules to allow ciao internal traffic. That should ideally fix the issue without any explicit change/reboot etc. Hope to have that PR out today. Or maybe even add to this PR.

mcastelino:
@mrkz With this change, there should be no change required to the deployment scripts.

tpepper:
LGTM.  Will you squash these when the 3rd patch completes CI?

mcastelino:
@tpepper Yes.. will squash them

markdryan:
I think I'm going to fix the docker version in ciao-down.  This is the second time docker has broken ciao in the last few month.  A bug in docker 1.12.3 also required a launcher change.  I'll wait a little bit to verify everything is working fine with 1.13.0 and if it is I'll fix the version.  We can update it from time to time once we're happy new releases are stable.

tpepper:
@markdryan regarding https://github.com/01org/ciao/pull/1071#issuecomment-275627842 do you think a CI run and BAT are sufficient currently as the basis of a broadly generic acceptance test for a new version of docker?

If we're freezing, we should also enforce that in the ansible scripts and note current preferred version in documentation.

markdryan:
That's what I was thinking.  The reason I made this comment is that I'd just contributed to Manohar's PR to adding the following comment about ciao-down:

> sets up an environment in which Ciao is known to work seamlessly.

I later ran verify.sh only to find out that ciao wasn't working seamlessly at all.

PR #1077, opened 2017-01-27
Quotas and Limits Phase 1

rbradford:
Here is a working prototype for quotas and limits (the only thing i've not published here is the ciao-cli support as i'm still tidying that up)

The core of my design for this feature is based around a Quota Service (QS) that handles the in-memory storing of quotas/limits and the appropriate usage.

The main entry points to the API that the QS provides to other parts of controller is to Consume() or to Release() sets of payloads.RequestedResource structs.

An example of how this works can be seen in the "ciao-controller: Add quota checking for shared storage" commit.

The QS is designed so that it is safe to be accessed from multiple goroutines without using mutexes and also the API for consuming is asynchronous around a channel to allow e.g. instances to be started and then shutdown later if the quota is exceeded.

The sqlite database is used only to persist the quotas and is read from at startup and updated when the API is used to update the quota (the QS value for the quota is updated at the same time.)

One thing that this patch set does not yet cover is populating the QS with existing volumes that already exist when controller starts up.

### DIFF
--- a/_release/bat/quotas_bat/quotas_bat_test.go
+++ b/_release/bat/quotas_bat/quotas_bat_test.go
@@ -1,0 +1,98 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// Package quotasbat is a placeholder package for the basic BAT tests.
+package quotasbat
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/01org/ciao/bat"
+)
+
+const standardTimeout = time.Second * 300
+
+func findQuota(qds []bat.QuotaDetails, name string) *bat.QuotaDetails {
+	for i := range qds {
+		if qds[i].Name == name {
+			return &qds[i]
+		}
+	}
+	return nil
+}
+
+func restoreQuotas(ctx context.Context, tenantID string, origQuotas []bat.QuotaDetails, currentQuotas []bat.QuotaDetails) error {
+	for i := range currentQuotas {
+		qd := findQuota(origQuotas, currentQuotas[i].Name)
+		if qd != nil && qd.Value != currentQuotas[i].Value {
+			err := bat.UpdateQuota(ctx, "", tenantID, qd.Name, qd.Value)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// Test getting and setting quotas
+//
+// Tests retrieving and setting quotas.
+//
+// Gets the current quotas, sets several, gets them again checks they've
+// changed, restores the original and checks the restoration.
+func TestQuotas(t *testing.T) {
+	qn := "tenant-vcpu-quota"
+	ctx, cancelFunc := context.WithTimeout(context.Background(), standardTimeout)
+	defer cancelFunc()
+
+	tenants, err := bat.GetUserTenants(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(tenants) < 1 {
+		t.Fatal("Expected user to have access to at least one tenant")
+	}
+
+	tenantID := tenants[0].ID
+	origQuotas, err := bat.ListQuotas(ctx, tenantID, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = bat.UpdateQuota(ctx, "", tenantID, qn, "10")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	updatedQuotas, err := bat.ListQuotas(ctx, tenantID, "")
+	if err != nil {
+		t.Error(err)
+	}
+
+	qd := findQuota(updatedQuotas, qn)
+	if qd.Value != "10" {
+		t.Error("Quota not expected value")
+	}
+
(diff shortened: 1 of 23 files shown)


rbradford:
@kaccardi it would be great if you could look at this PR to give me some feedback on whether you're happy with the approach. I'm not looking for code feedback at this time as there still needs to be more unit test coverage and i'm sure I can simplify some of the ways the internal data is stored.

I acknowledge it's quite different to the way we've done things in the past in Ciao but I can see a way forward to move some other bits over to this model for example I think our volume and storage code would benefit from some consolidation into a mini service like this.

kaccardi:
@rbradford love it! I think it's a very good approach, works with enough abstraction to make this easy to use and uncluttered, but not so much that we risk moving off into over architected, module hell.

rbradford:
I think it's best to split the work into separate PRs that tackle implementing the quotas for different resource types.

This PR covers the core functionality of the quota service / API / datastore / ciao-cli support plus (new) limiting for volume creation.

I'm proposing handling quota controls for instances and images in a separate PR.

The remaining work on this PR:

- BAT test coverage for general CLI
- BAT test coverage for ephemeral volume creation/specific volume creation
- further enhancing controller unit test coverage

rbradford:
BAT testing is now in so this is ready for review:

@kaccardi for controller changes
@markdryan for bat changes / bug fixes

bat/quotas.go
```diff
@@ -0,0 +1,58 @@
...
+// the specified tenant will be returned. Otherwise it will provide the quotas
+// for the current tenant.
+func ListQuotas(ctx context.Context, tenantID string, forTenantID string) ([]QuotaDetails, error) {
+	var qds []QuotaDetails
+	var err error
+	if forTenantID == "" {
+		args := []string{"quotas", "list", "-f", "{{tojson .}}"}
+		err = RunCIAOCLIJS(ctx, tenantID, args, &qds)
+	} else {
+		args := []string{"quotas", "list", "-for-tenant", tenantID, "-f", "{{tojson .}}"}
```
markdryan: Should this be forTenantID?

_release/bat/quotas_bat/quotas_bat_test.go
```diff
@@ -0,0 +1,93 @@
...
+		if qds[i].Name == name {
+			return &qds[i]
+		}
+	}
+	return nil
+}
+
+func restoreQuotas(ctx context.Context, tenantID string, origQuotas []bat.QuotaDetails, currentQuotas []bat.QuotaDetails) error {
+	for i := range currentQuotas {
+		qd := findQuota(origQuotas, currentQuotas[i].Name)
```
markdryan: Do we need to compare qd against nil or are we guaranteed that findQuota will succeed?

_release/bat/quotas_bat/quotas_bat_test.go
```diff
@@ -0,0 +1,93 @@
...
+func TestQuotas(t *testing.T) {
+	qn := "tenant-vcpu-quota"
+	ctx, cancelFunc := context.WithTimeout(context.Background(), standardTimeout)
+	defer cancelFunc()
+
+	tenants, err := bat.GetUserTenants(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tenantID := tenants[0].ID
```
markdryan: Are we guaranteed to have at least one user tenant?

markdryan:
Could you update the directory tree at the top of _release/bat/README.md?

ciao-cli/quotas.go
```diff
@@ -0,0 +1,218 @@
...
+// on the privilege level of user. Check privilege, then
+// if not privileged, build non-privileged URL.
+func (cmd *quotasListCommand) run(args []string) error {
+	url, err := getCiaoQuotasResource()
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	if cmd.tenantID != "" {
+		url = fmt.Sprintf("%s/%s/quotas", url, cmd.tenantID)
```
kaccardi: I guess if a user is not privileged, this call will result in a 404 error. It might be a bit confusing. I wonder if we should check privilege ahead of time in this case so that a better error message could be generated.

ciao-controller/api/api.go
```diff
@@ -461,6 +479,41 @@ func addWorkload(c *Context, w http.ResponseWriter, r *http.Request) (Response,
...
+
+	var req types.QuotaUpdateRequest
+	err = json.Unmarshal(body, &req)
+	if err != nil {
+		return errorResponse(err), err
+	}
+
+	c.UpdateQuotas(tenantID, req.Quotas)
+
+	return Response{http.StatusCreated, nil}, nil
```
kaccardi: with http.StatusCreated, you would not want to return an empty response. While it is not *required* to return a response with Created, it is standard practice and you never know what clients are going to expect. Here you need to either return a list of the quotas you updated, or change the response code to Accepted or NoContent. Since Accepted implies you are not done working on it, it's probably not 'acceptable' here :). IMO you should return the list of updated quotas.

ciao-controller/quotas.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/01org/ciao/ciao-controller/internal/quotas"
+	"github.com/01org/ciao/ciao-controller/types"
+	"github.com/01org/ciao/payloads"
+	"github.com/pkg/errors"
+)
+
+func (c *controller) UpdateQuotas(tenantID string, qds []types.QuotaDetails) {
+	c.qs.Update(tenantID, qds)
+	for i := range qds {
+		c.ds.UpdateQuota(tenantID, qds[i])
```
kaccardi: Can't this call to UpdateQuota fail? If so, you are not checking the return and could wind up with persistent data out of sync with quota service.

kaccardi: It seems to me that you should also propagate any datastore errors back to the caller

rbradford:
Thanks @markdryan  and @kaccardi for your feedback. I've addressed every issue.

@kaccardi RE the errors, i've used the transaction stuff in sqlite to handle that and also move the quota storage update to after so I think we can't get inconsistent data.

kaccardi:
I'm ready to merge if @markdryan is satisfied with your changes.

PR #1098, opened 2017-01-31
Enable node networks based scheduling

### LINKED ISSUE #1044: Ensure that restarted CNCI instances are rescheduled to an NN in the same network segment.
The scheduler need to ensure that the CNCI is rescheduled to an NN in the same network segment.  This can be done by checking the Compute Net IP of the Physical Node.  It should choose a NN which has an IP that is on the same Compute Subnet

tpepper:
This PR allows scheduling of a workload to a network node that satisfies a workload's request for specific network connectivity.

Partially addresses #1044, but completely closing that issue depends on addressing issue #912.  When controller is able to restart a CNCI, it will need to add a network resource request, which scheduler can honor as of this PR.

### DIFF
--- a/ciao-scheduler/scheduler.go
+++ b/ciao-scheduler/scheduler.go
@@ -93,6 +93,7 @@ type nodeStat struct {
 	load        int
 	cpus        int
 	isNetNode   bool
+	networks    []payloads.NetworkStat
 }
 
 type controllerStatus uint8
@@ -381,6 +382,7 @@ func (sched *ssntpSchedulerServer) updateNodeStat(node *nodeStat, status ssntp.S
 		node.diskAvailMB = stats.DiskAvailableMB
 		node.load = stats.Load
 		node.cpus = stats.CpusOnline
+		node.networks = stats.Networks
 
 		//any changes to the payloads.Ready struct should be
 		//accompanied by a change here
--- a/ciao-scheduler/scheduler.go
+++ b/ciao-scheduler/scheduler.go
@@ -426,19 +426,24 @@ type workResources struct {
 	memReqMB     int
 	diskReqMB    int
 	networkNode  bool
+	physNets     []string
 }
 
 func (sched *ssntpSchedulerServer) getWorkloadResources(work *payloads.Start) (workload workResources, err error) {
 	// loop the array to find resources
 	for idx := range work.Start.RequestedResources {
+		reqType := work.Start.RequestedResources[idx].Type
+		reqValue := work.Start.RequestedResources[idx].Value
+		reqString := work.Start.RequestedResources[idx].ValueString
+
 		// memory:
-		if work.Start.RequestedResources[idx].Type == payloads.MemMB {
-			workload.memReqMB = work.Start.RequestedResources[idx].Value
+		if reqType == payloads.MemMB {
+			workload.memReqMB = reqValue
 		}
 
 		// network node
-		if work.Start.RequestedResources[idx].Type == payloads.NetworkNode {
-			wantsNetworkNode := work.Start.RequestedResources[idx].Value
+		if reqType == payloads.NetworkNode {
+			wantsNetworkNode := reqValue
 
 			// validate input: requested resource values are always integers
 			if wantsNetworkNode != 0 && wantsNetworkNode != 1 {
@@ -450,6 +455,14 @@ func (sched *ssntpSchedulerServer) getWorkloadResources(work *payloads.Start) (w
 				workload.networkNode = true
 			} else { //wantsNetworkNode == 0
 				workload.networkNode = false
+			}
+
+		}
+
+		// network node physical networks
+		if workload.networkNode {
+			if reqType == payloads.PhysicalNetwork {
+				workload.physNets = append(workload.physNets, reqString)
 			}
 		}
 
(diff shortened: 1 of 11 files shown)


tpepper:
@markdryan and @mcastelino looking for early review on this

ciao-scheduler/scheduler.go
```diff
@@ -697,10 +709,10 @@ func startWorkload(sched *ssntpSchedulerServer, controllerUUID string, payload [
 
 	var targetNode *nodeStat
 
-	if workload.networkNode == 0 {
-		targetNode = pickComputeNode(sched, controllerUUID, &workload)
-	} else { //workload.network_node == 1
+	if workload.networkNode {
 		targetNode = sched.pickNetworkNode(controllerUUID, &workload)
+	} else { //workload.network_node == false
```
mcastelino: @tpepper un-related to this patch. But why do we have a method in one place and function in the other?

ciao-scheduler/scheduler.go
```diff
@@ -477,12 +490,35 @@ func (sched *ssntpSchedulerServer) getWorkloadResources(work *payloads.Start) (w
...
 
+func networkDemandsSatisfied(node *nodeStat, workload *workResources) bool {
+	if node.isNetNode {
+		var matchedNetworksCount int
+		var requestedNetworksCount int
+
+		for _, requestedNetwork := range workload.physNets {
+			requestedNetworksCount++
+			for _, availableNetwork := range node.networks {
+				if requestedNetwork == availableNetwork.NodeIP {
```
mcastelino: @tpepper we may want to provide a generic satisfies method for non integer matches. As the comparisons may get complicated in the future. Here for example a subnet that is a subnet of the demanded subnet may also be ok.

payloads/start.go
```diff
@@ -127,6 +127,9 @@ type RequestedResource struct {
 	// Value specifies the integer value associated with that resource.
 	Value int `yaml:"value"`
 
+	// ValueString is an optional string format value instead of integer
+	ValueString string `yaml:"value_string,omitempty"`
```
mcastelino: @tpepper does it make sense to add a satisfies method to things like strings that are not directly comparable. basically create a new type here with an associated method?

payloads/ready.go
```diff
@@ -45,6 +45,10 @@ type Ready struct {
 	// cpu[0-9]+ entries in /proc/stat.
 	CpusOnline int `yaml:"cpus_online"`
 
+	// Array containing one entry for each network interface present on the
+	// CN/NN
+	Networks []NetworkStat
```
tpepper: I'm not sure management makes sense to track.  It would make sense to have the launcher not report any/all networks, but rather just the compute networks.  I'm not sure if that's already what it does or if it reports any/all networks.  @markdryan what all is in the data member I grabbed out of controller?

ciao-scheduler/scheduler.go
tpepper: Yeah it's all slowly headed that direction.  If we would have had the foresight, it would have made sense to start this as a pretty and infinitely extensible interface.  In the meantime though I'm still biasing towards really basic elementary implementation.  I'd like to see a few more use cases come through before we decide we know what the core of the interface should be.

payloads/ready.go
mcastelino: @tpepper The launcher today only reports compute networks. The networking library provides both to launcher.

ciao-scheduler/scheduler.go
tpepper: This is related to unit test.  Until now the pickNetworkNode() was so trivial that I didn't export it in export_test.go for use in scheduler_internal_test.go.  The function route is required for the export to work.  So pickNetworkNode() will also shift to being a function instead of a method so that it can see isolated "go test" and "go bench" based testing yet also remain unexported (ie: we've deliberately chosen to not have an arbitrarily pluggable scheduler).

payloads/start.go
tpepper: Yes it probably will.  I thought about this as I was doing this simple first string which is directly comparable.  But I want to wait for the specific instance of non-comparable data to have a proof point of what a suitable interface would need to cover.  I don't want to architect too much without that usage data, because the interface ends up being an API that could need versioned or paint us into a corner.  We want to be able to add optional implementations of resource requests but always keep existing workload definitions functional.  I don't want resource A using version 1 of the comparison interface and resource B using version 2 of the comparison interface.  Plus I feel like I could make an argument for always in the future choosing to articulate resource requests as a list of requests which are each key:singlevalue pairs and the values always being comparable, but I'm also expecting some use case to come along and prove me wrong.

mcastelino: @tpepper YAML does support Marshall and Unmarshall of custom types by implementing the interface https://github.com/go-yaml/yaml/blob/v2/yaml.go#L31

However how that interacts with versioning I am not sure. As the message may be versioned, but the type may not be versioned, unless we start implementing top level Marshaling support

ciao-scheduler/scheduler.go
```diff
@@ -453,22 +483,42 @@ func (sched *ssntpSchedulerServer) getWorkloadResources(work *payloads.Start) (w
...
-	}
 
 	// note the uuid
 	workload.instanceUUID = work.Start.InstanceUUID
 
 	return workload, nil
 }
 
+func networkDemandsSatisfied(node *nodeStat, workload *workResources) bool {
+	if node.isNetNode {
```
markdryan: Stylistically it might be nice to change this test to 

```
if !node.isNetNode {
    return true
}
```

Doing so would allow you to get rid of one level of indentation.

tpepper: Ah yes..will do!

tpepper: Fixed!

rbradford:
Passes BAT & verify.sh on SingleVM.

PR #1101, opened 2017-02-01
ciao-cli: auto-generate template documentation

markdryan:
This PR adds some functions that can be used to auto-generate the documentation that describes how templates may be used for ciao-cli commands.  As, an example, to generate the template usage documentation for a command that operates on slices of  CiaoCNCIs you might call
```    
generateUsageUndecorated([]types.CiaoCNCI{})
```
and you would receive a string that looks something like this.
```    
    []struct {
        ID        string `json:"id"`
        TenantID  string `json:"tenant_id"`
        IPv4      string `json:"IPv4"`
        Geography string `json:"geography"`
        Subnets   []struct {
                Subnet string `json:"subnet_cidr"`
        } `json:"subnets"`
    }
```

There's another function, generateUsageDecorated that also includes information about all the additional functions ciao-cli provides for use in templates.

The PR also updates ciao-cli to use these new functions removing all of the hand coded template documentation.

### DIFF
--- a/ciao-cli/template.go
+++ b/ciao-cli/template.go
@@ -217,23 +217,22 @@ func createTemplate(name, tmplSrc string) *template.Template {
 }
 
 func exportedFields(typ reflect.Type) bool {
-	var i int
-	for i = 0; i < typ.NumField(); i++ {
+	for i := 0; i < typ.NumField(); i++ {
 		if typ.Field(i).PkgPath == "" {
-			break
+			return true
 		}
 	}
 
-	return i < typ.NumField()
+	return false
 }
 
 func ignoreKind(kind reflect.Kind) bool {
 	return (kind == reflect.Chan) || (kind == reflect.Invalid)
 }
 
-func generateStruct(buf io.Writer, dummy reflect.Type) {
-	for i := 0; i < dummy.NumField(); i++ {
-		field := dummy.Field(i)
+func generateStruct(buf io.Writer, typ reflect.Type) {
+	for i := 0; i < typ.NumField(); i++ {
+		field := typ.Field(i)
 		if field.PkgPath != "" || ignoreKind(field.Type.Kind()) {
 			continue
 		}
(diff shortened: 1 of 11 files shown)


ciao-cli/template.go
```diff
@@ -213,3 +215,120 @@ func createTemplate(name, tmplSrc string) *template.Template {
...
+		li := strings.LastIndex(pkg, "/")
+		if li != -1 {
+			pkg = pkg[li+1:]
+		}
+		pkg = fmt.Sprintf("%s.", pkg)
+	}
+	return fmt.Sprintf("%s%s", pkg, typ.Name())
+}
+
+func generateUsage(buf io.Writer, dummy reflect.Type, tag string) {
```
tpepper: Is the use of the variable name "dummy" idiomatic here?  If not, is there something we could use that would be more descriptive?

ciao-cli/template.go
```diff
@@ -213,3 +215,120 @@ func createTemplate(name, tmplSrc string) *template.Template {
 
 	return t
 }
+
+func exportedFields(typ reflect.Type) bool {
+	var i int
+	for i = 0; i < typ.NumField(); i++ {
+		if typ.Field(i).PkgPath == "" {
+			break
```
mcastelino: why not just return true here? and false at the end

markdryan: Just an idiom I used in C for years.  I'll change it as you've suggested.

ciao-cli/template.go
markdryan: The name dummy was chosen to reflect the fact that the value between passed to the function exists only to convey type information.  It's value is unimportant.  The only other example of a function like this i can find in the standard library is reflect.TypeOf, whose parameter is

i interface{}

I'll change the code to match.

markdryan: Code has been changed.  I use i for interface{} and typ for reflect.Type to match the reflect package.

ciao-cli/template.go
markdryan: Code updated.  Thanks.

PR #1103, opened 2017-02-01
ansible: Fix some issues with the README.md

markdryan:
This commit fixes two errors with the README.md file that prevent one
from installing ciao via ansible.


rbradford:
lgtm.

I don't understand why we bother with a container when it's privileged and has host networking too..

PR #1118, opened 2017-02-03
ciao-down: Fix docker version to 1.12.6

markdryan:
The latest version of docker seems to break the ceph/demo container.
This commit fixes the docker version at 1.12.6 which is known to
work well with ciao.

### DIFF
--- a/testutil/ciao-down/cloudinit.go
+++ b/testutil/ciao-down/cloudinit.go
@@ -112,7 +112,7 @@ runcmd:
  - {{template "CHECK" .}}
 
  - curl -X PUT -d "Installing Docker" 10.0.2.2:{{.HTTPServerPort}}
- - {{template "PROXIES" .}}apt-get install docker-engine -y
+ - {{template "PROXIES" .}}apt-get install docker-engine=1.12.6-0~ubuntu-xenial -y
  - {{template "CHECK" .}}
 
  - curl -X PUT -d "Installing GCC" 10.0.2.2:{{.HTTPServerPort}}


mcastelino:
@markdryan before we downgrade can we perform one check with docker 1.13
In the VM on bootup in the cloud-init set

sysctl -w net.ipv4.ip_forward=1

Also make this permanent

echo "net.ipv4.ip_forward=1" | sudo tee -a /etc/sysctl.conf

This will eliminate some of the docker iptables issues.

markdryan:
@mcastelino Yes, I'll try this on Monday.  If this works I'll update the PR to fix the docker release to 1.13.0.

mcastelino:
@markdryan if that works, we probably need to add more iptable rules in single VM to allow the traffic flows we need. But I will take care of that outside of this PR. I assume this issue does not impact a real cluster.

markdryan:
Okay, so it might be easier just to merge this now, while we work on a proper fix.  Otherwise ciao-down/singlevm won't work for people in the meantime.

mcastelino:
@markdryan yes.. Let us downgrade for now

markdryan:
Okay, so I just created a new ciao-down VM with docker 1.13.0 and this time the tests passed.

markdryan:
Here's the error I was getting

```
F0203 16:05:42.164590   16393 instance.go:510] ciao-cli FATAL: HTTP Error [500] for [POST https://singlevm:8774/v2.1/f7df638e2a204cb5bbe80bdc352cccd1/servers]: {"error":{"code":500,"name":"Internal Server Error","message":"Error when running: [rbd --id ciao clone 73a86d7e-93c0-480e-9c41-ab42f69b7799@ciao-image 1dff1d9b-1977-4595-babc-9d9e39a07629]: exit status 22: 2017-02-03 16:05:42.148107 7fcbfe790100 -1 librbd: couldn't set parent: (22) Invalid argument\nrbd: clone error: (22) Invalid argument\n"}}
null
```

markdryan:
Seeing as how the tests are passing with docker 1.13.0, fixing at 1.12.6 is probably not the right thing to do.

kaccardi:
@markdryan did you want to close this PR then?

markdryan:
@kaccardi I'm going to upgrade the PR to fix the version to 1.13.1 which works well.  I should probably also update it to fix ansible installed docker at the same version.

mcastelino:
@markdryan will you be fixing the version to 1.13.1

markdryan:
@mcastelino Actually, the version numbers and the repos seem to have changed (#1221).  I'll probably try the latest version of the new CE repos and if that works fix on that.  How does that sound?

mcastelino:
@markdryan makes sense.to switch to the CE repos. I need to fix the version to 1.12.1 for clear containers too.

markdryan:
This PR is no longer valid.  Abandoning.

PR #1119, opened 2017-02-03
Ciao image small fixes

dorileo:
These patches were initially part of another series reviewed on PR #1070. The PR was split after @rbradford requests.

The two patches are minor fixes and debugging log message inclusion ciao-image. Their descriptions are very self contained.


rbradford:
Requires @obedmr approval as image owner.

ciao-image/datastore/store.go
```diff
@@ -178,11 +178,13 @@ func (s *ImageStore) UploadImage(ID string, body io.Reader) error {
 	if s.rawDs != nil {
 		err = s.rawDs.Write(ID, body)
 		if err != nil {
+			database.Logger.Errorf("Could not write image: %v", err)
```
obedmr: I think we should have our own Logger in ``ciao-image``. According to what is defined [here](https://github.com/01org/ciao/tree/master/clogger) each component needs to implement its Logger. For this ``ciao-image`` we could reuse the already glog backend interface that clogger is proposing at https://github.com/01org/ciao/tree/master/clogger/gloginterface

Here are some examples how this clogger is being implemented:
- [database package](https://github.com/01org/ciao/blob/master/database/logger.go)
- [networking/libsnnet package](https://github.com/01org/ciao/blob/master/networking/libsnnet/logger.go)

As far I can see those are the packages that are using this already. @sboeuf could you please comment on this? I saw you already used that clogger in those 2 packages. For ``ciao-image``, we should add a new ``logger.go``, right?

sboeuf: @obedmr yes you should do that. The reason why we did that was to avoid "import" issues with vendored version of glog in ciao and the upstream version of glog in another project using ciao.

rbradford: Ping...progress on this PR?

dorileo: hey @rbradford I'll work on that today or earlier tomorrow. thanks.

rbradford: Ping? This PR hasn't been touched in 9 days?

rbradford:
Ping? Is this PR is relevant?

obedmr:
@rbradford, not at this moment, I could take a look on next days and re-send a new PR.

tpepper:
Re-ping... @obedmr any more thoughts?

obedmr:
@tpepper I added the required changes, I tried to do a PR to @dorileo's branch and something wirhd happen on this PR. Anyway, it contains my change, would it be better to open a new PR? it seems like this this PR CI validations are not completed and there's no sign of trying to start them.

tpepper:
I'm not sure what's happened.  There's a green check in the github UI next to the last commit prior to the merge commit which makes it look like it ran partly.  @mcastelino did you successfully commit on somebody else's branch earlier?  Or is it better to just pull their branch, add on it, and PR the local result of that, closing the prior PR.

Issue #1093, opened 2017-01-30
ciao-image,ciao-deploy: Add image visibility support in ciao-deploy


obedmr:
When creating initial images in Ciao deployment, we need to to update Ciao deployment scripts for setting the proper tags and options for indicating each image's visibility.

Issue #1048, opened 2017-01-20
ciao-image: Add BAT testing for CIAO Image Service


obedmr:
We need to cover all ``ciao-image`` functionality in BAT.

obedmr:
Addressed with #1123

PR #1123, opened 2017-02-03
ciao-image: Add support for private, public and internal images

obedmr:
It's using a new metadata field called ``visibility`` which defines the access
restrictions that users and CIAO inself have when trying to access the images
catalog. Below are the 3 main types of visibility that are supported:

- Internal
      - CIAO cluster's internal images (e.g. cnci)
      - Only created by members of admin tenant/project
- Private
      - Default visibility for any created image
      - Only available in the tenant/project where it was created
- Public
      - Only created/deleted by members of admin tenant/project
      - Available in all tenants/projects

### DIFF
--- a/ciao-cli/image.go
+++ b/ciao-cli/image.go
@@ -28,6 +28,7 @@ import (
 	"github.com/rackspace/gophercloud/openstack"
 	"github.com/rackspace/gophercloud/openstack/imageservice/v2/images"
 	"github.com/rackspace/gophercloud/pagination"
+	"strings"
 )
 
 var imageCommand = &command{
@@ -40,12 +41,18 @@ var imageCommand = &command{
 }
 
 type imageAddCommand struct {
-	Flag     flag.FlagSet
-	name     string
-	id       string
-	file     string
-	template string
+	Flag       flag.FlagSet
+	name       string
+	id         string
+	file       string
+	template   string
+	tags       string
+	visibility string
 }
+
+const (
+	internalImage images.ImageVisibility = "internal"
+)
 
 func (cmd *imageAddCommand) usage(...string) {
 	fmt.Fprintf(os.Stderr, `usage: ciao-cli [options] image add [flags]
@@ -65,6 +72,9 @@ func (cmd *imageAddCommand) parseArgs(args []string) []string {
 	cmd.Flag.StringVar(&cmd.id, "id", "", "Image UUID")
 	cmd.Flag.StringVar(&cmd.file, "file", "", "Image file to upload")
 	cmd.Flag.StringVar(&cmd.template, "f", "", "Template used to format output")
+	cmd.Flag.StringVar(&cmd.visibility, "visibility", string(images.ImageVisibilityPrivate),
+		"Image visibility (internal,public,private)")
+	cmd.Flag.StringVar(&cmd.tags, "tag", "", "Image tags (comma separated)")
 	cmd.Flag.Usage = func() { cmd.usage() }
 	cmd.Flag.Parse(args)
 	return cmd.Flag.Args()
@@ -89,9 +99,23 @@ func (cmd *imageAddCommand) run(args []string) error {
 		fatalf("Could not get Image service client [%s]\n", err)
 	}
 
+	imageVisibility := images.ImageVisibilityPrivate
+	if cmd.visibility != "" {
+		imageVisibility = images.ImageVisibility(cmd.visibility)
+		switch imageVisibility {
+		case images.ImageVisibilityPublic, images.ImageVisibilityPrivate, internalImage:
+		default:
+			fatalf("Invalid image visibility [%v]", imageVisibility)
+		}
+	}
+
+	tags := strings.Split(cmd.tags, ",")
+
 	opts := images.CreateOpts{
-		Name: cmd.name,
-		ID:   cmd.id,
+		Name:       cmd.name,
+		ID:         cmd.id,
+		Visibility: &imageVisibility,
+		Tags:       tags,
 	}
 
 	image, err := images.Create(client, opts).Extract()
@@ -408,6 +432,8 @@ func dumpImage(i *images.Image) {
 	fmt.Printf("\tSize             [%d bytes]\n", i.SizeBytes)
 	fmt.Printf("\tUUID             [%s]\n", i.ID)
 	fmt.Printf("\tStatus           [%s]\n", i.Status)
+	fmt.Printf("\tVisibility       [%s]\n", i.Visibility)
+	fmt.Printf("\tTags             %v\n", i.Tags)
 	fmt.Printf("\tCreatedDate      [%s]\n", i.CreatedDate)
 }
 
--- a/ciao-controller/openstack_image.go
+++ b/ciao-controller/openstack_image.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"net/http"
 	"path/filepath"
+	"strings"
 	"time"
 
 	imageDatastore "github.com/01org/ciao/ciao-image/datastore"
@@ -37,10 +38,11 @@ type ImageService struct {
 }
 
 // CreateImage will create an empty image in the image datastore.
-func (is ImageService) CreateImage(req image.CreateImageRequest) (image.DefaultResponse, error) {
+func (is ImageService) CreateImage(tenantID string, req image.CreateImageRequest) (image.DefaultResponse, error) {
 	// create an ImageInfo struct and store it in our image
 	// datastore.
 	glog.Infof("Creating Image: %v", req.ID)
+
 	id := req.ID
 	if id == "" {
 		id = uuid.Generate().String()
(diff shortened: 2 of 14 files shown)


markdryan:
@obedmr You need to update singleVM to set the visibility of the standard images it creates, otherwise it cannot create any of the default workloads.

@albertomurillo Do we create images in the ansible scripts with ciao-cli image add?  If so, these will need to be uploaded as well.

ciao-controller/openstack_image.go
```diff
@@ -80,10 +85,10 @@ func (is ImageService) CreateImage(req image.CreateImageRequest) (image.DefaultR
 	return image.DefaultResponse{
 		Status:     image.Queued,
 		CreatedAt:  i.CreateTime,
-		Tags:       make([]string, 0),
+		Tags:       strings.Split(i.Tags, " "),
```
markdryan: Yes, I think I'm correct

```
Image #3
	Name             [Clear Linux 12620]
	Size             [907018240 bytes]
	UUID             [df3768da-31f5-4ba6-82f0-127a1a705169]
	Status           [active]
	Visibility       [public]
	Tags             [[]]
	CreatedDate      [2017-02-16T17:37:30.561028069Z]
```

Doesn't look right.

markdryan:
@obed What are the rules about creation and deletion of public images?

In your comments above you say that they can only be created by admin.  However, I was able to create one in single VM as the normal user.

Also who can delete public images?  Currently, it only seems to be the user that created the images.  Admin can see public images but he cannot delete them.  If he tries he gets a weird 404 error.

ciao-image/datastore/metads.go
```diff
@@ -27,39 +28,51 @@ type MetaDs struct {
 
 // Write is the metadata write implementation.
 func (m *MetaDs) Write(i Image) error {
-	err := m.DbAdd("images", i.ID, &i)
+	tenant := i.TenantID
+	if i.Visibility == imageAPI.Public {
+		tenant = string(imageAPI.Public)
```
markdryan: Should this be

image.Public

?

ciao-image/datastore/metads.go
```diff
@@ -27,39 +28,51 @@ type MetaDs struct {
...
+	return m.DbDelete(tenant, id)
 }
 
 // Get is the metadata get implementation.
-func (m *MetaDs) Get(ID string) (Image, error) {
+func (m *MetaDs) Get(tenant, ID string) (Image, error) {
 
 	imageTable := &ImageMap{}
-	img, err := m.DbGet("images", ID, imageTable)
+	img, err := m.DbGet(string(imageAPI.Public), ID, imageTable)
```
markdryan: Again should this be image.Public?

ciao-cli/image.go
```diff
@@ -65,6 +71,8 @@ func (cmd *imageAddCommand) parseArgs(args []string) []string {
 	cmd.Flag.StringVar(&cmd.id, "id", "", "Image UUID")
 	cmd.Flag.StringVar(&cmd.file, "file", "", "Image file to upload")
 	cmd.Flag.StringVar(&cmd.template, "f", "", "Template used to format output")
+	cmd.Flag.BoolVar(&cmd.public, "public", false, "Public Image")
```
markdryan: I see, internal is overloaded.  It's used to indicate an internal image and for tags.  Is there any reason why we can't have tags on a public or private image?

ciao-cli/image.go
```diff
@@ -89,9 +97,26 @@ func (cmd *imageAddCommand) run(args []string) error {
...
+
+	imageVisibility := images.ImageVisibilityPrivate
+	if cmd.public {
+		imageVisibility = images.ImageVisibilityPublic
+	}
+
+	tags := []string{}
+	if cmd.internal != "" {
+		imageVisibility = internalImage
+		tags = append(tags, cmd.internal)
```
markdryan: Shouldn't we split the string?

ciao-cli/image.go
obedmr: @markdryan, agree, I just wanted to use tags for putting internal images information. But agree, let's allow public and private images have tags.
Regarding visibility flag, I like it. /me changing it

ciao-controller/openstack_image.go
```diff
@@ -50,23 +52,26 @@ func (is ImageService) CreateImage(req image.CreateImageRequest) (image.DefaultR
 			return image.DefaultResponse{}, image.ErrBadUUID
 		}
 
-		img, err := is.ds.GetImage(id)
-		if err != nil {
+		img, err := is.ds.GetImage(tenantID, id)
+		if err != nil && img != (imageDatastore.Image{}) {
```
obedmr: @markdryan good catch, thanks

ciao-controller/openstack_image.go
obedmr: sure, /me fixing it. thanks

ciao-image/datastore/datastore_test.go
```diff
@@ -28,11 +29,60 @@ var metaDsTables = []string{"images"}
...
+func createTmpFile(t *testing.T) *os.File {
+	content := []byte("temporary file's content")
+	tmpfile, err := ioutil.TempFile("", "example")
+	if err != nil {
+		t.Fatal(err)
+		return nil
+	}
+
+	if _, err := tmpfile.Write(content); err != nil {
+		t.Fatal(err)
```
obedmr: yeah, good catch, fixed

ciao-image/datastore/metads.go
obedmr: I added an alias ``imageAPI`` for image package. But, yeah, there's no need of that alias. /me changed it.

ciao-image/datastore/metads.go
obedmr: same approach as the previous comment. thanks

ciao-cli/image.go
obedmr: done

ciao-controller/openstack_image.go
```diff
@@ -114,16 +119,23 @@ func createImageResponse(img imageDatastore.Image) (image.DefaultResponse, error
...
 
-	images, err := is.ds.GetAllImages()
+	images, err := is.ds.GetAllImages(tenant)
 	if err != nil {
-		glog.Errorf("Error on listing images: %v", err)
+		glog.Errorf("Error on retrieving images from tenant [%v]: %v", tenant, err)
 		return response, err
 	}
 
+	publicImages, err := is.ds.GetAllImages(string(image.Public))
```
obedmr: yeah, actually, in database, there's a bucket called ``public`` which only contains public images. I'm thinking that I should also create another one for ``internal`` images, in order to make ``internal`` images retrieval faster.

obedmr:
@markdryan, I'm working on the missing coments, will update soon.

tpepper:
I'd like to see associated with this PR is a refresh of https://github.com/01org/ciao/wiki/Storage.

obedmr:
@markdryan answering your questions:

_In your comments above you say that they can only be created by admin. 
However, I was able to create one in single VM as the normal user._

**This is WIP, we'll be validating admin-privileges in order to create/delete public and internal images. And yes, only admins will be able to create and delete public images.**

_Also who can delete public images? Currently, it only seems to be the user 
that created the images. Admin can see public images but he cannot delete them. 
If he tries he gets a weird 404 error._

**Fixing it, it requires a tenant validation for delete process**

obedmr:
@markdryan  BAT, SingleVM and Deployment changes would be addressed on #1048, #1093 and #1225

obedmr:
@markdryan, current BAT shouldn't fail, it has been fixed. /me testing with singlevm.

obedmr:
singlevm is working.

obedmr:
@markdryan @kaccardi @rbradford  when have a change, could you take a look on this one?

markdryan:
@obedmr SingleVM and BAT are working for me now.  I did notice one odd thing though.

When I do a ciao-cli image list on singlevm it returns nothing.  I know the images are there as I can launch instances.   Is this expected?

obedmr:
@markdryan, are you loading the [~/local/demo.sh](https://github.com/01org/ciao/blob/master/testutil/singlevm/setup.sh#L60-L74) ? If so, demo user shouldn't have access to admin created images, they were private by default, in order to allow demo user to see this, those images should be public ``--visibility public``. I think that would be the required change in SingleVM with this new tenant-separated images approach.

obedmr:
/me adding the singlevm changes on this PR too.

markdryan:
>@markdryan, are you loading the ~/local/demo.sh ?

Yes exactly.

obedmr:
SingleVM changes were just added on a separate commit.

obedmr:
@markdryan, regarding the fact that you can start instances, it's because  workloads were defined with images that exist in ``admin`` tenant. We'll need a mechanism to validate workloads launching in order to verify that user from certain tenant is allowed to use workload's specified images.

@kaccardi, are workloads also tenant-separated? I mean, is there a concept of public or private workloads, if so, we could make sure that admin can create public workloads and private workloads will be the ones that can be only created in a specific tenant.

I think this can be addressed in a separate PR. @rbradford started an issue #1224  for addressing part of this issue.

obedmr:
@rbradford thanks for adding the quotas code.

rbradford:
Hi @obedmr please take a look at the changes I made for quota stuff and let me know if you're happy.

obedmr:
@rbradford sure, I just saw it and in LGTM. I'm happy with the change. Thanks for the help.

PR #1126, opened 2017-02-07
Ciao down fixes

markdryan:
This PR fixes a few small issues with ciao-down:

1. Values in the /etc/environment file were not quoted.
2. No limit was placed on the default values for CPU and memory usage.
3. ciao-down connect and ciao-down status did not work when ciao-down prepare was running.
4. apt-get upgrade was hanging.

### DIFF
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -325,14 +325,15 @@ func connect(ctx context.Context, errCh chan error) {
 		return
 	}
 
-	path, err := exec.LookPath("ssh")
+	_, err = os.Stat(ws.instanceDir)
 	if err != nil {
-		errCh <- fmt.Errorf("Unable to locate ssh binary")
+		errCh <- fmt.Errorf("instance does not exist")
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


mcastelino:
LGTM. Will test on confirm at my end

testutil/ciao-down/host_linux.go
```diff
@@ -32,11 +32,15 @@ func getMemAndCpus() (mem int, cpus int) {
 	cpus = getOnlineCPUs() / 2
 	if cpus < 0 {
 		cpus = 1
+	} else if cpus > 8 {
+		cpus = 8
 	}
```
rbradford: Github reviews don't let you comment on the commit message but you might want to s/MB/GB/ in the commit message.

testutil/ciao-down/ciao_down.go
```diff
@@ -325,14 +325,15 @@ func connect(ctx context.Context, errCh chan error) {
 		return
 	}
```
rbradford: In the commit message you'll want to change ./ciao to ./ciao-down

PR #1132, opened 2017-02-09
ansible: Use keystone:stable container

albertomurillo:
Use clearlinux/keystone:stable container rather than
clearlinux/keystone:latest


markdryan:
Matches what I did in ciao-down

PR #1133, opened 2017-02-09
update or remove template.go

tpepper:
We should either update the copyright date in template.go or remove the file completely.  If folks prefer the latter, just merge the PR.  If not, I'll drop the second patch and we can merge that.


kaccardi:
removing sounds like a good idea to me.

PR #1139, opened 2017-02-10
Controller fqdn from cert

tpepper:
This PR reverts the two patches which allow the FQDN to come in via a user configuration touch point with fallback to os.Hostname().  Instead it pulls the FQDN from the controller certificate, which ciao-cert populates.  If not present, it still falls back to os.Hostname().  Passes BAT and singlevm test automation.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -225,10 +225,7 @@ func main() {
 	wg.Add(1)
 	go ctl.startImageService()
 
-	host := clusterConfig.Configure.Controller.ControllerFQDN
-	if host == "" {
-		host, _ = os.Hostname()
-	}
+	host, _ := os.Hostname()
 	ctl.apiURL = fmt.Sprintf("https://%s:%d", host, controllerAPIPort)
 
 	wg.Add(1)
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -25,9 +25,12 @@ package main
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"sync"
@@ -225,7 +228,16 @@ func main() {
 	wg.Add(1)
 	go ctl.startImageService()
 
-	host, _ := os.Hostname()
+	host, err := getControllerFQDNFromCert(*cert)
+	if err != nil {
+		glog.Infof("Unable to determine hostname from controller certificate %s: %s", *cert, err)
+
+		host, err = os.Hostname()
+		if err != nil {
+			glog.Fatal("Unable to determine hostname from OS: ", err)
+			return
+		}
+	}
 	ctl.apiURL = fmt.Sprintf("https://%s:%d", host, controllerAPIPort)
 
 	wg.Add(1)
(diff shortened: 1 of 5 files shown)


ciao-controller/main.go
```diff
@@ -283,3 +295,33 @@ func (c *controller) startCiaoService() error {
...
+	}
+
+	parsedCert, err := x509.ParseCertificate(blockCert.Bytes)
+	if err != nil {
+		return "", fmt.Errorf("couldn't parse cert: %s", err)
+	}
+
+	role := ssntp.GetRoleFromOIDs(parsedCert.UnknownExtKeyUsage)
+	if !role.IsController() {
+		return "", fmt.Errorf("expected certificate role ssntp.Controller, got role %s", role.String())
```
rbradford: Your code is lgtm.

But i'm surprised you're taking it from the SSNTP cert. I guess that's fine but I would expected it to come from the HTTP cert that the server is actually served with. I think that should be present on clusterConfig.Configure.Controller. The only change should be for the path and to stop looking for the controller OID.

What do you think?

tpepper: I'd looked at a sample https cert from the controller web and was seeing slightly different data in it. Initially it was the one I was going to go for.  But I liked that the ssntp one enabled the sanity check of the role.  Now that I think about it though, the ssntp one is possibly...or even probably most desirably...on a different network and with a different name than the visible end-point name for the user access API's.  I think now this MUST use the https api web cert.

tpepper: Or not.  For that side of the house we have a CACert and a _key_.

The CA cert is only optionally created by us and in production I expect it would come from the cluster operator's IT certificate authority, not us.

The key is not a x509 struct.  I need to look into whether we could embed metadata in that...in general I suspect it's possible given my gpg key has a hunk of metadata that even includes a jpeg of me!

But even if we did a pivot to embed the desired information in this key, we'd still need to tweek the deploy automation to do this for that key in addition to having ciao-cert do it for the ssntp cert's.  AND we'd need to know that the embed works as an after the fact add-on, in case an operator's IT org hands them the key to use which was signed by the IT org's CAcert, not the operator creates the cert and gets it signed by the org's CAcert.  I think is something we must expect for a real production deployment.

Soooo...I think the ssntp cert is an easier route to go for now.  We're really just looking for a name and there is a name in the ssntp cert.  As long as the ssntp cert was created with reference to the external name of the controller, it can be a reliable source of the needed name.  Then we don't have to try to interact with the operator to coordinate modifications to a key that is signed by their org.

rbradford: I'm not super convinced that pulling it from the SSNTP cert is a good idea. The SSNTP communication could be on it's own private network with "internal" dns entries like cn01.ciao.internal, controller.ciao.internal, scheduler.ciao.internal and whereas controller's HTTP service would be on a more exposed network and name.

If we can't extract the hostname from the HTTP cert then i'm leaning towards:

1. Keeping the configuration entry or..
2. Removing any api where we need to expose absolute URIs to the client

tpepper: I don't like the added config option, but I think it's a required piece of data and symmetric to the CA and cert for compute in the config.

tpepper:
Superseded by PR 1142 (a minor documentation update).

PR #1144, opened 2017-02-13
openstack: Export extractProject functionality

### LINKED ISSUE #1082: ciao-image, identity: Make tenant ID available in ciao-image service.
In order to support private images, image service requires tenant ID information. This will include changes on Ciao identity layer and ciao-image service itself.

obedmr:
Some service API requests in OpenStack will not contain the project/tenant
in them. It requires a mechanism for retrieving project data and make
the proper treatment of requests.

All API requests will contain a tokenID, this information can be used in
order to extract the tenant/project from identity service.

Fixes #1082

### DIFF
--- a/openstack/identity/identity.go
+++ b/openstack/identity/identity.go
@@ -62,16 +62,15 @@ type Services struct {
 	Entries []ServiceEntry
 }
 
-type getResult struct {
-	v3tokens.GetResult
-}
+// GetResult adds behavior for supporting more token-information gathering
+type GetResult v3tokens.GetResult
 
-// extractProject
+// ExtractProject extracts project from token result structure
 // Ideally we would actually contribute this functionality
 // back to the gophercloud project, but for now we extend
 // their object to allow us to get project information out
 // of the response from the GET token validation request.
-func (r getResult) extractProject() (*Project, error) {
+func (r GetResult) ExtractProject() (*Project, error) {
 	if r.Err != nil {
 		glog.V(2).Info(r.Err)
 		return nil, r.Err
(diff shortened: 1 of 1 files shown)


rbradford:
@kaccardi needs to approve this.

kaccardi:
@obedmr I think I'd like to see how you'd use it inside the image service. I'm I guess a little nervous about this - mainly because I wonder if we are letting ourselves get too dependent on the keystone layer, and I wonder if there might be some other way to do this that might be more generic. It'd probably help if I saw the way you were going to use the function.

obedmr:
@kaccardi here's how image service would be using it. https://github.com/01org/ciao/pull/1123/files#diff-1fa51fda8b9d49d13b81a7985fcd7092R338

mrkz:
@kaccardi I'd also be needing this for the privilege check in the Configuration API checks in ciao-controller, so I asked obed to create a different PR for this, hence I'm also looking forward to agree on how we would implement this.

kaccardi:
@obedmr @mrkz i think I've got a way to do this that is a bit more generic - I'll post a PR in the next day or so.

kaccardi:
I submitted a PR #1199 for review which allows the identity middleware to report whether a user is privileged or not. I was thinking if we also need the tenantID, we can similarly have the identity service report the tenantID in a context var rather than having the lower layers mess around with tokens.

obedmr:
Hey @kaccardi, actually, we need the tenantID, it will allow the tenant-separated images, and the privileged user validation would help for public and internal images management.

obedmr:
Closed in favor of using #1199

PR #1146, opened 2017-02-14
ciao-down: Use apt section of cloudinit file

### LINKED ISSUE #1131: ciao-down should use apt sections of cloudinit files
The cloud-init file has a number of sections for configuring apt.  You can request an upgrade, setup new sources, setup proxies and install new packages.  We should probably change our cloudinit files to take advantage of these sections.  We probably want to continue installing our packages in runcmd so we can continue to retrieve progress information at the console.

markdryan:
The cloudinit file has a dedicated section for configuring apt but
we weren't using it.  Instead we were executing all of the apt
commands in the runcmd and write_files section of the cloud-init file.
This commit moves the instructions that set up the apt proxies and
initiate the apt-get upgrade to the apt section.  The other commands
such as setting up sources and installing packages can not be used
as these sections do not handle proxies correctly.

Fixes #1131

### DIFF
--- a/testutil/ciao-down/cloudinit.go
+++ b/testutil/ciao-down/cloudinit.go
@@ -35,9 +35,6 @@ mounts:
 write_files:
 {{- if len $.HTTPProxy }}
  - content: |
-     Acquire::http::Proxy "{{$.HTTPProxy}}";
-   path: /etc/apt/apt.conf
- - content: |
      [Service]
      Environment="HTTP_PROXY={{$.HTTPProxy}}"{{if len .HTTPSProxy}} "HTTPS_PROXY={{.HTTPSProxy}}{{end}}"{{if len .NoProxy}} "NO_PROXY={{.NoProxy}},singlevm{{end}}"
    path: /etc/systemd/system/docker.service.d/http-proxy.conf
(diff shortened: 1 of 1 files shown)


testutil/ciao-down/cloudinit.go
```diff
@@ -257,9 +259,6 @@ mounts:
 write_files:
 {{- if len $.HTTPProxy }}
  - content: |
-     Acquire::http::Proxy "{{$.HTTPProxy}}";
-   path: /etc/apt/apt.conf
- - content: |
```
rbradford: Can we keep this in....for post "ciao-down prepare" use of the image?

markdryan: We don't need it.

``
apt:
  proxy:
  https_proxy:
```

creates this file for us.  On my newly built VM with proxies configured I have

cat /etc/apt/apt.conf.d/90cloud-init-aptproxy

Acquire::http::Proxy "<my http proxy>";
Acquire::https::Proxy "<my https proxy>";

rbradford: Ninja.

PR #1150, opened 2017-02-15
ansible: Add support for Dual Nodes

### LINKED ISSUE #1081: ansible: Change ciao-launcher role to support dual node
Make ciao-launcher role install certificate for network nodes, compute nodes and dual nodes

### LINKED ISSUE #1091: ansible: Store certificates in an absolute path
Currently, certificates are stored in ./certificates/... which prevent the script from running
cli command from different directories.

albertomurillo:
This PR main fix is to allow a ciao node to appear under
network and compute sections of the hosts file
Fixes #1081

The other main fix from this PR is the addition of {{ workdir }}
defaulting to an absolute path
Fixes #1091


_DeploymentAndDistroPackaging/ansible/roles/keystone/templates/openrc.j2
```diff
@@ -7,6 +7,6 @@ export OS_TENANT_NAME=admin
 export OS_USERNAME=admin
 export OS_PASSWORD={{ keystone_admin_password }}
 export OS_AUTH_URL=https://{{ keystone_fqdn }}:35357/v3
-export OS_CACERT=certificates/keystone/keystone_cert.pem
+export OS_CACERT={{ workdir }}/keystone_cert.pem
```
obedmr: If we're installing certificates in OS, we shouldn't need this.

albertomurillo: We are installing the certificate on the controller node only.

This file is created on the deployment node and allows the user to run the cli on any machine without need to install the certificate

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/tasks/certificates.yml
```diff
@@ -43,6 +43,8 @@
         creates: cert-CNCIAgent-localhost.pem
       - role: controller
         creates: cert-Controller-localhost.pem
+      - role: agent,netagent
+        creates: cert-CNAgent-NetworkingAgent-localhost.pem
```
tpepper: Does this mean you will always create a dual role cert, in case it might be needed?

_DeploymentAndDistroPackaging/ansible/group_vars/all
```diff
@@ -1,4 +1,5 @@
 ---
+workdir: /tmp/ciao
```
tpepper: What are the permissions on the ciao directory if it is created?  What happens if it is already present and globally readable before the start of the ansible script?

Do you leak the private certificates to other users on the deployment machine?

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/tasks/certificates.yml
albertomurillo: Yes, the cert is created and stays on the deployment machine. It will only be deployed to the launcher node if that launcher node is present in both [networks] and [computes] sections

_DeploymentAndDistroPackaging/ansible/group_vars/all
tpepper: http://docs.ansible.com/ansible/file_module.html says the mode is optional and doesn't say what happens if it is not given.

albertomurillo: <b>What are the permissions on the ciao directory if it is created?</b>
Default permissions. I can add `mode=600` to make it only readable by the user running ansible.

<b> What happens if it is already present and globally readable before the start of the ansible script?</b>
If I use `mode=600` and the directory already exists, the permissions will be changed to 600

<b>Do you leak the private certificates to other users on the deployment machine?</b>
I didn't realize but yes.

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/tasks/create_cnci_image.yml
```diff
@@ -15,28 +15,28 @@
...
+    stat: path={{ workdir }}/.cnci_ready
     register: st
     changed_when: False
 
   - name: Download cnci image
     become: no
     get_url:
       url: "{{ cnci_image_url }}"
-      dest: images/{{ cnci_image }}.xz
+      dest: "{{ workdir }}/{{ cnci_image }}.xz"
```
tpepper: /tmp is often a tmpfs which means it is an ephemeral storage backed by RAM.  Many distros set it by default to be only a medium/small percentage of system RAM (eg: Ubuntu is 50% by default).  I worry storing the images there could amount to a DoS on the deployment machine by running it out of memory.

_DeploymentAndDistroPackaging/ansible/roles/ciao-launcher/tasks/install.yml
```diff
@@ -37,5 +37,14 @@
...
+    become: no
+    connection: local
+    get_url:
+      url: "{{ firmware_image_url }}"
+      dest: "{{ workdir }}/OVMF.fd"
+      validate_certs: no
+      force: no
+
   - name: Copy OVMF firmware
     copy: src={{ workdir }}/OVMF.fd dest=/usr/share/qemu/OVMF.fd
```
tpepper: Not related to this patch, but related to my comments in the other patch in this series:  what is the mode of the created /usr/share/qemu/OVMF.fd file?  I prefer an explicit 0644 over something implicit.

_DeploymentAndDistroPackaging/ansible/cleanup/ciao.yml
```diff
@@ -17,16 +17,11 @@
...
-        - ./admin-ciaorc
-        - ./demo-ciaorc
-        - ./clouds.yaml
-        - ./certificates
-        - ./images/.cnci_ready
-        - ./fetch
-        - /tmp/go
+      file: path={{ item }} state=absent
+      with_fileglob:
+        - /tmp/go/*
```
tpepper: This looks insecure.  If I'm on the deployment machine you're about to run on, and I place a rootkitted go in there ahead of you, you will inherit my damage.  Or?

_DeploymentAndDistroPackaging/ansible/roles/ciao-launcher/tasks/install.yml
albertomurillo: Agree with that but for all files not just this one.
I created https://github.com/01org/ciao/issues/1163 to track all files created by the deployment script and ensure they have correct permissions.

_DeploymentAndDistroPackaging/ansible/cleanup/ciao.yml
albertomurillo: I didnt get that

tpepper: Deploy host:  has no /tmp/go

Me:  mkdir /tmp/go
        put bad stuff in there

You: run ansible scripts
        use my bad stuff in /tmp/go

Is that a possible outcome?

albertomurillo: Ansible creates `{{ gopath }}` which defaults to /tmp/go if the directory doesn't exist.

The workflow is if you dont have `{{ gopath }}` created, ansible creates the directory.
Ansible pulls ciao from github with `go get -v` not `-u`.

This allows the admin to`cd {{gopath}}` `git checkout -b myitem` `hack hack`
run ansible and test results.

tpepper: Will ansible create it if it already exists?

albertomurillo: ansible does not create a directory if the directory already exists. thats how the [file](http://docs.ansible.com/ansible/file_module.html) module works

tpepper: So then a malicious user can pre-populate the world writable /tmp with bad data, and the ansible scripts will not overwrite the bad data and will use the bad data.

_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/tasks/certificates.yml
tpepper: This is bad from a security perspective.  It creates a broader attack surface for malicious cluster membership.  It should only be created if it is going to be used.

albertomurillo:
Im breaking this PR in smaller PR's

PR #1151, opened 2017-02-15
Fix Cancelling of QMP commands and some spelling mistakes

markdryan:
This PR fixes three issues:

1. Cancelling of QMP commands did not work properly.  The commands could be cancelled but commands issued after a cancelled command did not work correctly.  Rather than executing the new command the old cancelled command was re-executed.

2. The word Command was incorrectly spelt in the qmp_test.go file.

3. The code that handles cancelling is a little complicated and needed documenting.

### DIFF
--- a/qemu/qmp.go
+++ b/qemu/qmp.go
@@ -346,6 +346,39 @@ func (q *QMP) parseVersion(version []byte) *QMPVersion {
 	}
 }
 
+// The qemu package allows multiple QMP commands to be submitted concurrently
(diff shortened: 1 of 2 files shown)


markdryan:
I think there might be one use case I've missed.  I'm adding a new testcase to make sure.  Will update the PR soon.

markdryan:
I think I've fixed the issue correctly now but would like to add some more tests and comments.  Not ready for merge yet.

markdryan:
@sameo This is good to go now.

sameo:
LGTM

Issue #1114, opened 2017-02-02
Refactor resource usage for instances in datastore


rbradford:
As a precursor #1075 some refactoring is worthwhile on the way that the workload defaults are preserved in the datastore/database. They currently seem to be mapped to/from integers. 

I think it's best to wait until #991 lands as that touches much of the similar code.

PR #1156, opened 2017-02-16
Workload defaults refactor

rbradford:
This PR addresses #1114 and switches the workload_resources table, used to store the resources needed for a workload to not store the resource type as an integer but as a string. Removing conversion to & fro.


rbradford:
@kaccardi to review as controller owner.

tpepper:
While a reasonable clean up, this is a breaking change right?  Have we decided yet if we're going to manage those more actively or just break things and move fast indefinitely?

rbradford:
It's a database change yes. We have no schema migration yet. I think we would want that when we're close to being database stable - somewhere we're not quite yet.

tpepper:
I agree with @rbradford's assessment that we're not quite yet at the point of saying we have a stable db, so I'm ok with this.  Looking for @kaccardi review still.

PR #1159, opened 2017-02-16
singlevm: do not use service user "csr" in singlevm

kaccardi:
The service user "csr" is for the controller service. We should use a non-service
user for testing.


tpepper:
Looks good to me.  Addressed a pain point I've had with singlevm.

tpepper:
Passes BAT.

PR #1164, opened 2017-02-16
ansible: Allow dual node in ciao-launcher

### LINKED ISSUE #1081: ansible: Change ciao-launcher role to support dual node

albertomurillo:
If a node is in both [networks] and [computes] sections, start
launcher as dual node.

This allows the scripts to be used for singlevm

Fixes #1081


_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/tasks/certificates.yml
```diff
@@ -48,6 +48,8 @@
         creates: cert-CNCIAgent-localhost.pem
       - role: controller
         creates: cert-Controller-localhost.pem
+      - role: agent,netagent
+        creates: cert-CNAgent-NetworkingAgent-localhost.pem
```
tpepper: I still prefer this to be conditional.  If there is no dual role machine, this needlessly creates an extra certificate which would provide elevated privileges to the holder.

albertomurillo: There is no easy way to iterate the list of host and determine if there are dual nodes.
Also its better to have the cert creation in the same place since the cert is created once.

And lastly, while I agree that is a bit dirty to generate the file even if it is not going to be used... It doesnt mean it makes the deployment less secure. If an attacker somehows get access to that file, it will get access to the rest of the certificates anyways that are always generated.

tpepper:
As indicated above, I don't like the change, but Alberto's in charge of ansible.  Anybody else have opinions?  If not Alberto's the lead of this component.

markdryan:
@albertomurillo Is there really no way to do this?  It's possible to access the data in the hosts file I guess.  Is it then not possible to read the networks and compute sections into some sort of sets and do an intersection?

albertomurillo:
So after contemplating a different dirty ways to skip the creation of dual node certificates when there are no dual nodes in the hosts file. I found the `intersect` filter which provides a clean way to write this validation.

I have updated the PR and rebased.

PR #1166, opened 2017-02-16
ciao-controller: Remove non-required image client instance

obedmr:
The usage of ciao-image client is not necesessary for now,
it's only containing a value. It was useful when we had the
Posix raw datastore.

Ciao-image client requires a re-thinking in order to validate the
usability of this.


markdryan:
Looks okay to me.  Pending a review from @kaccardi.

PR #1182, opened 2017-02-20
ansible: Fix logging on systemd units

albertomurillo:
Fix systemd units for ciao-launcher, ciao-scheduler and ciao-controller
to log to the default location instead of stderr.


markdryan:
Works for me.

PR #1190, opened 2017-02-22
[RFR] systemd: Report to systemd when service has finished loading.

### LINKED ISSUE #1181: Ansible: Race condition in image upload
I did a cleanup and and tried to re-install my cluster.  The install failed with the following error:

```
TASK [ciao-controller : Upload "Clear Cloud"] **********************************
fatal: [ciao-ctl.ld.intel.com]: FAILED! => {"changed": true, "cmd": "/tmp/go/bin/ciao-cli image add --file images/clear-11960-cloud.img --name \"Clear Cloud\" --id df3768da-31f5-4ba6-82f0-127a1a705169", "delta": "0:00:00.261870", "end": "2017-02-20 09:37:15.844457", "failed": true, "rc": 255, "start": "2017-02-20 09:37:15.582587", "stderr": "F0220 09:37:15.841965   22124 image.go:99] ciao-cli FATAL: Could not create image [Post https://controller:9292/v2/images: dial tcp 10.103.239.202:9292: getsockopt: connection refused
...
```

I tried again and it worked.  It could be that the image service simply isn't ready when we try to upload the images.  This case is handled in SingleVM setup.sh by querying the image service in a loop until it is ready before uploading the images.  See https://github.com/01org/ciao/blob/master/testutil/singlevm/setup.sh#L703

albertomurillo:
By setting Type=notify in systemd unit files, systemd will wait until
the service has actually been started to return success or fail if the
service does not start correctly.

The previous behaviour was that systemctl start ciao-*.service will
return successfully and immediately regardless of the state of the
service.

This change also eliminates some race-condition where automaed deployment
scripts continue to make use of the api when the service is not fully
started.

Fixes #1181

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/templates/ciao-controller.service.j2
+++ b/_DeploymentAndDistroPackaging/ansible/roles/ciao-controller/templates/ciao-controller.service.j2
@@ -3,7 +3,7 @@ Description=CIAO Controller Service
 After=network.target
 
 [Service]
-Type=simple
+Type=notify
 ExecStart=/usr/local/bin/ciao-controller --cacert=/etc/pki/ciao/CAcert-{{ ciao_controller_fqdn }}.pem \
                                    --cert=/etc/pki/ciao/cert-Controller-localhost.pem \
                                    --v 3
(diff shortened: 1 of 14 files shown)


markdryan:
@albertomurillo Perhaps we could use socket activation instead for both scheduler and controller.

This is what the CC team are doing.  They sent me some links to how they use this stuff.

https://github.com/clearcontainers/proxy/blob/master/socket_activation.go
https://github.com/clearcontainers/proxy/blob/master/cc-proxy.socket.in
https://github.com/clearcontainers/proxy/blob/master/proxy.go#L289

tpepper:
I'm not sure socket activated makes sense.  That means scheduler starts when the first SSNTP packet comes in.  That seems late.  Similarly controller starting only at the point a user talks to its API for the first time seems later than desirable.

PR #1194, opened 2017-02-22
ciao-down: cache large files

### LINKED ISSUE #957: ciao-down should cache base images and other downloads
Currently, when running ciao-down prepare, all the images needed by the guest VM are downloaded inside the VM, e.g., go, Fedora, clear, etc.  It would be nice to download this stuff on the host and cache it in the ~/.ciao-down folder, along with the base Ubuntu image.  This would speed up subsequent guest creations.

markdryan:
This commit adds a new function that can be used when creating
ciao down cloud init files.  The function called download
instructs ciao-down to download and cache the downloaded files on
the host.  The next time a VM requests the same file ciao-down will
serve the file from its cache rather than downloading it
from the Internet.  This speeds up the creation of ciao VMs by 5.5
minutes and clear container VMs by about 20 seconds.

Fixes #957

### DIFF
--- a/testutil/ciao-down/cloudinit.go
+++ b/testutil/ciao-down/cloudinit.go
@@ -94,7 +94,7 @@ runcmd:
  - echo "PATH=\"$PATH:/usr/local/go/bin:{{$.GoPath}}/bin:/usr/local/nodejs/bin\""  >> /etc/environment
 
  - curl -X PUT -d "Downloading Go" 10.0.2.2:{{.HTTPServerPort}}
- - {{template "ENV" .}}wget https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz -O /tmp/go1.8.linux-amd64.tar.gz
+ - {{download . "https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz" "/tmp/go1.8.linux-amd64.tar.gz"}}
  - {{template "CHECK" .}}
  - curl -X PUT -d "Unpacking Go" 10.0.2.2:{{.HTTPServerPort}}
  - tar -C /usr/local -xzf /tmp/go1.8.linux-amd64.tar.gz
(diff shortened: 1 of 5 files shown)


testutil/ciao-down/cloudinit.go
```diff
@@ -94,7 +94,7 @@ runcmd:
  - echo "PATH=\"$PATH:/usr/local/go/bin:{{$.GoPath}}/bin:/usr/local/nodejs/bin\""  >> /etc/environment
 
  - curl -X PUT -d "Downloading Go" 10.0.2.2:{{.HTTPServerPort}}
- - {{template "ENV" .}}wget https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz -O /tmp/go1.8.linux-amd64.tar.gz
```
tpepper: Not really related to what you're changing in this patch, but:

If this were split into an url and file in a const struct like with guestDownloadURL, there wouldn't be duplication of the strings for the filename (eg: "go1.8.linux-amd64.tar.gz").

It would be better to not do this in something world writeable like /tmp, versus ~/local.

Is there a reason to use both curl and wget here?  I suggest removing wget and simply using curl consistently, instead of propagating the wget usage over into testutil/ciao-down/prepare.go downloadFN().

markdryan: > If this were split into an url and file in a const struct like with guestDownloadURL, there wouldn't be duplication of the strings for the filename (eg: "go1.8.linux-amd64.tar.gz").

I could add a downloadTo function which could be used when the last part of the url matches the desired filename.  That way you'd only need to specify a directory, e.g.,

```
{{downloadTo . "https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz" "/tmp"}}
```


> It would be better to not do this in something world writeable like /tmp, versus ~/local.

Okay

> Is there a reason to use both curl and wget here? 

To be honest I don't remember but I agree it would make sense to standardise on one.  Going forward these programs should mostly be hidden from the script writer.  Curl is currently used to send status information back to ciao-down.  This task will soon be replaced by a function.

```
- curl -X PUT -d "Downloading Go" 10.0.2.2:{{.HTTPServerPort}}
```

will become

```
- {{info . "Downloading Go"}}
```

markdryan:
@mcastelino Is this okay for you?

testutil/ciao-down/download.go
```diff
@@ -88,25 +106,50 @@ func getUbuntu(ctx context.Context, dest io.WriteCloser, cb progressCB) (err err
...
-	imgPath := path.Join(instanceDir, guestImageName)
+func downloadProgress(p progress) {
+	if p.totalMB >= 0 {
+		fmt.Printf("Downloaded %d MB of %d\n", p.downloadedMB, p.totalMB)
+	} else {
+		fmt.Printf("Downloaded %d MB\n", p.downloadedMB)
+	}
+}
+
+func downloadFile(ctx context.Context, URL, ciaoDir string, cb progressCB) (string, error) {
```
mcastelino: @markdryan will there be issues here if we support multiple ciao-down instances in the future?

markdryan: Yes, assuming we allow concurrent prepares.  But there will be lots of other issues too unrelated to downloaded that will also need to be addressed, for example downloading of the ubuntu base image and modifying the UI to cope with parallel prepares.

Issue #1191, opened 2017-02-22
api: block storage: bootable volume flag inconsistent


jorgevgut:
Hi,

I'm testing out the bootable volume feature through the webui and I'm sending this data in a POST request to create a bootable volume.
```
{volume:{name:testboot,
size:1,
description:"test boot",
bootable:"true",
imageRef:"df3768da-31f5-4ba6-82f0-127a1a705169"}}
```

controller's POST response is the following (endpoint is /v2/{tenant}/volumes POST method):
```
{"volume":{
"status":"available",
"migration_status":null,
"user_id":"6ce8104700414973b7954defb582739f",
"attachments":[],
"links":[],
"availability_zone":null,
"bootable":"true",
"encrypted":false,
"created_at":"2017-02-22T00:04:49.294973562Z",
"description":null,
"updated_at":null,
"volume_type":null,
"name":null,
"replication_status":"",
"consistencygroup_id":null,
"source_volid":null,
"snapshot_id":null,
"multiattach":false,
"metadata":null,
"id":"ec2b330b-ce74-4a8d-908a-c76ae5d44ed1",
"size":1}}
```

This response clearly states that bootable value is set as "true".
However /v2/{tenant}/volumes/detail endpoint (Method GET)
does not show bootable as true, but empty which does not match the response from the POST endpoint.
```
{
migration_status: null,
attachments: null,
links: null,
availability_zone: null,
os-vol-host-attr:host: "",
encrypted: false,
updated_at: null,
replication_status: "",
snapshot_id: null,
id: "ec2b330b-ce74-4a8d-908a-c76ae5d44ed1",
size: 1,
user_id: "",
os-vol-tenant-attr:tenant_id: "6ce8104700414973b7954defb582739f",
os-vol-mig-status-attr:migstat: null,
metadata: null,
status: "available",
description: "test boot",
multiattach: false,
source_volid: null,
consistencygroup_id: null,
os-vol-mig-status-attr:name_id: null,
name: "testboot",
bootable: "",
created_at: "2017-02-22T00:04:49.294973562Z",
volume_type: null
}
```

Impact: I currently can't validate if request to create the bootable volume is working properly due to inconsistency in the responses from the api endpoints.

rbradford:
@jorgevgut feel free to test this change if you're interested.

jorgevgut:
Thank you @rbradford  I'll test it today and let know the outcome.

jorgevgut:
[...]

PR #1195, opened 2017-02-22
Fix bootable volume handling in Open Stack API

rbradford:
* Record that volumes created via the API from an ImageRef are marked bootable as per spec
* Report out volume bootable status via API
* Unit testing of the above

Fixes: #1191


rbradford:
@kaccardi to review as this touches controller code

PR #1197, opened 2017-02-22
Add instance tracking to quotas

rbradford:
This PR adds quota tracking for instances. When instances are created the resources specified in the workload are consumed. When the instance is deleted (or fails to start) then the resources are released.

On controller startup any existing instances also have their resources consumed.

### DIFF
--- a/_release/bat/quotas_bat/quotas_bat_test.go
+++ b/_release/bat/quotas_bat/quotas_bat_test.go
@@ -265,3 +265,75 @@ func TestInstanceLimited(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+// Try launching instance that is over quota and check usage
+//
+// Sets a quota that would deny the instance from being created and
+// then check the usage reflects the instance failed to start.
+//
+// Set an memory quota limit < workload usage. Try and create instance
+// of workload and check that instance launch fails and that the usage
+// reflects that.
+func TestInstanceUsageAfterDenial(t *testing.T) {
+	ctx, cancelFunc := context.WithTimeout(context.Background(), standardTimeout)
+	defer cancelFunc()
+	tenants, err := bat.GetUserTenants(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(tenants) < 1 {
+		t.Fatal("Expected user to have access to at least one tenant")
+	}
+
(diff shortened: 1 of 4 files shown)


rbradford:
The structure of this change is the same as was needed for volumes. It's had some manual testing but I will try to create a BAT test for this.

rbradford:
Forgot to add [WIP] - not looking for merge until BAT tests are in this PR.

I would like to see review of the controller changes as they are ready.

rbradford:
I've removed the WIP on this - ready for review.

kaccardi:
forgot to add that the changes seem ok to me.

rbradford:
@kaccardi thanks for the lgtm. Also added a new BAT test to check usage after instance launch is denief by being over quota.

_release/bat/quotas_bat/quotas_bat_test.go
```diff
@@ -96,3 +97,235 @@ func TestQuotas(t *testing.T) {
...
+	}
+
+	instances, err := bat.LaunchInstances(ctx, "", wl.ID, 3)
+	if err != nil {
+		t.Error(err)
+	}
+
+	scheduled, err := bat.WaitForInstancesLaunch(ctx, "", instances, false)
+	if err != nil {
+		t.Fatalf("Instance %s did not launch: %v", instances[0], err)
```
markdryan: We don't know it was instances[0] that failed to launch.  To know instances actually failed you'd need to compare scheduled and instances.

markdryan: Also, as this is a Fatal, it means the instances that did launch will not get cleaned up.

_release/bat/quotas_bat/quotas_bat_test.go
```diff
@@ -96,3 +97,235 @@ func TestQuotas(t *testing.T) {
...
+	if !checkUsage(updatedQuotas, wl, 0) {
+		t.Error("Usage not recorded correctly")
+	}
+
+	err = restoreQuotas(ctx, tenantID, origQuotas, updatedQuotas)
+	if err != nil {
+		t.Error(err)
+	}
+
+	for _, i := range scheduled {
```
markdryan: Is there any reason not to just call DeleteInstances here?  Might speed things up a bit.

PR #1198, opened 2017-02-22
ansible: Run ciao-controller as ciao user

albertomurillo:
ciao-controller was run as root to allow it to run ceph commands.

By changing the ownership of /etc/ceph and its content to be owned by
ciao, ciao-controller can be run without root privileges.

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/cleanup/ciao.yml
+++ b/_DeploymentAndDistroPackaging/ansible/cleanup/ciao.yml
@@ -141,6 +141,10 @@
         - /etc/pki/ciao
         - /etc/pki/keystone
         - /var/lib/ciao
+        - /tmp/lock/ciao
+        - /tmp/ciao-controller-stats.db
+        - /tmp/ciao-controller-stats.db-shm
+        - /tmp/ciao-controller-stats.db-wal
         # ClearLinux cers
         - /etc/ssl/certs/keystone_cert.pem
         # Ubuntu certs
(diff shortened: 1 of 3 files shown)


markdryan:
So the problem is the database files ciao-controller stores in /tmp

```
ls -l /tmp
total 86976
-rw-r--r-- 1 root   root   74387456 Feb 24 13:24 ciao-controller-stats.db
-rw-r--r-- 1 root   root      32768 Feb 24 13:46 ciao-controller-stats.db-shm
-rw-r--r-- 1 root   root    4144752 Feb 24 13:32 ciao-controller-stats.db-wal
```

If we remove these ciao-controller can start and run as ciao.  The cleanup script needs to be updated.

Thanks for @rbradford for the diagnosis.

albertomurillo:
@markdryan  I fixed the cleanup scrip

markdryan:
The issue with the database files in /tmp is now fixed and controller does now get started.  Unfortunately, the cluster is not functional as the CNCI cannot be started.

```
. bat-ciaorc 
ciao-cli volume list
F0228 17:06:27.826308    9305 main.go:469] ciao-cli FATAL: Expected HTTP response code [200 204] when accessing [GET https://ciao-ctl.ld.intel.com:8776/v2/d6c53b95660c4a0ab8a2f596468dbf10/volumes/detail], but got 500 instead
```

This works fine without this patch applied.

This

markdryan:
It seems that the issue is that I'm using skip_ceph=true and so the permissions of my ceph directory were not correct.  Could you update the [development doc](https://github.com/01org/ciao/blob/master/_DeploymentAndDistroPackaging/ansible/doc/development.md) to say that if you deploy ceph yourself the /etc/ceph directory must be readable by the ciao user.

albertomurillo:
@markdryan https://github.com/01org/ciao/pull/1208 Provides a way to deploy the ceph/demo container or use the existing ceph configuration on the ciao nodes.

The PR above changes the permission and modes of `/etc/ceph/` and its files even if `ceph_config=none` is set.

Issue #1158, opened 2017-02-16
Ciao lets any user clear all cluster events


rbradford:
The clear event API does not take a tenant and there is no admin only check:

```
rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli instance add -workload 46e2ac70-e913-45d7-8a36-da84425b1cd6
Created new (pending) instance: b08701bf-e04c-4ed4-840b-6deeaefce6dd

rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli instance delete -instance b08701bf-e04c-4ed4-840b-6deeaefce6dd
Deleted instance: b08701bf-e04c-4ed4-840b-6deeaefce6dd

rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli event list1 Ciao event(s):
	[1] 2017-02-16 17:02:19 +0000 UTC: info:Deleted Instance b08701bf-e04c-4ed4-840b-6deeaefce6dd (Tenant c14b1f782180411e8af3cd031c94f279)

rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ ciao-cli event delete
Deleted all event logs

rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli event list
0 Ciao event(s):
```

kaccardi:
yes - event delete when called from non-privileged context should only delete the tenant's own logs.

kaccardi:
@rbradford which user are you using when you do ciao-cli event delete? If you are using a privileged user, this call should succeed. For example, user "csr" is actually a privileged user.

rbradford:
Possible for demo user too:

```
rob@singlevm:~/src/go/src/github.com/01org/ciao/testutil/singlevm$ CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli event list
1 Ciao event(s):
[...]

PR #1199, opened 2017-02-22
Add ability for http services to check for user privilege

kaccardi:
Let the identity middleware communicate privilege to the service handlers via context. Modify the ciao API and the legacy ciao API to specify whether routes should be for privileged users only, and send http status Unauthorized if the user doesn't have sufficient privilege.

Fixes: #1158

### DIFF
--- a/ciao-controller/api/api.go
+++ b/ciao-controller/api/api.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 
 	"github.com/01org/ciao/ciao-controller/types"
+	"github.com/01org/ciao/service"
 	"github.com/01org/ciao/ssntp/uuid"
 	"github.com/golang/glog"
 	"github.com/gorilla/mux"
@@ -97,10 +98,20 @@ func errorResponse(err error) Response {
 // and pass some package level context into the handler.
 type Handler struct {
 	*Context
-	Handler func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Handler    func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Privileged bool
 }
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// check whether we should send permission denied for this route.
+	if h.Privileged {
+		privileged := service.GetPrivilege(r.Context())
+		if !privileged {
+			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+			return
+		}
+	}
+
 	// set the content type to whatever was requested.
 	contentType := r.Header.Get("Content-Type")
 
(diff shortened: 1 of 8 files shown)


openstack/identity/identity.go
```diff
@@ -209,6 +212,7 @@ func (h Handler) checkToken(r *http.Request, tenant string, tokenGetResult getRe
 	/* TODO Caching or PKI */
 	for _, a := range h.ValidAdmins {
 		if validateProjectRole(tokenGetResult, a.Project, a.Role) == true {
+			context.Set(r, service.PrivKey, "true")
 			return true
```
obedmr: Is here where we can add the tenantID information to the context? it's going to be useful for ``ciao-image`` service which requires tenantID for doing the images tenant/project separation.

ciao-controller/api/api.go
```diff
@@ -96,9 +98,18 @@ func errorResponse(err error) Response {
...
 	Handler func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Priv    bool
 }
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// check whether we should send permission denied for this route.
+	if h.Priv {
+		_, ok := context.GetOk(r, service.PrivKey)
+		if !ok {
+			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
```
rbradford: I'm a bit confused. Does the code below still not get called? And thus although you've sent a 403 and side effects will still happen by the call into h.Handler(h.Context, w, r) ?

ciao-controller/api/api.go
```diff
@@ -96,9 +98,18 @@ func errorResponse(err error) Response {
 type Handler struct {
 	*Context
 	Handler func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Priv    bool
```
rbradford: Very minor but i'd rather this was named something like Privileged/AdminOnly or something like that. Priv is a teeny bit vague.

ciao-controller/legacy_api.go
```diff
@@ -270,43 +281,43 @@ func listServerDetailsFlavors(c *controller, w http.ResponseWriter, r *http.Requ
...
-		legacyAPIHandler{ctl, listTenantResources}).Methods("GET")
+		legacyAPIHandler{ctl, listTenantResources, false}).Methods("GET")
 
 	r.Handle("/v2.1/{tenant}/quotas",
-		legacyAPIHandler{ctl, listTenantQuotas}).Methods("GET")
+		legacyAPIHandler{ctl, listTenantQuotas, false}).Methods("GET")
 
 	r.Handle("/v2.1/tenants",
-		legacyAPIHandler{ctl, legacyListTenants}).Methods("GET")
+		legacyAPIHandler{ctl, legacyListTenants, true}).Methods("GET")
```
rbradford: I wonder if this one shouldn't be admin only? I know BAT tests rely on user ciao-cli tenant list working.

openstack/identity/identity.go
```diff
@@ -209,6 +212,7 @@ func (h Handler) checkToken(r *http.Request, tenant string, tokenGetResult getRe
 	/* TODO Caching or PKI */
 	for _, a := range h.ValidAdmins {
 		if validateProjectRole(tokenGetResult, a.Project, a.Role) == true {
+			context.Set(r, service.PrivKey, "true")
```
markdryan: I don't think we need to use gorilla context.  The standard library now has everything we need.  If I understand correctly, what we're trying to do is to associate some information with a request object and retrieve this info at a later state.

This could be done as follows

r = r.WithContext(context.WithValue(r.Context(), service.PrivKey, "true")) 

Note you'd need to return r from this function and from validateToken.

Then in ServeHTTP, you'd do a 

ok := r.Context().Value(service.PrivKey)

ciao-controller/legacy_api.go
rbradford: Please disregard. I ran BAT tests and ciao-cli tenant list as the normal user and everything worked as expected so this entry point is not the one I thought it was.

ciao-controller/api/api.go
kaccardi: looks like I forgot the return - thanks for catching that.

ciao-controller/api/api.go
kaccardi: ok.

openstack/identity/identity.go
kaccardi: we'd do it in a different place, but similarly. We'd need to use a different Key value. This PR shows you how I'd do it - I'll submit a follow on PR to actually store the tenant ID.

kaccardi:
@markdryan updated to use the built in context pkg.

ciao-controller/api/api.go
```diff
@@ -95,10 +96,20 @@ func errorResponse(err error) Response {
...
 	*Context
-	Handler func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Handler    func(*Context, http.ResponseWriter, *http.Request) (Response, error)
+	Privileged bool
 }
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// check whether we should send permission denied for this route.
+	if h.Privileged {
+		_, ok := service.GetPrivilege(r.Context())
```
markdryan: Are we checking the correct bool here?  Ok is true if the value associated with PrivKey is of type bool.  We don't actually check to see whether that bool is true or not.

It might be simpler to remove the check in the type assertion in  GetPrividege so there is only one bool

```
func GetPrivilege(ctx context.Context) (bool, bool) {
	return  ctx.Value(PrivKey).(bool)
}
```

Not this will obviously panic if the value is not a boot, but if this panic occurs there's something badly wrong with controller and a panic is probably the right thing to do, like dereferencing null.

service/service.go
```diff
@@ -0,0 +1,36 @@
...
+const PrivKey key = 0
+
+// GetPrivilege returns the value of PrivKey
+func GetPrivilege(ctx context.Context) (bool, bool) {
+	privilege, ok := ctx.Value(PrivKey).(bool)
+	return privilege, ok
+}
+
+// SetPrivilege is used to set the value of PrivKey
+func SetPrivilege(ctx context.Context, privileged bool) context.Context {
```
markdryan: We seem to ignore the privileged parameter here and just set the value to true.  Maybe we could get rid of it.

ciao-controller/api/api.go
kaccardi: right! thanks, what was I thinking?

kaccardi:
@markdryan updated to address your most recent feedback.

mrkz:
lgtm.

I've tested this with a privileged/non-privileged tenants for a privileged route, returning 401 for non-admin request and handling correctly the admin request, so I'm OK with the change.

ciao-controller/legacy_api.go
```diff
@@ -270,43 +281,43 @@ func listServerDetailsFlavors(c *controller, w http.ResponseWriter, r *http.Requ
 
 func legacyComputeRoutes(ctl *controller, r *mux.Router) *mux.Router {
 	r.Handle("/v2.1/{tenant}/servers/action",
-		legacyAPIHandler{ctl, tenantServersAction}).Methods("POST")
+		legacyAPIHandler{ctl, tenantServersAction, false}).Methods("POST")
 
 	r.Handle("/v2.1/flavors/{flavor}/servers/detail",
-		legacyAPIHandler{ctl, listServerDetailsFlavors}).Methods("GET")
+		legacyAPIHandler{ctl, listServerDetailsFlavors, true}).Methods("GET")
```
markdryan: verify.sh is failing for me.  I see

```
Waiting for instance to boot
F0228 16:25:17.518129   13880 instance.go:767] ciao-cli FATAL: HTTP Error [401] for [GET https://singlevm:8774/v2.1/flavors/38acdccd-506b-4cad-acc4-ee092d898a2f/servers/detail]: Unauthorized
```

Should this one be false perhaps or does verify.sh need updating?

ciao-controller/legacy_api.go
```diff
@@ -270,43 +281,43 @@ func listServerDetailsFlavors(c *controller, w http.ResponseWriter, r *http.Requ
...
 	r.Handle("/v2.1/cncis",
-		legacyAPIHandler{ctl, legacyListCNCIs}).Methods("GET")
+		legacyAPIHandler{ctl, legacyListCNCIs, true}).Methods("GET")
 	r.Handle("/v2.1/cncis/{cnci}/detail",
-		legacyAPIHandler{ctl, legacyListCNCIDetails}).Methods("GET")
+		legacyAPIHandler{ctl, legacyListCNCIDetails, true}).Methods("GET")
 
 	r.Handle("/v2.1/events",
-		legacyAPIHandler{ctl, legacyListEvents}).Methods("GET")
+		legacyAPIHandler{ctl, legacyListEvents, true}).Methods("GET")
```
markdryan: verify.sh is also complaining about this one.

```
F0228 16:34:53.131197   21626 event.go:131] ciao-cli FATAL: HTTP Error [401] for [DELETE https://singlevm:8774/v2.1/events]: Unauthorized
```

kaccardi: This one is a problem with verify.sh - unless we are admin, we shouldn't use this api, we should use the per tenant api.

ciao-controller/legacy_api.go
kaccardi: This is a problem with verify.sh - this route should be admin only.

kaccardi: after some thought - I'm going to modify ciao-cli to use the per tenant listServerDetails route and modify it to accept a query parameter of "flavor". That way a normal user can still use ciao-cli instance list -workload <someuuid>

obedmr:
LGTM, it works at my end and will be useful for ciao-image service.

markdryan:
@kaccardi We should probably add a README.md to the service directory.  I've entered a separate issue for this #1212, so we can go ahead and merge this PR now.

Issue #1135, opened 2017-02-09
BAT tests for workload delete


kaccardi:
create BAT tests for workload delete for both privileged and non-privileged users.

PR #1201, opened 2017-02-23
Add support for deleting workloads

kaccardi:
Create the API and ciao-cli support for deleting unused public or private workloads. Modify BAT tests to delete the test workload.

Fixes: #949, #950, #951, #952, #1135

### DIFF
--- a/_release/bat/workload_bat/workload_bat_test.go
+++ b/_release/bat/workload_bat/workload_bat_test.go
@@ -40,9 +40,6 @@ users:
 const vmWorkloadImageName = "Fedora Cloud Base 24-1.2"
 
 func testCreateWorkload(t *testing.T, public bool) {
-	// until we support delete workload, we will explicitly skip this test.
-	t.Skip()
-
 	// we'll use empty string for now
 	tenant := ""
 
@@ -114,6 +111,23 @@ func testCreateWorkload(t *testing.T, public bool) {
 
 	if w.Name != opt.Description || w.CPUs != opt.Defaults.VCPUs || w.Mem != opt.Defaults.MemMB {
 		t.Fatalf("Workload not defined correctly")
+	}
+
+	// delete the workload.
+	if public {
+		err = bat.DeletePublicWorkload(ctx, w.ID)
+	} else {
+		err = bat.DeleteWorkload(ctx, tenant, w.ID)
+	}
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// now try to retrieve the workload from controller.
+	_, err = bat.GetWorkloadByID(ctx, "", ID)
+	if err == nil {
+		t.Fatalf("Workload not deleted correctly")
 	}
 }
 
--- a/ciao-controller/api/api.go
+++ b/ciao-controller/api/api.go
@@ -70,7 +70,8 @@ func errorResponse(err error) Response {
 	case types.ErrPoolNotFound,
 		types.ErrTenantNotFound,
 		types.ErrAddressNotFound,
-		types.ErrInstanceNotFound:
+		types.ErrInstanceNotFound,
+		types.ErrWorkloadNotFound:
 		return Response{http.StatusNotFound, nil}
 
 	case types.ErrQuota,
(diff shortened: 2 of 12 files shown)


ciao-controller/internal/datastore/sqlite3db.go
```diff
@@ -1139,6 +1155,46 @@ func (ds *sqliteDB) updateWorkload(w types.Workload) error {
...
+	}
+
+	err = ds.deleteWorkloadStorage(tx, ID)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	filename := fmt.Sprintf("%s_config.yaml", ID)
+	path := fmt.Sprintf("%s/%s", ds.workloadsPath, filename)
```
markdryan: Would it not be better to delete the file right at the end.  If the "DELETE FROM workload_template WHERE id = ?" fails, the transaction will be rolled back but the _config.yaml file will be gone.

markdryan:
I'm getting 500 Internal server error for the two error cases, workload does not exist and workload is in use, e.g.,

```
{"error":{"code":500,"name":"Internal Server Error","message":"Workload definition still in use"}}
```

Is this expected?

markdryan:
@kaccardi Just a couple of minor comments and questions.  I gave it a good test and everything is working for me on SingleVM.  Nice to see all the BAT tests running as well.

ciao-controller/api/api.go
```diff
@@ -650,10 +669,18 @@ func Routes(config Config) *mux.Router {
 	route.Methods("POST")
 	route.HeadersRegexp("Content-Type", matchContent)
 
-	route = r.Handle("/{tenant:[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[8|9|aA|bB][a-f0-9]{3}-?[a-f0-9]{12}}/workloads", Handler{context, addWorkload})
```
rbradford: Pleased to see you fixed this - I spotted this on my other review.

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -115,6 +112,23 @@ func testCreateWorkload(t *testing.T, public bool) {
 	if w.Name != opt.Description || w.CPUs != opt.Defaults.VCPUs || w.Mem != opt.Defaults.MemMB {
 		t.Fatalf("Workload not defined correctly")
 	}
+
+	// delete the workload. We only support delete of public workloads.
```
rbradford: This comment is confusing. You delete a non-public workload in the second part of the if.

ciao-controller/types/types.go
```diff
@@ -573,6 +573,9 @@ var (
 
 	// ErrWorkloadNotFound is returned when a workload ID cannot be found
 	ErrWorkloadNotFound = errors.New("Workload not found")
+
+	// ErrWorkloadInUse is returned by DeleteWorkload when an instance of a workload has been created.
```
rbradford: s/has been created/is still active/

ciao-controller/internal/datastore/sqlite3db.go
rbradford: As it currently stands this protects the workload in the database if the remove fails. I think we should follow @markdryan's suggestion. As if you're failing to delete it's probably because it's missing or some other problem so it will be better to delete it from the database than have an unlauchable workload as the required yaml file is missing.

kaccardi:
@markdryan I forgot to fix the error response for those 2 error conditions. They should have returned NotFound or Forbidden.

_release/bat/workload_bat/workload_bat_test.go
kaccardi: sorry, forgot to update the comment after I added support for private workload delete.

ciao-controller/api/api.go
kaccardi: one or the other is going to get a merge conflict depending on who goes in first....

kaccardi:
@markdryan @rbradford updated to address your feedback.

PR #1205, opened 2017-02-27
Improve partial instance launching

rbradford:
If launching some instances out of the request fail then we can end up in situating with unexpected "dangling" instances where an error is returned but also some instances started.

With this change if some instances did get launched correctly then report then out and asynchronously send an event. If only one instance was started if or no instances got launched correctly then return the error directly

With a quota volume limit:

```
rob@singlevm:~/go/src/github.com/01org/ciao/testutil/singlevm$ ciao-cli instance add -workload ed39a2c3-19f2-48ee-a650-483426d44760 --instances 8
Some instances failed to start - check the event log for details.
Created new (pending) instance: 08b78ca4-4fd1-445e-bce6-dd91ff5d9408
Created new (pending) instance: 804abebd-c112-45c7-9c74-3d27ed137922
Created new (pending) instance: 9ee79eb6-927b-46af-9620-5a6febc6c4df
Created new (pending) instance: 31e91e94-acd0-4913-99c1-012c2b2d8fda
rob@singlevm:~/go/src/github.com/01org/ciao/testutil/singlevm$ ciao-cli event list
1 Ciao event(s):
        [1] 2017-02-27 19:01:36 +0000 UTC: error:Error launching instance(s): Error creating instance: Error creating volume: Over quota (Tenant 56342297a18946a6b1863c6697449cf9)
```

### DIFF
--- a/ciao-cli/instance.go
+++ b/ciao-cli/instance.go
@@ -519,9 +519,14 @@ func (cmd *instanceAddCommand) run(args []string) error {
 		fatalf(err.Error())
 	}
 
+	if len(servers.Servers) < cmd.instances {
+		fmt.Println("Some instances failed to start - check the event log for details.")
+	}
+
 	for _, server := range servers.Servers {
 		fmt.Printf("Created new (pending) instance: %s\n", server.ID)
 	}
+
 	return nil
 }
 
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -17,13 +17,12 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
 	"github.com/01org/ciao/ciao-controller/types"
 	"github.com/01org/ciao/payloads"
-	"github.com/golang/glog"
+	"github.com/pkg/errors"
 )
 
 func (c *controller) evacuateNode(nodeID string) error {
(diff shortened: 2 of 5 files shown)


rbradford:
Hi @kaccardi. Please could you review as ciao-controller owner.

ciao-cli/instance.go
```diff
@@ -519,9 +519,14 @@ func (cmd *instanceAddCommand) run(args []string) error {
 		fatalf(err.Error())
 	}
 
+	if len(servers.Servers) < cmd.instances {
```
markdryan: So if at least one instance starts, ciao-cli does not return an error?  Is this correct?  

If so, I think we should also change,  [bat.parseInstances](https://github.com/01org/ciao/blob/master/bat/compute.go#L519).  Currently, this function would return an error as it would not be able to parse "Some instances failed to start - check the event log for details."  What we really want it to do is to return information about the instances that were successfully created.  This will allow the calling BAT test to clean them up properly.  The other option would be to move the

"Some instances failed to start - check the event log for details."

after the server ids have been output but this is arguable more fragile.

rbradford: And there is also the expectation that there would be as many lines as requested instances whereas there might be fewer. I'll look at addressing that.

markdryan: On reflection, this is not a new problem.  LaunchInstances returns an error without this patch if any of the instances fail to launch.  I'll therefore enter a separate bug for this.

PR #1208, opened 2017-02-28
ansible: Install ceph container

### LINKED ISSUE #1084: ansible: Add an option to use a ceph/demo container as Storage BackEnd

albertomurillo:
Introduce ceph_config variable which behaves acordingly to the
following values.

- none: Ceph is already configured and ansible will not setup ceph
- container: Ansible will deploy ceph/demo container on the controller
             node and configure the rest of the nodes to use it.
- files: An existing ceph cluster exists and ansible will configure all
         nodes using the existing ceph configuratino and authentication
         files found in ceph_config_dir variable.

Fixes #1084

### DIFF
--- a/_DeploymentAndDistroPackaging/ansible/ciao_controllers.yml
+++ b/_DeploymentAndDistroPackaging/ansible/ciao_controllers.yml
@@ -16,7 +16,9 @@
   - hosts: controllers
     become: yes
     roles:
+      - ciao-common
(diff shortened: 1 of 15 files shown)


rbradford:
Note: I only tested the ceph_config=none path as that reflects our cluster design. I'm trusting that you tested the other paths as part of you development process.

Issue #1209, opened 2017-03-01
bat.LaunchInstances fails if any of the instances fail to start


markdryan:
PR #1205 modified ciao-cli instance add to not fail it it succeeded in launching at least one instance.   However, bat.LaunchInstances does not take advantage of this new behaviour.  It will still fail if any one of the request instances fails to launch.  This isn't great as it makes it difficult for the BAT tests to cleanup after themselves in the case of failure.  The behaviour of this function should be changed so that it mirrors the new behaviour of ciao-cli, i.e., that it only fails if no instances were started.  When some instances successfully launched, but others did not, a list of the successfully launched instances should be returned so they can be deleted.

The easiest way to fix this may be to add a -f option to the ciao-cli instance add command.

PR #1210, opened 2017-03-01
Add templating to ciao-cli instance add

rbradford:
Add templating to ciao-cli instance add and then use that in the BAT tests for obtaining the list of launched instances.

Fixes: #1209


rbradford:
@markdryan PTAL.

PR #1211, opened 2017-03-01
ciao-image, bat: Fix zero listed images in BAT tests

obedmr:
This commit fixes the failure of ciao-image BAT tests that runs on a Ciao
cluster that has 0 images.

It will fix the following parsing error:
```
--- FAIL: TestImageList (0.16s)
        image_bat_test.go:110: Unable to count number of images: strconv.ParseInt: parsing "": invalid syntax
```

### DIFF
--- a/bat/image.go
+++ b/bat/image.go
@@ -194,7 +194,12 @@ func GetImageCount(ctx context.Context, tenant string) (int, error) {
 		return 0, err
 	}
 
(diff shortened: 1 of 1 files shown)


markdryan:
And here it is

https://github.com/01org/ciao/pull/1217

rbradford:
@obedmr thanks for finding this bug but the solution in #1217 is the correct one.

PR #1213, opened 2017-03-02
database: Add tables listing functionality

obedmr:
New features is added in order to list all tables that exists
in a database. For now, it will be useful for ciao-image service which
is storing each tenant's images in a single table (bucket).

It will help on automation when a full database scan or search
is required. Once you have the list of tables (buckets), you can
go and search for the specific element's ID.


obedmr:
@rbradford ok, will add it to my ciao-image branch PR.

PR #1214, opened 2017-03-02
identity: Add tenantID get and set for service calls

obedmr:
This commit adds tenant ID setting and getting from the identity
middleware, it will add tenant ID to the service call context in
order to be used on each service separately.

### DIFF
--- a/openstack/identity/identity.go
+++ b/openstack/identity/identity.go
@@ -253,6 +253,7 @@ func (h Handler) validateToken(ctx context.Context, r *http.Request) (context.Co
 
 	if tenantFromVars == "" {
 		glog.V(2).Infof("Token validation for [%s]", tenantFromToken)
+		ctx = service.SetTenantID(ctx, tenantFromToken)
 		return h.checkToken(ctx, r, tenantFromToken, tokenResult)
 	}
 	// verify that tenant from token is consistent with the tenant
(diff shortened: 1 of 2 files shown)


obedmr:
@kaccardi could you take a look on it?

obedmr:
@rbradford @mcastelino could you take a look on this?

rbradford:
The code looks fine but as far as I cant tell nothing uses the newly added functions. It would be better to include this change on the PR that uses it.

obedmr:
@rbradford , ok, will add it to the #1123

mrkz:
@rbradford Actually, I'd like to reconsider pulling this PR instead of adding it as part of #1123.

I ask this because I'd be using this bits as well, my use case is as follows:

At some point, for the cluster configuration update, the ciao-controller will need to perform some API test call when any endpoint setting is changed (e.g: the `controller.compute_port`).
For that test to run, I'd need to perform an http request to the new endpoint (on this case, the URL for the request contains the tenantID as part of it: `https://controller.example.com:8774/v2.1/{tenant_id}/servers/detail`).

As the API call for the configuration change did not include the tenant_id (due to being a privileged call/route), I'd have no other way to know the tenantID but by performing some operations with the received token (which is the reason of all this functionality was created).

By having this merged, I could use it and then I could form the following request:
```
[...]
if computeFQDN == "" {
        computeEndpointURL = fmt.Sprintf("https://localhost:%s/v2.1/%s/servers/detail", newConf.Value, tenantID)
} else {
             computeEndpointURL = fmt.Sprintf("https://%s:%s/v2.1/%s/servers/detail", computeFQDN, newConf.Value, tenantID)
}
fmt.Printf("Will test new compute endpoint: %s\n", computeEndpointURL)
[...]
```

obedmr:
Closing because #1123 contains this change.

PR #1217, opened 2017-03-02
ciao-cli: Fix template handling of ciao-cli image list

markdryan:
ciao-cli image list was not handling templates correctly.  It was
calling the template function once for each page rather than once
for the entire list.  This breaks scripts such as {{len .}} which
attempt to compute the total number of images.


obedmr:
@markdryan cool, thanks.

PR #1219, opened 2017-03-02
Add SSH functions into bat package, fix capital letter and add ctx to…

arturo-garcia-garcia:
… functions that will  be exported

Update base_bat_test.go to use the functions form the bat package instead of use code directly

Update name of function SendCommand to SSHSendCommand in the ssh.go file to be more clear of what it does

Add the delete of the workload

Fix call of functions in ssh.go to return error if something fails

Trim the spaces in the retunr of the SSH call to validate is the right user


bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+	ClientPass string
+}
+
+//CreateSSHClient creates a SSH session for the server specified
+//
+//Receive a struct with the details of the server that will be connected
+//crete a tcp session that will be use to send and receive data
+//
+//sesion sould be created succesfully and wait for commands
+func CreateSSHClient(WLclient SSHClient) (*ssh.Client, error) {
```
markdryan: WLclient isn't really in camel case.  It should be WLClient, or maybe just client or even c would be fine.

markdryan: We need to pass a [context](https://golang.org/pkg/context/) in here.  Instead of calling Dial we would then call (DialContext)[https://golang.org/pkg/net/#Dialer.DialContext].  This allows us to cancel and to timeout the SSH request.

bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+//Recive a struct with the details of the server that will be connected
+//and the command that will be send to the server, set up the connection,
+//send the command and returns server's response
+//
+//function returns output from the server to the specified command
+func SSHSendCommand(ctx context.Context, Client SSHClient, command string) (string, error) {
+	//create a connection
+	Conection, err := CreateSSHClient(Client)
+	if err != nil {
+		panic("Failed to dial: " + err.Error())
```
markdryan: return an error here.  Don't panic.

bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+//
+
+package bat
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"golang.org/x/crypto/ssh"
```
markdryan: We'll need to vendor this package. I can do this (and push the commit directly to your branch) when we're ready to merge.  So don't worry about this.

bat/ssh.go
markdryan: I see that ssh.Dial doesn't have a ssh.DialContext version.  So we'll need to write our own.  It should be simple.

Just copy

https://github.com/golang/crypto/blob/master/ssh/client.go#L164

replacing net.DialTimeout with net.DialContext

bat/ssh.go
markdryan: Also we need to close Connection somewhere.

bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+//SSHSendCommand  send a command to a specified server
+//
+//Recive a struct with the details of the server that will be connected
+//and the command that will be send to the server, set up the connection,
+//send the command and returns server's response
+//
+//function returns output from the server to the specified command
+func SSHSendCommand(ctx context.Context, Client SSHClient, command string) (string, error) {
+	//create a connection
+	Conection, err := CreateSSHClient(Client)
```
markdryan: Should be connection.

bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+//SSHSendCommand  send a command to a specified server
+//
+//Recive a struct with the details of the server that will be connected
+//and the command that will be send to the server, set up the connection,
+//send the command and returns server's response
+//
+//function returns output from the server to the specified command
+func SSHSendCommand(ctx context.Context, Client SSHClient, command string) (string, error) {
```
markdryan: should be client and not Client

bat/ssh.go
markdryan: There's not really much point in exporting this function as other packages won't be able to use it without also importing the ssh package as well.  We could say that for now we just export SSHSendCommand and assume that tests won't need to send more than one command.

bat/ssh.go
```diff
@@ -0,0 +1,85 @@
...
+			ssh.Password(WLclient.ClientPass),
+		},
+	}
+	Conection, err := ssh.Dial("tcp", ClientFullAddress, config)
+	return Conection, err
+}
+
+//SSHSendCommand  send a command to a specified server
+//
+//Recive a struct with the details of the server that will be connected
```
markdryan: Receive not Recive.

Note that this format of documentation where we have 3 paragraphs is only needed for functions that begin Test, i.e., for the actually test cases.  There's nothing stopping you using it for other functions as well but I just wanted to point out that you don't actually need to do this.

_release/bat/base_bat/base_bat_test.go
```diff
@@ -311,6 +326,160 @@ func TestDeleteAllInstances(t *testing.T) {
...
+		FWType:          "legacy",
+		Defaults:        defaults,
+		CloudConfigFile: "CirrosInit.yaml",
+		Disks:           []bat.Disk{disk},
+	}
+	y, err := yaml.Marshal(WLopts)
+	if err != nil {
+		t.Fatal("fail to read yaml file $v", err)
+	}
+	ioutil.WriteFile("CirrosWorkload.yaml", y, 0644)
```
markdryan: We should check the errors here.

_release/bat/base_bat/base_bat_test.go
```diff
@@ -311,6 +326,160 @@ func TestDeleteAllInstances(t *testing.T) {
...
+	defer cancelFunc()
+
+	// TODO:  The only options currently supported by the image service are
+	// ID and Name.  This code needs to be updated when the image service's
+	// support for meta data improves.
+	options := bat.ImageOptions{
+		Name: name,
+	}
+	//downaload cirros image
+	resp, err := http.Get("http://download.cirros-cloud.net/0.3.4/cirros-0.3.4-x86_64-disk.img")
```
markdryan: I'm just wondering why we download this image and whether this is a good idea.   Another alternative would be to check for an existing image with a known name and use this.  For example, Fedora Cloud Base 24-1.2.  If the image is not present you can just call t.Skip() and the test won't run.  This would make your test simpler.  There are other tests that verify that we can upload things to the image service.

bat/ssh.go
arturo-garcia-garcia: I see that SSH  have a parameter for timeout, defaulted to 0, which means no timeout, but we can set it to any value we want (i.e Timeout: time.Duration(time.Second * 5), set timeout to 5 secs),I can add a check if no value is provided (like the port), set it to 5 secs I think this is better than add a new Dial function, unless there are some extra benefit of using context

bat/ssh.go
arturo-garcia-garcia: Fixed, now it return the error instead of panic

bat/ssh.go
arturo-garcia-garcia: Added to close the connection, and thank you with the vendoring :)

bat/ssh.go
arturo-garcia-garcia: Fixed

bat/ssh.go
arturo-garcia-garcia: fixed

bat/ssh.go
arturo-garcia-garcia: Updated, I'll keep this in mind for the future, this time it help me to remember what the function should be doing

bat/ssh.go
markdryan: We need the extra dial function I'm afraid.  In fact I'm surprised it doesn't already exist in the ssh package.   The reason is that each BAT test creates a context at the start of the test and they use this to time out all the calls they make.  This lets us set a timeout per test rather than per call.  The function should be easy to write though.

```
func DialSSH(network, addr string, config *ClientConfig, ctx context.Context) (*ssh.Client, error) {
	conn, err := net.DialContext(ctx, network, addr)
	if err != nil {
		return nil, fmt.Errorf("Failed to connect to ssh server: %v", err)
	}
	c, chs, r, err := ssh.NewClientConn(conn, addr, config)
	if err != nil {
		return nil, fmt.Errorf("Failed to create new SSH client connection : %v", err)
	}
	return ssh.NewClient(c, chs, r), nil
}
```

or something like that.

arturo-garcia-garcia: Ok, then I'll add the new function, quick question, should it be in the ssh.go file that I created (I think this is the right place) or in the client.go from the crypto package?

markdryan: Well in can go in ssh.go for now but we could try to submit it upstream.  I'm really surprised it doesn't already exist.  I wonder if there's a bug tracker for that repo.

arturo-garcia-garcia: ok, let me add it to ssh.go in the mean time, then we can look to send it upstream

arturo-garcia-garcia: Done, dialSSH function added p ssh.go

_release/bat/base_bat/base_bat_test.go
arturo-garcia-garcia: I download the cirros image just to be safe due the performance issue using the image service, but i think will be better using the an existing image, I'll update this section

_release/bat/base_bat/base_bat_test.go
arturo-garcia-garcia: Done

markdryan:
@arturo-garcia-garcia it's a little hard to review the latest version of this patch as it's not rebased on master and includes tons of changes from other patches.   Could you rebase and re-push.

arturo-garcia-garcia:
Hi @markdryan , I'm sorry all the confusion, I made a huge mess with my branch (wrong set of remotes, missing pull, etc ) and it's taking me forever to get all straight, I'm going to edit this PR to only include the ciao/bat/ssh.go file, later I'll create a new one to add the rest of the tests (also my test env dies and I can't validate my code is right)

arturo-garcia-garcia:
I have some issues with my repositories and all get screwed, I'll create a new PR once all is clean and up to date

Issue #1038, opened 2017-01-20
Instances remain forever active when their compute node goes down


markdryan:
Here's a log generated using SingleVM

```
$ ciao-cli instance add --workload=ab68111c-03a6-11e6-87de-001320fb6e31 --instances=1
Created new (pending) instance: 589fd808-151b-4959-8a18-8c486806dcdc

$ ciao-cli node list --compute
Compute Node 1
	UUID: f0286130-1d76-4234-b59b-2fd8cf1fc435
	Status: READY
	Load: 0
	Available/Total memory: 5566/6975 MB
	Available/Total disk: 52250/60921 MB
	Total Instances: 1
		Running Instances: 1
		Pending Instances: 0
		Paused Instances: 0

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 589fd808-151b-4959-8a18-8c486806dcdc active 172.16.0.2 198.51.100.70 33002

$ sudo killall ciao-launcher

$ ciao-cli node list --compute

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 589fd808-151b-4959-8a18-8c486806dcdc active 172.16.0.2 198.51.100.70 33002

$ ciao-cli instance stop --instance 589fd808-151b-4959-8a18-8c486806dcdc
Instance 589fd808-151b-4959-8a18-8c486806dcdc stopped

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 589fd808-151b-4959-8a18-8c486806dcdc active 172.16.0.2 198.51.100.70 33002

[ Instance isn't stopped and cannot be manipulated in any way ever again ]

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 589fd808-151b-4959-8a18-8c486806dcdc active 172.16.0.2 198.51.100.70 33002

$ ~/local/run_launcher.sh &

[ Now run-launcher actually wipes the launcher state so the instance is deleted.  But it still appears in ciao-cli instance list ]

$ ciao-cli instance stop --instance 589fd808-151b-4959-8a18-8c486806dcdc
Instance 589fd808-151b-4959-8a18-8c486806dcdc stopped

$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 589fd808-151b-4959-8a18-8c486806dcdc active 172.16.0.2 198.51.100.70 33002
```

```

[...]

PR #1226, opened 2017-03-06
Implement instance migration

markdryan:
This commit implements migration of instances running on compute nodes.  It works as follows.

When the user stops an instance, either from the ui or by typing ciao-cli instance stop, [controller](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-221fd79430ab2bcdbe15305ac482b808R478) issues an SSNTP DELETE command, and not an SSNTP STOP command as it used to do in the past.   It sets a new [flag](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-3a7e0c030d17a05304dab60236cca86eR32) that has been added to the DELETE payload called migration, to true.  Scheduler delivers the DELETE command to the launcher instance running on the appropriate compute node which duly deletes the instance entirely from the node.  When processing the delete command it notices that the migration flag was set to true and rather than sending an InstanceDeleted SSNTP event it sends a new InstanceStopped SSNTP event.  When controller [processes](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-8d268b1ee13fa436f60ebc3d2d1ce9cfR1134) this event it sets the instance's status to exited, removes the association with the compute node, updates the tenant resources, and generates a fake statistic that indicates that the instance has indeed exited.  At this point the instance exists only in controller's database.  It has no state on any other node in the cluster.

At some later point the user will choose to restart the instance, via ciao-cli instance restart.  Controller, rather than sending a SSNTP RESTART command, sends a START command instead.  In order to construct this command [it](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-221fd79430ab2bcdbe15305ac482b808R490) needs to reconstruct the START payload from data stored in controller's database.  To be able to do this I needed to add two new fields to the instance [table](https://github.com/01org/ciao/pull/1226/files#diff-0da47715aa2ffec97772fa62edb63530R124).  When controller sends the START command it also sets a new flag in the START payload called migration.  This is needed to correctly handle errors.  It also sets the instance state from exited to pending.  The START command is routed to a compute node, which may or may not be the same node that had previously hosted the instance, and is processed by launcher like any other START command.   Assuming the START command succeeds the instance is migrated.

If the START command fails we need to be a little careful.  In almost all cases when controller receives a StartFailure error it automatically deletes the instance.  We don't want to do this when restarting an instance.  For this reason a new [migration](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-b469cd1551434c09c6ae780985ec5cf0R94) field has been added to the StartFailure payload.  This field is set to true by launcher if the START command that generated the error had its migration field set to true.   Controller then has all the information it needs to decide whether or not to delete an instance that fails to start.  It performs the check [here](https://github.com/01org/ciao/compare/master...markdryan:migration?expand=1#diff-221fd79430ab2bcdbe15305ac482b808R286).

The SSNTP START and STOP commands are no longer used and will be removed in a later PR.

To test simply create a new instance using the web-ui or ciao-cli.  Then stop it, restart it and check that the instance still exists and has the expected state.  Ideally, you'll tests this on a multiple node cluster and note that the instance's HostID changes from one restart to another.

You can also run the new BAT tests

TestListComputeNodes
TestStopRestartInstance
TestDeleteStoppedInstance
TestStartBadWorkload


mcastelino:
@markdryan can this be tested in single vm with just stop, start

markdryan:
@mcastelino Yes it can.   In some ways testing on single VM is easier as you can easily verify that the instance has been deleted from launcher.  For example,

Start a new instance

```
$ ciao-cli instance add --workload 8ce0da93-4aae-49cd-8b64-e8ff48dd4a4a
Created new (pending) instance: 7e70c0ff-db90-4406-809f-0c4176502321
```

Check it's running

```
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 7e70c0ff-db90-4406-809f-0c4176502321 active 172.16.0.2 198.51.100.89 33002
```

Check the compute node and other instance details.

```
$ ciao-cli instance show --instance 7e70c0ff-db90-4406-809f-0c4176502321
	UUID: 7e70c0ff-db90-4406-809f-0c4176502321
	Status: active
	Private IP: 172.16.0.2
	MAC Address: 02:00:ac:10:00:02
	CN UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Image UUID: 
	Tenant UUID: b340352e7105411a81f787983f91ceee
	SSH IP: 198.51.100.89
	SSH Port: 33002
	Volume: 5717ab90-58d7-4dd5-a85d-1348536fa19e
```

Check the resources consumed by the host node

```
$ CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
Compute Node 1
	UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Status: READY
	Load: 0
	Available/Total memory: 4907/6975 MB
	Available/Total disk: 52220/60921 MB
	Total Instances: 1
		Running Instances: 1
		Pending Instances: 0
		Paused Instances: 0
```

Stop the instance

```
$ ciao-cli instance stop --instance 7e70c0ff-db90-4406-809f-0c4176502321
Instance 7e70c0ff-db90-4406-809f-0c4176502321 stopped
```

Check its status

```
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 7e70c0ff-db90-4406-809f-0c4176502321 exited 172.16.0.2 198.51.100.89 33002
```

Have a detailed look at the instance.  Note that the CN UUID is now blank

```
$ ciao-cli instance show --instance 7e70c0ff-db90-4406-809f-0c4176502321
	UUID: 7e70c0ff-db90-4406-809f-0c4176502321
	Status: exited
	Private IP: 172.16.0.2
	MAC Address: 02:00:ac:10:00:02
	CN UUID: 
	Image UUID: 
	Tenant UUID: b340352e7105411a81f787983f91ceee
	SSH IP: 198.51.100.89
	SSH Port: 33002
	Volume: 5717ab90-58d7-4dd5-a85d-1348536fa19e
```

Have a look in launcher's instance directory.  Our instance 7e70c0ff-db90-4406-809f-0c4176502321 is not listed.  Launcher no longer maintains any state on this node.

```
$ ls -1 /var/lib/ciao/instances/
df802abc-5cf9-4cda-adaa-447047cb3bf5
```

Check the node once more.  Note that no instances are assigned to it.

```
$ CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
Compute Node 1
	UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Status: READY
	Load: 0
	Available/Total memory: 5167/6975 MB
	Available/Total disk: 52219/60921 MB
	Total Instances: 0
		Running Instances: 0
		Pending Instances: 0
		Paused Instances: 0
```

Restart the instance

```
$ ciao-cli instance restart --instance 7e70c0ff-db90-4406-809f-0c4176502321
Instance 7e70c0ff-db90-4406-809f-0c4176502321 restarted
```

Check its status

```
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 7e70c0ff-db90-4406-809f-0c4176502321 active 172.16.0.2 198.51.100.89 33002
```

Get the instance details.  Note that the CN UUID field is now field in.  If you are testing on a real cluster you would expect this UUID to be different from the previously assigned one.  On singleVM it will of course be the same.

```
$ ciao-cli instance show --instance 7e70c0ff-db90-4406-809f-0c4176502321
	UUID: 7e70c0ff-db90-4406-809f-0c4176502321
	Status: active
	Private IP: 172.16.0.2
	MAC Address: 02:00:ac:10:00:02
	CN UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Image UUID: 
	Tenant UUID: b340352e7105411a81f787983f91ceee
	SSH IP: 198.51.100.89
	SSH Port: 33002
	Volume: 5717ab90-58d7-4dd5-a85d-1348536fa19e
```

Check the node status again.  Note that one instance is now running on this node.

```
$ CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
Compute Node 1
	UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Status: READY
	Load: 0
	Available/Total memory: 4927/6975 MB
	Available/Total disk: 52208/60921 MB
	Total Instances: 1
		Running Instances: 1
		Pending Instances: 0
		Paused Instances: 0
```

Check the list of instances maintained by launcher.  We now have two.

```
$ ls -1 /var/lib/ciao/instances/
7e70c0ff-db90-4406-809f-0c4176502321
df802abc-5cf9-4cda-adaa-447047cb3bf5
```

Kill the qemu process associated with the instance.

```
$ sudo kill 32428
```

Check the instance's state

```
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 7e70c0ff-db90-4406-809f-0c4176502321 exited 172.16.0.2 198.51.100.89 33002
```

Get the instance's detailed information.  Note that CN UUID is blank.

```
$ ciao-cli instance show --instance 7e70c0ff-db90-4406-809f-0c4176502321
	UUID: 7e70c0ff-db90-4406-809f-0c4176502321
	Status: exited
	Private IP: 172.16.0.2
	MAC Address: 02:00:ac:10:00:02
	CN UUID: 
	Image UUID: 
	Tenant UUID: b340352e7105411a81f787983f91ceee
	SSH IP: 198.51.100.89
	SSH Port: 33002
	Volume: 5717ab90-58d7-4dd5-a85d-1348536fa19e
```

Check the node info.  The node should be running 0 compute instances once more.

```
$ CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
Compute Node 1
	UUID: aa658746-bed6-42f8-a217-5805fd280e46
	Status: READY
	Load: 0
	Available/Total memory: 5143/6975 MB
	Available/Total disk: 52205/60921 MB
	Total Instances: 0
		Running Instances: 0
		Pending Instances: 0
		Paused Instances: 0
```

Make sure launcher has really deleted the instance.

```
$ ls -1 /var/lib/ciao/instances/
df802abc-5cf9-4cda-adaa-447047cb3bf5
```

Check the instance can be restarted.

```
$ ciao-cli instance restart --instance 7e70c0ff-db90-4406-809f-0c4176502321
Instance 7e70c0ff-db90-4406-809f-0c4176502321 restarted
```

Yep it's there.

```
$ ciao-cli instance list
# UUID                                 Status Private IP SSH IP        SSH PORT
1 7e70c0ff-db90-4406-809f-0c4176502321 active 172.16.0.2 198.51.100.89 33002

$ ls -1 /var/lib/ciao/instances/
7e70c0ff-db90-4406-809f-0c4176502321
df802abc-5cf9-4cda-adaa-447047cb3bf5
```

markdryan:
@mcastelino I did a quick test of external ip.  I simply added the following code to my local copy of verify.sh and it seems to work.  I'm not sure whether to submit this or whether I need some cleaner code.  Even this snippet was stretching my bash skills.

```
+echo "Stopping test instance $testinstance"
+
+"$ciao_gobin"/ciao-cli instance stop -instance $testinstance
+
+sleep 5
+
+echo "Restarting test instance $testinstance"
+"$ciao_gobin"/ciao-cli instance restart -instance $testinstance
+
+checkExtIPConnectivity
+
```

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1140,6 +1166,90 @@ func (ds *Datastore) DeleteInstance(instanceID string) error {
...
+	ds.tenantsLock.Unlock()
+
+	// we may not have received any node stats for this instance
+	if oldNodeID != "" {
+		ds.nodesLock.Lock()
+		delete(ds.nodes[oldNodeID].instances, instanceID)
+		ds.nodesLock.Unlock()
+	}
+
+	instanceStat := types.CiaoServerStats{
```
markdryan: I should probably factorize this code as there is some very similiar code in the preceeding function.

markdryan:
@tpepper There's a small scheduler [change](https://github.com/01org/ciao/pull/1226/files#diff-69db8bfe90bdb1f233fd0ad2bf5cb4afR1096) to route the new InstanceStopped event

markdryan:
@kaccardi I'd also like to draw your attention to https://github.com/01org/ciao/pull/1226/commits/6b438b7df55e8b0382b80b1477abfa3e8e73f7a1.  I noticed while testing that controller in memory cache was not correctly reconstructed from the DB when killing it and restarting it.  This commit fixes the issue for me.  I was testing whether a stopped instance could be restarted after controller had been killed and itself restarted.  With this commit it can.

ciao-controller/command.go
```diff
@@ -33,21 +33,26 @@ func (c *controller) evacuateNode(nodeID string) error {
...
 	if err != nil {
 		return err
 	}
 
-	if i.NodeID == "" {
-		return types.ErrInstanceNotAssigned
-	}
-
 	if i.State != "exited" {
 		return errors.New("You may only restart paused instances")
```
rbradford: Can we clarify this message. "paused" is never used elsewhere.

payloads/start.go
```diff
@@ -242,6 +242,10 @@ type StartCmd struct {
 	// Storage contains all the information required to attach or boot
 	// from storage for the new instance.
 	Storage []StorageResource `yaml:"storage,omitempty"`
+
+	// Migration is set to true if the payload represents a request to
+	// restart an existing instance on a new node.
+	Migration bool
```
rbradford: Again, not convinced that the name Migration is the best here as for me migration means strictly moving from A to B.

rbradford: And further could this be resolved by tracking the state in the the controller database rather than relaying information back from launcher that controller told it in the first place?

markdryan: @kaccardi What do you think about this proposal? I did things this way as it seemed more in line with controller's design, in that it relies on launcher to provide information about the state of the instances, although in some [ways](https://github.com/01org/ciao/pull/1226/files#diff-8d268b1ee13fa436f60ebc3d2d1ce9cfR1115) I'm already breaking the existing model by explicitly setting the state of instances inside controller.

But we could add new instance states, e.g., stopping, and set this in controller.  This would simplify launcher but would require more changes to controller and changes to the way we process stats and may break other things.

If you would prefer this approach I can investigate to see how hard this would be.  It may make sense to do this as a separate PR, in which we introduce other states such as deleting, move the instance status from temporary storage, check the stats for missing instances (in case we missed delete events, etc.)

ciao-controller/command.go
markdryan: This message was not introduced by this PR.  Paused is used elsewhere, for example in the node summaries, but I was planning to remove this in a later PR.

payloads/stop.go
```diff
@@ -25,6 +25,11 @@ type StopCmd struct {
 	// running.  This information is needed by the scheduler to route
 	// the command to the correct CN/NN.
 	WorkloadAgentUUID string `yaml:"workload_agent_uuid"`
+
+	// Migration is true if the instance is being migrated to another node.
+	// In this case the delete command should only delete the instance from
+	// the node to which it is sent and not the entire cluster.
+	Migration bool
```
markdryan: Sure.  Assuming we keep these flags I'll change their names to Stop and Restart.

ciao-launcher/instance.go
```diff
@@ -57,12 +57,12 @@ type insStartCmd struct {
 	cfg      *vmConfig
 	rcvStamp time.Time
 }
-type insRestartCmd struct{}
 type insDeleteCmd struct {
-	suicide bool
-	running ovsRunningState
+	suicide         bool
```
mcastelino: @markdryan Do you mind documenting the fields here. As they convey quite a bit of information in the code.

ciao-launcher/instance.go
```diff
@@ -393,7 +362,8 @@ DONE:
 			id.statsTimer = nil
 			id.ovsCh <- &ovsStateChange{id.instance, ovsStopped}
 			id.st = nil
-			id.unmapVolumes()
+			killMe(id.instance, false, true, id.doneCh, id.ac, &id.instanceWg)
```
mcastelino: @markdryan we moved from unmap to killing the VM?

mcastelino:
@markdryan question/observation about impact of stopping.. If you follow this sequence

```
ciao-cli instance add --workload a6ecf84a-280f-4508-bc98-3fd3dae997f7
ciao-cli instance show --instance  08d6ef66-f19b-4c29-8b44-0bf6f2286a0b
CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
```
you now see
```
Compute Node 1
        UUID: 2ff72bff-3b20-4c94-a659-d26d35d77f01
        Status: READY
        Load: 1
        Available/Total memory: 6382/7982 MB
        Available/Total disk: 50603/60921 MB
        Total Instances: 1
                Running Instances: 1
                Pending Instances: 0
                Paused Instances: 0
```

If you now stop it
```
ciao-cli instance stop --instance 08d6ef66-f19b-4c29-8b44-0bf6f2286a0b
CIAO_USERNAME=admin CIAO_PASSWORD=$CIAO_ADMIN_PASSWORD ciao-cli node list --compute
```
you see
```
Compute Node 1
        UUID: 2ff72bff-3b20-4c94-a659-d26d35d77f01
        Status: READY
        Load: 0
        Available/Total memory: 6584/7982 MB
        Available/Total disk: 50601/60921 MB
        Total Instances: 0
                Running Instances: 0
                Pending Instances: 0
                Paused Instances: 0
```

Should there be a field for stopped instances?

mcastelino:
@markdryan The behavior of stop/restart in the case of containers is different from VMs.

We do not actually restart the same container. We just restart the same image but any modifications done to that image are not preserved. However the network properties of the image are retained.

This should be acceptable as you should not preserve state in the image anyway, but we need to document this behavior. So in the case of containers we support starting of a new containers with network properties preserved.

mcastelino:
@markdryan @rbradford should we merge this as it works well, and perform any rework on top of this. This is a large PR already.

ciao-launcher/instance.go
markdryan: Sure no problems.

ciao-launcher/instance.go
markdryan: This piece of code gets called when the VM shuts down.  Pre-migration the instance continues to reside on the compute node even though it is no longer running, so the instance loop keeps running.

With migration instances that shutdown or just finish executing are automatically evacuated.  Thats what the call to killMe does.  It basically inserts a SSNTP DELETE command into launcher's command queue so the instance's state gets deleted.  When this DELETE command is processed by the instance go routine in the deleteCommand function the volumes are unmapped so there is no longer any need to unmap them here.

markdryan:
@mcastelino 

> We do not actually restart the same container. We just restart the same image but any modifications done to that image are not preserved. However the network properties of the image are retained.

Yes this is true.  A new image is created that runs the same workload.  However, any volumes mapped to the old container will be mapped to the restarted container after migration.  So as long as you write your data to a volume all is well.  I wonder where the best place to document this is?

markdryan:
> @markdryan @rbradford should we merge this as it works well, and perform any rework on top of this. This is a large PR already.

We really need @kaccardi to review the controller parts.  Also, I have a few little things to fix.  So if @kaccardi is okay with the general approach we can merge once I've made the relevant fixes.

payloads/start.go
markdryan: > although in some ways I'm already breaking the existing model by explicitly setting the state of instances inside controller.

Thinking about this a bit more I don't think this PR does violate controller's current approach to instance state, in which an instances state is something that is owned by launcher.  It's true that this PR sets the instance state to Stopped but this is done as a response to an event received from launcher.  It could just as easily be done via STATS.

Getting rid of the Migration flags and InstanceStopped, introducing extra instance states in controller and having controller decide an instance's state (e.g., having it ignore instance states received in STATS from launcher) won't work in the current implementation as state is stored in /tmp.  A controller node reboot could result in the accidental deletion of a instance that the user restarted prior to the reboot.

That said the implementation in this PR has a state bug.  Stop an instance and reboot controller.  After the reboot, the instance's set is forever pending and it cannot be restarted.  The instance is no longer present on the compute node so launcher won't send any more stats for it.  This could be fixed by setting the status of instances that have not reported in for a fixed period of time to exited.  This would also fix the existing issue https://github.com/01org/ciao/issues/1038, providing the node doesn't come back up.

Moving instance state to permanent storage would also fix these problems.

I'm happy with either option, although I'm not sure how much of this work should be done as part of this PR.  @kaccardi do you have a preference or indeed a better way of solving these problems?

markdryan:
Should there be a field for stopped instances?

No.  Once an instance is stopped it no longer has any association with a compute node.  What needs to happen is that the Paused Instances field needs to be removed.  I was planning to do this in a separate PR.




---------------------------------------------------------------------
Intel Corporation (UK) Limited
Registered No. 1134945 (England)
Registered Office: Pipers Way, Swindon SN3 1RJ
VAT No: 860 2173 47

This e-mail and any attachments may contain confidential material for
the sole use of the intended recipient(s). Any review or distribution
by others is strictly prohibited. If you are not the intended
recipient, please contact the sender and delete all copies.

rbradford:
> @markdryan @rbradford should we merge this as it works well, and perform any rework on top of this. This is a large PR already.

Yes. My feedback was not intended to be a blocker. There are lots of reasons why this is the best approach given the current architecture within ciao-controller. I've spoken at great length offline with @markdryan about this.

PR #1229, opened 2017-03-07
ciao-down: clearcontainers : fix docker version to 1.12.1

mcastelino:
Fix the docker version to 1.12.1 for now.
Note: docker-engine is now docker-ce
The cloud-init performs the setup needed for docker-ce but
currently keeps using the older docker-engine.

### DIFF
--- a/testutil/ciao-down/cloudinit.go
+++ b/testutil/ciao-down/cloudinit.go
@@ -333,12 +333,15 @@ runcmd:
  - groupadd docker
  - sudo gpasswd -a {{.User}} docker
  - curl -X PUT -d "Installing apt-transport-https and ca-certificates" 10.0.2.2:{{.HTTPServerPort}}
- - {{template "ENV" .}}apt-get install apt-transport-https ca-certificates
+ - {{template "ENV" .}}sudo apt-get -y install  apt-transport-https ca-certificates
  - {{template "CHECK" .}}
 
  - curl -X PUT -d "Add docker GPG key" 10.0.2.2:{{.HTTPServerPort}}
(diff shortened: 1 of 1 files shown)


mcastelino:
@markdryan this has support for CE. But intentionally installing 1.12.1 for now.
Similar changes need to be done for the ciao VM.

@singlevm:~$ apt-cache policy docker-engine
docker-engine:
  Installed: 1.12.1-0~xenial
  Candidate: 17.03.0~ce-0~ubuntu-xenial

testutil/ciao-down/cloudinit.go
```diff
@@ -332,29 +332,32 @@ runcmd:
 
  - groupadd docker
  - sudo gpasswd -a {{.User}} docker
- - curl -X PUT -d "Installing apt-transport-https and ca-certificates" 10.0.2.2:{{.HTTPServerPort}}
- - {{template "ENV" .}}apt-get install apt-transport-https ca-certificates
+ - curl -X PUT -d "Installing curl, apt-transport-https and ca-certificates" 10.0.2.2:{{.HTTPServerPort}}
+ - {{template "ENV" .}}sudo apt-get -y install  apt-transport-https ca-certificates  curl
```
markdryan: Isn't curl already installed?  It's being used on the preceding line.

markdryan:
@mcastelino Thanks for figuring out how to do this.  I'll have a play tomorrow and will create a PR for ciao based on this one.

markdryan:
@mcastelino This is working fine for me (better than before in fact where the containers I launched never exitted).   My only comment is that we don't need to install curl.

testutil/ciao-down/cloudinit.go
mcastelino: @will fix this.

will:
I would but like I don't use docker or know who you are so

PR #1233, opened 2017-03-08
ciao-controller: Fix workload's image ID reference in ciao-cli output

### LINKED ISSUE #1232: ciao-controller: Image ID is not shown when workloads listing
ImageID is not shown when running:

```
ciao-cli workload list
```

Steps to reproduce it
----------------------
- Create a workload with this https://github.com/01org/ciao/blob/master/ciao-cli/examples/fedora_cloud.yaml
- Run ``ciao-cli workload list``

You will see this output:
```
Workload 1
        Name: Fedora VM no storage
        UUID:3f875e60-da16-46c9-8a9a-dfc60d790611
        Image UUID: 
        CPUs: 2
        Memory: 128 MB
```

obedmr:
This commit fixes OpenStack's compatibility for Disk variable, it's not the Image ID, it refers to the primary disk's size according to OpenStack [Flavor Details API method](https://developer.openstack.org/api-ref/compute/?expanded=list-flavors-with-details-detail#list-flavors-with-details).

A new CIAO specific API will be required in order to show image and storage details. This should be addressed in a separate PR.

Addresses part of #1232

### DIFF
--- a/ciao-controller/openstack_compute.go
+++ b/ciao-controller/openstack_compute.go
@@ -571,7 +571,6 @@ func buildFlavorDetails(workload types.Workload) (compute.FlavorDetails, error) 
 
 	details.OsFlavorAccessIsPublic = true
 	details.ID = workload.ID
-	details.Disk = workload.ImageID
 	details.Name = workload.Description
 
 	for r := range defaults {
--- a/bat/compute.go
+++ b/bat/compute.go
@@ -56,11 +56,10 @@ type Tenant struct {
 
 // Workload contains detailed information about a workload
 type Workload struct {
-	ID        string `json:"id"`
-	Name      string `json:"name"`
-	ImageUUID string `json:"disk"`
-	CPUs      int    `json:"vcpus"`
-	Mem       int    `json:"ram"`
+	ID   string `json:"id"`
+	Name string `json:"name"`
+	CPUs int    `json:"vcpus"`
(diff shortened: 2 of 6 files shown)


ciao-controller/openstack_compute.go
```diff
@@ -569,6 +569,10 @@ func buildFlavorDetails(workload types.Workload) (compute.FlavorDetails, error)
 	details.Disk = workload.ImageID
 	details.Name = workload.Description
 
+	if details.Disk == "" {
+		details.Disk = workload.Storage[0].SourceID
```
rbradford: This would be wrong if the workload used a volume rather than an image.

rbradford: It will also still be wrong for containers too.

obedmr: @rbradford ok got it, so, is this ok to have empty value in ``ciao-cli`` output when we use ciao-image provided image?

obedmr:
@rbradford a got it, showing disks information will be more useful and clear.

obedmr:
/me making proper changes to be in complience with openstack API https://developer.openstack.org/api-ref/compute/?expanded=list-flavors-with-details-detail#list-flavors-with-details

obedmr:
@rbradford I changed the change to remove the Image ID from current implementation.

Issue #1234, opened 2017-03-08
Track quotas for external IP


rbradford:
Map consumes, UnMap releases.

PR #1235, opened 2017-03-08
ciao-controller: Track external IP map/unmap in quota system

rbradford:
When an external IP is mapped for a tenant consume the resource and
release when IP unmapped.

Fixes: #1234

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -219,6 +219,8 @@ func (client *ssntpClient) unassignEvent(payload []byte) {
 		return
 	}
 
+	client.ctl.qs.Release(i.TenantID, payloads.RequestedResource{Type: payloads.ExternalIP, Value: 1})
+
 	msg := fmt.Sprintf("Unmapped %s from %s", event.UnassignedIP.PublicIP, event.UnassignedIP.PrivateIP)
 	client.ctl.ds.LogEvent(i.TenantID, msg)
 }
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -364,6 +364,8 @@ func (client *ssntpClient) assignError(payload []byte) {
 		glog.Warningf("Error unmapping external IP: %v", err)
 	}
 
+	client.ctl.qs.Release(failure.TenantUUID, payloads.RequestedResource{Type: payloads.ExternalIP, Value: 1})
+
 	msg := fmt.Sprintf("Failed to map %s to %s: %s", failure.PublicIP, failure.InstanceUUID, failure.Reason.String())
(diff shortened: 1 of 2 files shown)


mcastelino:
@rbradford LGTM. How do I test this?

markdryan:
I seem to be getting an resource leak and I don't understand why.  If I start a new cluster and run verify.sh it passes fine.  Then if I look at all the external-ips with

CIAO_USERNAME=admin CIAO_PASSWORD=giveciaoatry ciao-cli external-ip list

I see no external-ips.

However, when I do a 

$ ciao-cli quotas list

I see

Quotas for tenant: 
tenant-volumes-quota:           1 of unlimited
tenant-storage-quota:           1 of unlimited
tenant-instances-quota:         0 of unlimited
tenant-images-quota:            0 of unlimited
```tenant-external-ips-quota:      1 of unlimited```
tenant-vcpu-quota:              0 of unlimited
tenant-mem-quota:               0 of unlimited
tenant-vcpu-per-instance-limit: unlimited
tenant-mem-per-instance-limit:  unlimited
tenant-volume-size-limit:       unlimited

For some reason there is still one external-ip address consumed.  I've tried unmapping external-ips by hand and the counter does go down, so I'm not sure what verify.sh is doing.  I need to investigate further.

markdryan:
Okay, I've managed to reproduce the problem.  The verify.sh executed a map command that fails.  When this happens we leak a quota, e.g,

ciao-cli external-ip map --instance a0fc19a6-ef2c-41a5-9f3d-5f329a294ce1 --pool tes

I've written tes instead of test.  The mapping fails but the quota is consumed.

mcastelino:
@markdryan In verify.sh I delete the pool in the end
https://github.com/01org/ciao/blob/master/testutil/singlevm/verify.sh#L350

This maybe the reason for the issue.

mcastelino:
@rbradford @markdryan I make the external IP mapping fail using a synthetic error condition. The goal of that was to simulate the error where something has gone bad on the cluster. But yes, that IP should be reclaimed on error. I also ensure it can be re-used again.

rbradford:
I resolved some of the leaking external IP quotas. But after running verify.sh I find i'm still leaking 1. @mcastelino i'm very ignorant of the code around this, do you have any thoughts?

rbradford:
@markdryan @mcastelino Was able to resolve all the quota leaks from verify.sh now. This was a good test for the functionality!

mcastelino:
@rbradford do you mind adding a test in verify.sh to check for quota... as it will help both users of the quota API and provide a way to verify this w/o BAT.

verify.sh helps users understand how to use our API's from the command line. As quota is a new feature, it might help having a simple test.

PR #1236, opened 2017-03-09
qemu: Add PCI option to the NetDevice

sboeuf:
The existing NetDevice relies on virtio-net driver, but there is a
useful PCI variant which was not available: virtio-net-pci.
This patch adds this new driver and adds two parameters specific to
this: "bus" and "addr".

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 
 	"context"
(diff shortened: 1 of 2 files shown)


sboeuf:
@sameo please take a look a this patch and let me know if it is okay with you.

sameo:
cc @mcastelino

sameo:
LGTM

mcastelino:
@sboeuf makes sense. We use that in the legacy runtime today
https://github.com/01org/cc-oci-runtime/blob/master/src/hypervisor.c#L93

PR #1238, opened 2017-03-13
Instance migration refinements

markdryan:
This PR addresses some of the comments made on PR #1226.  It renames some variables, factorizes some code, adds comment to some confusing parts of launcher and fixes a nasty bug in scheduler that could have lead to restarted instances being deleted if the attempt to restart them failed in scheduler.


markdryan:
@tpepper Could you check the scheduler changes?

PR #1241, opened 2017-03-14
[RFC, DONTMERGE] Deployment: Add keystone dockerfile and travis integration

erick0z:
This PR adds all the necessary files to build the new ubuntu-based keystone container for Ciao. The image will be hosted under the ciaoproject/keystone repository in dockerhub.


erick0z:
@albertomurillo those packages should be removed, mb. Thanks!

_DeploymentAndDistroPackaging/keystone/README.md
```diff
@@ -0,0 +1,63 @@
+Keystone with SSL
+=================
+[![](https://images.microbadger.com/badges/image/ciaoproject/keystone.svg)](http://microbadger.com/images/clearlinux/keystone "Get your own image badge on microbadger.com")
+[![](https://images.microbadger.com/badges/version/ciaoproject/keystone.svg)](http://microbadger.com/images/clearlinux/keystone "Get your own version badge on microbadger.com")
+
```
obedmr: I think we should change this microbadgge

erick0z:
@obedmr, can you review?

erick0z:
@kaccardi @markdryan Can you take a look at this? Since I've modified the .travis,yml file It would be good to have your approval. @albertomurillo I've modified the ansible scripts, can you test the new keystone container in your environment?

markdryan:
@erick0z These changes will slow down the travis builds considerably.  They seem to double the build time.

Also, does this mean we now upload a new keystone container on every travis build, when keystone has changed?  We do three travis builds per merge so do we upload the new build three times.

Could we separate out these changes so that the run in a parallel self contained job separate from the rest of the travis checks?  I realise this might take some reworking of the travis scripts as they are not designed this way.

erick0z:
@markdryan I'm looking at the last release build and it skips the deployment section if the release.txt exists. Which means it will only run the after_deploy section once, so just one upload will occur. I will take a look on how to do the docker build/test in parallel.

erick0z:
@markdryan I don't think the keystone files will be modified very often, so the travis build times will remain the same.

markdryan:
@erick0z Okay, I see.  So normal builds where we just update ciao Go code won't be any slower.  

@tpepper What do you think?  Does it make sense to you to stick all of this in travis?  I was a bit shocked at first but I'm sort of coming around to the idea now.  If we don't put it in travis, where do we put it?

We do need to factorize our builds though.  For example, the swagger stuff and ansible checks should be done in separate builds as should the creation of the keystone container added by this PR.  However, I guess we could split things up in a separate PR.

tpepper:
What's the rationale for us building our own in the first place?  Is there not a suitable standard one existing that we could just 'docker pull'?  Does OpenStack not have that level of stability?

.travis.yml
```diff
@@ -10,7 +10,11 @@ go:
...
+    - secure: QLks228wsIZw1S/LSn6arcUnyVD+OqTw6IDWNxdbtVUZ7MTIUsq3qXVSKLN+LY7T9mP02qkGfZNNyGN5vcfPSkXeCuOnnnFjIhhjHjpOvSrCcjPZtJOXmu1orYApV9eSwL2l7qnWf5s9Rfpi1OjL0ElZcQ6/t0tL0aY7KZsrBFTxB4SMP/BkXreqP73bRvhoQfidc6/u9PR1I/Ngug3tEVNKST6ykT0hrn49EKxPyCIweVgwkXm1Wji1ZzKut5iB40nw+3I0fiPDP+Sevmqz8CIMDwJZDgoPGFgJon+cY+hB3kviahhGK6h3Jg75YA61qWmVfVOBQT6lhhl2TD1haPzYYntkv05q76VY0RqBwbgscEEWEc6mvlChdt0MgeReP6RBOhh+eDUMxykvGAFc4KNa20/cTL4QrignAfdu2OoGobM9Xi1C+W/vXirPWNreMjBT9Wba1Fxq/cHHVczqKEvy8fg4Xo0RDjBVUuEmpg/LOA6A61MXxzIOGINU7DmW+XVErvB4AGHlFPreI1UsYXriZzljhRPuGkeXfCK9/JWxwjAZKGQfKHV2Rp8G1uQVJ2XgHx4YWKzu4/gd1u4XItcFqtahJ7x/DUit0c7e28yOp7dy3c1WMT3+Y7ZXHTgBnNjeYQHXzhoHIj/7mdn/1bxAATmBVVNO3DP2TOua4zg=
```
tpepper: Is this still secure now that everybody on the web can read it?  Does it mean everybody can now push to docker hub as us?

_DeploymentAndDistroPackaging/keystone/README.md
```diff
@@ -0,0 +1,63 @@
...
+| 24620586335a473fb56fc2be2f6bfb53 | admin |
++----------------------------------+-------+
+```
+
+Environment Variables
+---------------------
+- ``IDENTITY_HOST``
+  Identity (Keystone) host
+- (Optional) ``KEYSTONE_ADMIN_PASSWORD``
+  Keystone Admin user password. Default: ``bb915e9ce0ae4b46e82a069b2ef0f8d7``
```
tpepper: Can this be left unset, and specified at container first start through an argument?

.travis.yml
bryteise: It uses https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml so it should be fine.

tpepper: So this is a travis feature: https://github.com/travis-ci/travis.rb#encrypt

Possibly magically secured environment variables which decrypt only for jobs on a specific repo.  I can guess that these two secrets are DOCKER_USER=ciaoproject and DOCKER_PASSWORD=somethingmaybesecret.  Then I just need a commit on master which reads and emites somewhere those environment variables at runtime, for example from a malicious bit of vendor unit test code?  And the password is public.

In the meantime, it's probably only known to Travis and the patch author?

_DeploymentAndDistroPackaging/keystone/README.md
obedmr: @tpepper, yes we can do that. @erick0z let's put this password auto-generated on Keystone first run.

.travis.yml
gorozco1: This is a travis feature, https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml the credentials can only be decrypted by the author and travis repo

obedmr: @tpepper this is a common practice and actually recommended in Travis. As it's said in the Travis documentation:
```
 Encrypted variables are not available to untrusted builds such as pull requests coming from another repository.
```

tpepper: Of course travis would recommend it.  We should still be skeptical, cautious.

The statement by @gorozco1 that "the credentials can only be decrypted by the author and travis repo" is not true.  They are decrypted by travis during any CI run on the repo for which they were encrypted, given any/all unit test code access to the variable once it is in the environment.  There's nothing preventing malicious code in tests from stealing/exposing this sensitive data.

_DeploymentAndDistroPackaging/keystone/README.md
mrkz: @tpepper If I'm not mistaken, this could be done, but then the `KEYSTONE_ADMIN_PASSWORD` variable wouldn't be optional and container will likely fail to start if this var is not set.
@erick0z, @obedmr   can you please confirm my guess?

.travis.yml
gorozco1: That's why code reviews are for..

obedmr:
@tpepper regarding an official OpenStack keystone image, there's no an Official Keystone docker image. The reason we're building it at our own, is that we're making sure that it's secured with SSL encryption, most of the available docker images are not SSL-enabled. Also, this image contains the latest stable Keystone release. 

This will not be updated with high frequency, only when a new Keystone release is avaiable  or in each 6-month cycle OpenStack project releases. The effort to upgrade will be minimal, in most of the cases, it will only require a docker image rebuild. Our test scripts make sure that keystone is functional for CIAO authentication requirements.

_DeploymentAndDistroPackaging/keystone/README.md
obedmr: @mrkz. yes, this variable is optional, we'll be auto-generating one in case this is not specified.

.travis.yml
markdryan: Perhaps if we separated the keystone stuff out into a separate build using matrix, we could restrict access to the unencrypted vars to the commands needed to create the container and nothing else.

obedmr: @erick0z could we do this separate build in travis as @markdryan suggest?, it would only go in action when a change is  performed in keystone docker files, if there's no change, it will just return an OK status.

erick0z: @obedmr Sure, I'm looking at it.

tpepper: Why do it in travis at all?

erick0z: @tpepper Following the decision to maintain everything that is related to deployment/packaging in this repo, this was the first option. IMO, this should have it's own repository.

erick0z:
@markdryan @tpepper @obedmr I've completely removed the dockerhub stuff from the travis file. The image build can be triggered (via dockerhub) from time to time to add the security updates. The dockerfilles remains here just as reference for those who want to build their own image.

PR #1245, opened 2017-03-16
ciao-down: clearcontainers: use 'go get' to retrieve sources

grahamwhaley:
You now need to have cc-oci-runtime in your GOPATH for it to build,
so use 'go get' to retrieve the sources rather than 'git clone', to
help ensure the source ends up in the right place.

Signed-off-by: Graham Whaley <graham.whaley@intel.com>


grahamwhaley:
Hi @markdryan. Sure, I knew the limitations when I made the PR. For reference, for the same issue over in Clear Containers runtime I noted the `get` would print a (benign) warning: https://github.com/01org/cc-oci-runtime/pull/752
Would it be acceptable to add such a warning print to this PR? The simple `go get` is much nicer to my eyes than the `git clone` with the $GOPATH in it - but, if the warning will not suffice, then I'll probably go with the `git clone`.
Annoyingly, `go get -d` exists to 'just download', but I guess it does slightly more than that, and still issues the warning.

Oh, and, no, we don't want to build the go source that lives in that tree during the get/download process - I believe it does not normally get installed into the $GOPATH/bin, but is installed as part of the larger build of that project to a different path.

markdryan:
@grahamwhaley 

You could do a

go get -x -v -d github.com/01org/cc-oci-runtime/...

This would download the code and place it where you want it but it won't build anything.

markdryan:
> This would download the code and place it where you want it but it won't build anything.

And won't print out any warnings.

grahamwhaley:
Great suggestion @markdryan - new commit pushed. Thanks.

markdryan:
@grahamwhaley You're missing the -d.  

go get github.com/01org/cc-oci-runtime/...

won't return an error but it will compile and install the code, which I thought you wanted to avoid.

grahamwhaley:
Apologies - let's try again...

PR #1247, opened 2017-03-17
ansible: docker: Use runc runtime in ClearLinux

albertomurillo:
Use runc runtime rather than cc-oci (clear containers)
when installing in ClearLinux


obedmr:
@tpepper could you take a look on it?

PR #1248, opened 2017-03-18
ansible: Add sysadmin-remote bundle as dependency for ClearLinux

albertomurillo:
This bundle contains docker-py which is required on the controller
node by ansible in order to launch containers


obedmr:
@tpepper could you take a look?

PR #1249, opened 2017-03-20
ciao-down: Ensure qemu-img is installed

markdryan:
This commit updates the deps.go file to ensure that qemu-img is installed on
Ubuntu and Fedora.  It is not part of the qemu-system-x86 package on these
distributions and so we cannot assume that it is installed by virtue of installing
qemu itself.


mcastelino:
LGTM

PR #1251, opened 2017-03-24
Add a file that will host common SSH operations and configuration

arturo-garcia-garcia:
This pull request will only include this file, I'm a bit busy
and can't take care of the rest of the tets for now, I'll
start a new pull request as soon as i can.

This file suppose to be included in PR #1219 , but I have some issues and need to clean it up

### DIFF
--- a/bat/ssh.go
+++ b/bat/ssh.go
@@ -1,0 +1,117 @@
+//
+// Copyright (c) 2016 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package bat
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net"
+	"time"
+
+	"golang.org/x/crypto/ssh"
+)
+
+//SSHClient contains details of the server to be connect
+type SSHClient struct {
+	ClientIP      string
+	ClientPort    string
+	ClientUser    string
+	ClientPass    string
+	ClientTimeout int64
+}
+
+func dialSSH(ctx context.Context, network, addr string, config *ssh.ClientConfig) (*ssh.Client, error) {
+	dialer := net.Dialer{
+		Timeout: config.Timeout,
+	}
+	conn, err := dialer.DialContext(ctx, network, addr)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to connect to ssh server: %v", err)
+	}
+	c, chs, r, err := ssh.NewClientConn(conn, addr, config)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create new SSH client connection : %v", err)
+	}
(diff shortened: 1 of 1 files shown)


bat/ssh.go
```diff
@@ -0,0 +1,108 @@
...
+//session sould be created succesfully and wait for commands
+func createSSHClient(ctx context.Context, c SSHClient) (*ssh.Client, error) {
+	//If no port is defined, it will be default to 22
+	if c.ClientPort == "" {
+		c.ClientPort = "22"
+	}
+	if c.ClientTimeout == 0 {
+		c.ClientTimeout = 5
+	}
+	ClientFullAddress := fmt.Sprint(c.ClientIP + ":" + c.ClientPort)
```
markdryan: Best to use net.JoinHostPort 

https://golang.org/pkg/net/#JoinHostPort

Otherwise the URL will be invalid if ClientIP is IPv6

bat/ssh.go
```diff
@@ -0,0 +1,108 @@
...
+	config := &ssh.ClientConfig{
+		User: c.ClientUser,
+		Auth: []ssh.AuthMethod{
+			ssh.Password(c.ClientPass),
+		},
+		Timeout: time.Duration(c.ClientTimeout) * time.Second,
+	}
+	Connection, err := dialSSH(ctx, "tcp", ClientFullAddress, config)
```
markdryan: You could use return dialSSH here.

bat/ssh.go
```diff
@@ -0,0 +1,108 @@
...
+	//Create a session
+	session, err := Connection.NewSession()
+	if err != nil {
+		return "", err
+	}
+	defer session.Close()
+	//send the command and get the result
+	var b bytes.Buffer
+	session.Stdout = &b
+	err = session.Run(command)
```
markdryan: We need to be able to cancel the Run command if the context is signalled.  Again this isn't that easy as the ssh package doesn't have explicit support for this.  But we can do it as follows:
```
ch := make(chan struct{})
go func() {
  err = session.Run(command)
  close(ch)
}
select {
   case <-ch:
   case <-ctx.Done():
    ssh.Signal(ssh.SIGQUIT)
    <-ch
}
```

So if our context is closed while Run is still running we send a signal to the remote process and then wait for Run to finish.  The hope is that the signal would interrupt the remote process started by Run.  Run would detect that it has finished and return.  It would be interesting to try this out which a sleep 100 command to make sure we could really interrupt it.

markdryan:
@arturo-garcia-garcia The travis build is failing due to some spelling errors

```
5.75s$ go list ./... | grep -v github.com/01org/ciao/vendor | xargs go list -f '{{.Dir}}/*.go' | xargs -I % bash -c "misspell -error %"

/home/travis/gopath/src/github.com/01org/ciao/bat/ssh.go:58:10: "sould" is a misspelling of "could"

/home/travis/gopath/src/github.com/01org/ciao/bat/ssh.go:58:27: "succesfully" is a misspelling of "successfully"
```

Also, you need to sign off your commit and the first line in the command should start with the component name, e.g.,

bat: Add a file ...

I've made a few comments on the code as well, but I think we're nearly there.

As previously mentioned, I'll take care of the vendoring once the patch is ready.

bat/ssh.go
arturo-garcia-garcia: Updated

bat/ssh.go
arturo-garcia-garcia: Hi @markdryan I just added the code you suggest, i'm setting and environment to test the interruption of the ssh call, if need, I'll update this section

bat/ssh.go
arturo-garcia-garcia: Done

Issue #1012, opened 2017-01-12
ciao-cli: Implement workload show


kaccardi:
Allow the user to see details about a workload via "ciao-cli workload show".

Issue #1134, opened 2017-02-09
BAT tests for workload show


kaccardi:
create BAT tests for workload show command for both privileged and non-privileged users

Issue #1013, opened 2017-01-12
ciao-controller: implement workload show API


kaccardi:
Implement a workloads API which allows the user to retrieve details about a specific workload.

Issue #1136, opened 2017-02-09
Allow privileged user to operated on tenant workloads


kaccardi:
Add API to allow a privileged user to show/list/delete workloads per tenant.

PR #1252, opened 2017-03-27
Add support for workload show

kaccardi:
Modify controller and ciao-cli to support a "workload show" command. This command will display a workload definition and config in yaml format.

Fixes: #1012 #1013 #1134

### DIFF
--- a/ciao-cli/workload.go
+++ b/ciao-cli/workload.go
@@ -39,6 +39,7 @@ var workloadCommand = &command{
 		"list":   new(workloadListCommand),
 		"create": new(workloadCreateCommand),
 		"delete": new(workloadDeleteCommand),
+		"show":   new(workloadShowCommand),
 	},
 }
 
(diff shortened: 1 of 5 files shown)


ciao-cli/workload.go
```diff
@@ -256,6 +257,52 @@ func optToReq(opt workloadOptions, req *types.Workload) error {
...
+	}
+
+	b, err := yaml.Marshal(opt)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	fmt.Println(string(b))
+	fmt.Println(w.Config)
+
```
markdryan: There's an extra new line here.

markdryan:
@kaccardi Works well for me.  One question.  Can you use this command to see other people's workloads?

kaccardi:
@markdryan no, it will only show the workload associated with a particular tenant for non-privileged and for privileged it will show only public workloads for now. the next phase of this project is to allow privileged users to operate on tenant private workloads (show/delete/list). #1136 is the issue associated with this task.

markdryan:
@kaccardi in which case it's good for me.   My only comment is an extra newline.  Feel free to ignore.

ciao-cli/workload.go
kaccardi: @markdryan thanks for catching that, I pushed an updated commit without the extraneous newline.

PR #1254, opened 2017-03-31
Templateutils: Separate template utility code into a separate package

markdryan:
This PR moves the template code in ciao-cli into a separate package called templateutils.  The old template code has been enhanced by:

1. Some documentation
2. More unit tests.
3. Some new functions, e.g., table, sort, cols, head, rows, tail.
4. The ability to select which functions to make available to script writers.
5. The ability to add your own functions.

Here's an example of some of the new functions in action

```
ciao-cli workload list -f '{{table (cols . "ID" "Name" "RAM" "Vcpus")}}'
ID                                   Name                         RAM     Vcpus   
b1650b0a-2813-4382-b08f-f9b45713ae13 Fedora test VM               128     2       
528c7f8d-c87e-4b55-8aae-b292c0213cfa Clear Linux test VM          128     2       
e56801a4-a6e5-4d9b-bdf7-4d0ebb389067 Ubuntu latest test container 128     2       
28cc2f7d-7371-4c81-912b-245877b186f2 Debian latest test container 128     2       
```


templateutils/templateutils.go
```diff
@@ -63,8 +64,27 @@ Some new functions have been added to Go's template language
   field for each struct on a new line , e.g.,
 
   {{select . "Name"}}
+- table outputs a table given an array or a slice of structs.  The table headings are taken
+  from the names of the structs fields.  Hidden fields and fields of type channel are ignored.
```
rbradford: > The table headings are taken from the names of the structs fields

Would it be nice to have a tag that could be associated with the struct fileld to contain the human readable column header?

templateutils/templateutils.go
```diff
@@ -108,6 +118,154 @@ var funcMap = template.FuncMap{
...
+	for index = 0; index < sTyp.NumField(); index++ {
+		fTyp = sTyp.Field(index)
+		if fTyp.Name == field {
+			break
+		}
+	}
+	if index == sTyp.NumField() {
+		panic(fmt.Sprintf("%s is not a valid field name", field))
+	}
+	fKind := fTyp.Type.Kind()
```
rbradford: Do you also need to handle pointers here? So that pointers to integers do the right thing?

templateutils/templateutils.go
```diff
@@ -344,6 +502,25 @@ func cols(obj interface{}, fields ...string) interface{} {
...
+	} else if len(direction) == 1 {
+		if direction[0] == "dsc" {
+			ascending = false
+		} else if direction[0] != "asc" {
+			panic("direction parameter must be \"asc\" or \"dsc\"")
+		}
+	}
+	assertCollectionOfStructs(obj)
+
+	// TODO: Need to copy
```
rbradford: >	// TODO: Need to copy

I think it's important to fix this one sooner rather than later.

templateutils/templateutils.go
rbradford: Okay scratch that: I see you fix it in a later patch.

templateutils/templateutils.go
rbradford: Okay scratch that: I see you fix it in a later patch.

templateutils/templateutils.go
```diff
@@ -135,6 +135,12 @@ func (c *Config) AddCustomFn(fn interface{}, name, helpText string) error {
 	return nil
 }
 
+// OptAllFNs enables all template extension functions provided by this package
+func OptAllFNs(c *Config) {
```
rbradford: Is name correct, should it be OptAllFns?

rbradford:
lgt

templateutils/templateutils.go
markdryan: Probably yes.  It's very much a corner case but it's easy to handle.

rbradford:
lgtm. only nit is that OptAllFNs -> OptAllFns naming.

templateutils/templateutils.go
```diff
@@ -90,7 +92,7 @@ func filterField(obj interface{}, field, val string, cmp func(string, string) bo
 	defer func() {
 		err := recover()
 		if err != nil {
-			fatalf("Invalid use of filter: %v", err)
+			panic(fmt.Errorf("Invalid use of filter: %v", err))
```
tpepper: Is it idiomatic in go for a package to panic instead of cleanly return an error?  In C, it's really horrible for a library to panic and result in the death of the caller.

templateutils/templateutils.go
```diff
@@ -29,22 +29,24 @@ import (
 	"text/template"
 )
 
-const templateFunctionHelp = `
-ciao-cli adds some new functions to Go's template language
+// TemplateFunctionHelp contains formatted documentation that describes the
+// additional functions that templateutils adds to Go's templating language.
+const TemplateFunctionHelp = `
+Some new functions have been added to Go's template language
```
tpepper: perhaps "...added beyond those already available in Go's template language"?

templateutils/templateutils.go
```diff
@@ -293,20 +304,41 @@ func generateIndentedUsage(buf *bytes.Buffer, i interface{}) {
 	formatType(buf, source.Bytes())
 }
 
-func generateUsageUndecorated(i interface{}) string {
```
tpepper: It's awesome that the export now forces documentation :)

templateutils/templateutils.go
```diff
@@ -63,8 +64,27 @@ Some new functions have been added to Go's template language
   field for each struct on a new line , e.g.,
 
   {{select . "Name"}}
+- table outputs a table given an array or a slice of structs.  The table headings are taken
+  from the names of the structs fields.  Hidden fields and fields of type channel are ignored.
+  The tabwidth and minimum column width are hardcoded to 8.  An example of table's usage is
+
+  {{table .}}
+- tablex is similar to table but it allows the caller more control over the table's appearance.
```
tpepper: I'm blanking on something more descriptive, but the name bugs me.  I'm not in love with "tablef" either, but that's all that comes to mind.

templateutils/templateutils.go
```diff
@@ -277,6 +296,54 @@ func tablex(obj interface{}, minWidth, tabWidth int, userHeadings ...string) str
...
+		if j == len(fields) {
+			continue
+		}
+
+		indicies = append(indicies, i)
+		newFields = append(newFields, field)
+	}
+
+	if len(indicies) != len(fields) {
+		panic("not all column names are valid")
```
tpepper: Do you also need a check for uniqueness of field names?

templateutils/templateutils.go
```diff
@@ -108,6 +118,154 @@ var funcMap = template.FuncMap{
...
+	reflect.Float64: func(v1, v2 interface{}) bool {
+		return v1.(float64) < v2.(float64)
+	},
+	reflect.Float32: func(v1, v2 interface{}) bool {
+		return v1.(float32) < v2.(float32)
+	},
+	reflect.String: func(v1, v2 interface{}) bool {
+		return v1.(string) < v2.(string)
+	},
+}
```
tpepper: Kinda makes you want a language with operator overloading!

templateutils/funcs.go
```diff
@@ -198,11 +206,11 @@ func findField(fieldPath []string, v reflect.Value) reflect.Value {
 	return f
 }
 
-func filterField(obj interface{}, field, val string, cmp func(string, string) bool) (retval interface{}) {
+func filterField(fnname string, obj interface{}, field, val string, cmp func(string, string) bool) (retval interface{}) {
```
tpepper: Do our CI tools complain these aren't fnName and retVal?

tpepper:
My only real problem with this PR is the risk Google's golang team sees the useful types of things Mark is doing and tries to hire him away from us ;)

templateutils/templateutils.go
markdryan: Good question but this is already handled as we iterate through the fields of the structure and simply check whether these fields are present in the list of arguments passed to the function.  Thus if you call {{cols . "Name" "Name"}}, len(indicies) will be 1 and len(fields) will be two, and we'll error out.

markdryan: As an example,

```
./tmplex -f  '{{cols . "Name" "Name"}}' stocks
Unable to execute template : not all column names are valid
```

templateutils/templateutils.go
markdryan: Yep, it's not great is it?

templateutils/funcs.go
markdryan: No they don't.  @rbradford spotted a similar naming issue that was not detected by golint.  I'll  fix the names.  Thanks.

markdryan: This turned out to be a good catch.  retval was not needed at all!  I've removed the identifier and renamed fnname.

Issue #1258, opened 2017-04-11
ciao-down: singlevm: External network access does not work


mcastelino:
The firewall rules within a ciao-down virtual machine prevent traffic from instances launched within ciao-down using ciao from gaining external connectivity. This is due to the absence for iptables rules that allow connectivity between the physical network and the ciao network.

PR #1259, opened 2017-04-11
ciao-down: single vm: Provide external network access for ciao instances

mcastelino:
The firewall rules within a ciao-down virtual machine prevent traffic
from instances launched within ciao-down using ciao from gaining
external connectivity. This is due to the absence for iptables rules
that allow connectivity between the physical network and the ciao
network.

Explicitly enable packet forwarding between the external network and
the logical ciao-down network by adding firewall rules.

However do not allow forwarding between the external network and ciao
network in the case of a bare metal deployment. This ensures that
user defined firewall rules are not violated. The user in this case
needs to explicitly enable forwarding of traffic from the external
network.

Fixes https://github.com/01org/ciao/issues/1258

### DIFF
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -13,6 +13,7 @@ sudo iptables -D FORWARD -i ciao_br -j ACCEPT
 sudo iptables -D FORWARD -i ciaovlan -j ACCEPT
 if [ "$ciao_host" == "singlevm" ]; then
 	sudo iptables -D FORWARD -i ens2 -j ACCEPT
+	sudo iptables -t nat -D POSTROUTING -o ens2 -j MASQUERADE
 fi
 sudo ip link del ciao_br
 sudo pkill -F /tmp/dnsmasq.ciaovlan.pid
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -3,12 +3,17 @@
 . ~/local/demo.sh
 
 ciao_gobin="$GOPATH"/bin
+ciao_host=$(hostname)
 sudo killall ciao-scheduler
 sudo killall ciao-controller
 sudo killall ciao-launcher
 sleep 2
 sudo "$ciao_gobin"/ciao-launcher --alsologtostderr -v 3 --hard-reset
-sudo iptables -D FORWARD -p all -i ciao_br -j ACCEPT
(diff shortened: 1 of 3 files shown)


mcastelino:
@kaccardi @matthewrsj this should fix the issues that you faced on providing internet connectivity for ciao instances running within ciao-down.

testutil/singlevm/cleanup.sh
```diff
@@ -3,12 +3,17 @@
...
 sudo killall ciao-scheduler
 sudo killall ciao-controller
 sudo killall ciao-launcher
 sleep 2
 sudo "$ciao_gobin"/ciao-launcher --alsologtostderr -v 3 --hard-reset
-sudo iptables -D FORWARD -p all -i ciao_br -j ACCEPT
+sudo iptables -D FORWARD -i ciao_br -j ACCEPT
+sudo iptables -D FORWARD -i ciaovlan -j ACCEPT
+if [ "$ciao_host" == "singlevm" ]; then
+	sudo iptables -D FORWARD -i ens2 -j ACCEPT
```
tpepper: I'm not seeing it in the code elsewhere, but is there something for a systemd/udev rule that insures the interface is named "ens2"?

mcastelino: Right now I am making this assumption as we use a canned ubuntu 16 image. Hence the check for singlevm. But to make it more generic, it would be best to determine the exiting physical interface by checking the route to the internet.

Something like

`ip -o route get 8.8.8.8 | cut -d ' ' -f 5`

That would be more generic

kaccardi:
@mcastelino I'm afraid it didn't work for me.

mcastelino:
@kaccardi @markdryan 

ciao-down runs using the QEMU SLIRP stack. However this SLIRP stack
only NATs the traffic that originates using the SLIRP assigned IP
address. Traffic originating from ciao instances however does not
get handled properly. The traffic will be NATed out, but on return
the SLIRP stack will drop the traffic destined to the ciao instanced.

Added a change to auto NAT in the case of ciao-down.
This change sets up ciao-down to always NAT all traffic headed out
of ciao-down to map to the SLIRP assigned IP.

This corresponds to our documentation
https://github.com/01org/ciao/blob/master/DeveloperQuickStart.md#cluster-external-network-access

kaccardi:
I tested and it works for me now.

PR #1260, opened 2017-04-19
ciao-down: Only use ciao-down ssh key

markdryan:
This commit modifies the ciao-down connect command to only use the
identity key generated by ciao-down.  We've had some reports that
users with a lot of keys in their .ssh directories are unable to
connect to a ciao-down VM, as ssh seems to be trying these keys
in preference to the key supplied by ciao-down, and is giving
up after a number of connection attempts have failed.


grahamwhaley:
Tested-By: me
Worked for me - I can see the extra argument in `ciao-down status`, and `ciao-down connect` now works with my 6 other keys still in place in my ~/.ssh

lgtm

PR #1274, opened 2017-05-10
Support naming instances

rbradford:
This PR implements some basic support for names for instances. These names must be tenant unique and will be used for the hostname inside the instance (and thus the DHCP created DNS entry.)

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -196,19 +196,10 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance, 
 		startTime := time.Now()
 
 		name := w.Name
-		if w.Instances > 1 {
-			name = fmt.Sprintf("%s-%d", name, i)
-		}
-
-		id, err := c.ds.ResolveInstance(w.TenantID, name)
-		if err != nil {
-			e = errors.Wrap(err, "error trying to resolve name")
-			continue
-		}
-
-		if id != "" {
-			e = fmt.Errorf("Instance name already in use: %s", name)
-			continue
+		if name != "" {
+			if w.Instances > 1 {
+				name = fmt.Sprintf("%s-%d", name, i)
+			}
 		}
 
 		instance, err := newInstance(c, w.TenantID, &wl, w.Volumes, name)
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -194,7 +194,13 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance, 
 
 	for i := 0; i < w.Instances && e == nil; i++ {
 		startTime := time.Now()
-		instance, err := newInstance(c, w.TenantID, &wl, w.Volumes)
+
+		name := w.Name
+		if w.Instances > 1 {
+			name = fmt.Sprintf("%s-%d", name, i)
+		}
+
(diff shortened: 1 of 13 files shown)


ciao-controller/instance.go
```diff
@@ -62,10 +62,22 @@ func isCNCIWorkload(workload *types.Workload) bool {
...
-	volumes []storage.BlockDevice) (*instance, error) {
+	volumes []storage.BlockDevice, name string) (*instance, error) {
 	id := uuid.Generate()
 
-	config, err := newConfig(ctl, workload, id.String(), tenantID, volumes)
+	if name != "" {
+		existingID, err := ctl.ds.ResolveInstance(tenantID, name)
+		if err != nil {
+			return nil, errors.Wrap(err, "error trying to resolve name")
+
```
rbradford: Fixed.

ciao-controller/instance.go
```diff
@@ -62,10 +62,22 @@ func isCNCIWorkload(workload *types.Workload) bool {
...
+		}
+
+		if existingID != "" {
```
rbradford: I've added a database constraint to enforce this beyond this test.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -2393,3 +2393,25 @@ func (ds *Datastore) GetQuotas(tenantID string) ([]types.QuotaDetails, error) {
...
+	t, ok := ds.tenants[tenantID]
+	if ok {
+		for _, i := range t.instances {
+			if i.Name == name || i.ID == name {
+				return i.ID, nil
+			}
+		}
+	}
+
+	if !ok {
```
rbradford: Fixed.

ciao-controller/command.go
```diff
@@ -247,6 +255,7 @@ func (c *controller) launchCNCI(tenantID string) error {
 		WorkloadID: workloadID,
 		TenantID:   tenantID,
 		Instances:  1,
+		Name:       "cnci",
```
kaccardi: will it be a problem to have multiple hosts named "cnci"?

kaccardi: what if a user decides to name their instance "cnci"? how will we use this hostname?

ciao-controller/openstack_compute.go
```diff
@@ -359,27 +361,29 @@ func (c *controller) CreateServer(tenant string, server compute.CreateServerRequ
 		nInstances = server.Server.MinInstances
 	}
 
+	if server.Server.Name != "" {
+		r := regexp.MustCompile("^[a-z0-9-]*$")
```
kaccardi: do you think we should put a limit on the length for the name as well? wondering if someone could be evil here.

ciao-controller/command.go
rbradford: As names are optional I can avoid this problem by just not setting a name. The CNCI already gets a special hostname in the code anyway.

ciao-controller/openstack_compute.go
rbradford: Yes, we should limit to HOST_NAME_MAX - 64.

rbradford:
@kaccardi Addressed your feedback. Thanks!

kaccardi:
LGTM

PR #1291, opened 2017-05-23
kubicle: A tool for installing kubernetes clusters on ciao

markdryan:
kubicle is a small command line tool that makes it easy to set up k8s clusters in ciao.  You need simply to provide a base image, which must be some version of Ubuntu server, and kubicle will do the rest.  In short it creates workloads for k8s master and worker nodes, creates instances from these workloads and joins these instances together to form a k8s cluster.  The tool accepts a number of options, such as the number of worker nodes to create, configuration information for those nodes and the user accounts to create on those nodes.

### DIFF
--- a/bat/external_ip.go
+++ b/bat/external_ip.go
@@ -1,0 +1,94 @@
+//
+// Copyright (c) 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package bat
+
+import "context"
+
+// ExternalIP contains information about a single external ip.
+type ExternalIP struct {
+	ExternalIP string `json:"external_ip"`
+	InternalIP string `json:"internal_ip"`
+	InstanceID string `json:"instance_id"`
+	TenantID   string `json:"tenant_id"`
+	PoolName   string `json:"pool_name"`
+}
+
+// CreateExternalIPPool creates a new pool for external ips.  The pool is created
+// using the ciao-cli pool create command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY,  CIAO_CONTROLLER,
+// CIAO_ADMIN_USERNAME, CIAO_ADMIN_PASSWORD.
+func CreateExternalIPPool(ctx context.Context, tenant, name string) error {
+	args := []string{"pool", "create", "-name", name}
+	_, err := RunCIAOCLIAsAdmin(ctx, tenant, args)
+	return err
+}
+
+// AddExternalIPToPool adds an external ips to an existing pool.  The address
+// is added using the ciao-cli pool add command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY,  CIAO_CONTROLLER,
+// CIAO_ADMIN_USERNAME, CIAO_ADMIN_PASSWORD.
+func AddExternalIPToPool(ctx context.Context, tenant, name, ip string) error {
+	args := []string{"pool", "add", "-name", name, ip}
+	_, err := RunCIAOCLIAsAdmin(ctx, tenant, args)
+	return err
+}
+
+// MapExternalIP maps an external ip from a given pool to an instance.  The
+// address is mapped using the ciao-cli external-ip map command.  An error
+// will be returned if the following environment are not set;  CIAO_IDENTITY,
+// CIAO_CONTROLLER, CIAO_USERNAME, CIAO_PASSWORD.
+func MapExternalIP(ctx context.Context, tenant, pool, instance string) error {
+	args := []string{"external-ip", "map", "-instance", instance, "-pool", pool}
+	_, err := RunCIAOCLI(ctx, tenant, args)
+	return err
+}
+
+// UnmapExternalIP unmaps an external ip from an instance.  The address is unmapped
+// using the ciao-cli external-ip unmap command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY, CIAO_CONTROLLER, CIAO_USERNAME,
+// CIAO_PASSWORD.
+func UnmapExternalIP(ctx context.Context, tenant, address string) error {
+	args := []string{"external-ip", "unmap", "-address", address}
+	_, err := RunCIAOCLI(ctx, tenant, args)
+	return err
+}
+
+// DeleteExternalIPPool deletes an external-ip pool.  The pool is deleted using
+// the ciao-cli pool delete command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY,  CIAO_CONTROLLER,
+// CIAO_ADMIN_USERNAME, CIAO_ADMIN_PASSWORD.
+func DeleteExternalIPPool(ctx context.Context, tenant, name string) error {
+	args := []string{"pool", "delete", "-name", name}
+	_, err := RunCIAOCLIAsAdmin(ctx, tenant, args)
+	return err
+}
+
+// ListExternalIPs returns detailed information about all the external ips
+// defined for the given tenant.  The information is retrieved using the
+// ciao-cli external-ip list command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY,  CIAO_CONTROLLER,
+// CIAO_ADMIN_USERNAME, CIAO_ADMIN_PASSWORD.
+func ListExternalIPs(ctx context.Context, tenant string) ([]*ExternalIP, error) {
+	var externalIPs []*ExternalIP
+	args := []string{"external-ip", "list", "-f", "{{tojson .}}"}
+	err := RunCIAOCLIAsAdminJS(ctx, tenant, args, &externalIPs)
+	if err != nil {
+		return nil, err
+	}
+
+	return externalIPs, nil
+}
--- a/bat/workload.go
+++ b/bat/workload.go
@@ -128,27 +128,7 @@ func createWorkload(ctx context.Context, tenant string, opt WorkloadOptions, con
 		return "", err
 	}
 
-	// send the workload create command to ciao-cli
-	args := []string{"workload", "create", "-yaml", "workload.yaml"}
-
-	var data []byte
-	if public {
-		data, err = RunCIAOCLIAsAdmin(ctx, tenant, args)
-	} else {
-		data, err = RunCIAOCLI(ctx, tenant, args)
-	}
-
-	if err != nil {
-		return "", err
-	}
-
-	// get the returned workload ID
-	s := strings.SplitAfter(string(data), ":")
-	if len(s) != 2 {
-		return "", fmt.Errorf("Problem parsing workload ID")
-	}
-
-	return strings.TrimSpace(s[1]), nil
+	return CreateWorkloadFromFile(ctx, public, tenant, "workload.yaml")
 }
 
 func deleteWorkload(ctx context.Context, tenant string, workload string, public bool) (err error) {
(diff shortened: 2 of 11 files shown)


markdryan:
This is ready for review and merge now, pending the semaphore build.

k8s/kubicle/create.go
```diff
@@ -0,0 +1,644 @@
...
+	err = ioutil.WriteFile(c.masterPath, buf.Bytes(), 0600)
+	if err != nil {
+		return err
+	}
+
+	buf.Reset()
+
+	err = masterUDTmpl.Execute(&buf, c.mc)
+	if err != nil {
+		_ = os.Remove(c.masterPath)
```
rbradford: Make this conditional on !keep?

k8s/kubicle/create.go
```diff
@@ -0,0 +1,644 @@
...
+		return err
+	}
+
+	c.masterUDPath = filepath.Join(c.cwd, c.mc.UserDataFile)
+
+	err = ioutil.WriteFile(c.masterUDPath, buf.Bytes(), 0600)
+	if err != nil {
+		_ = os.Remove(c.masterPath)
```
rbradford: ditto

k8s/kubicle/create.go
```diff
@@ -0,0 +1,644 @@
...
+	err = ioutil.WriteFile(c.workerPath, buf.Bytes(), 0600)
+	if err != nil {
+		return err
+	}
+
+	buf.Reset()
+
+	err = workerUDTmpl.Execute(&buf, c.wc)
+	if err != nil {
+		_ = os.Remove(c.workerPath)
```
rbradford: ditto

k8s/kubicle/create.go
```diff
@@ -0,0 +1,644 @@
...
+		return err
+	}
+
+	c.workerUDPath = filepath.Join(c.cwd, c.wc.UserDataFile)
+
+	err = ioutil.WriteFile(c.workerUDPath, buf.Bytes(), 0600)
+	if err != nil {
+		_ = os.Remove(c.workerPath)
```
rbradford: ditto

k8s/kubicle/kubicle.go
```diff
@@ -161,13 +163,42 @@ func runCommand(signalCh <-chan os.Signal) error {
 	return err
 }
 
+// TODO: copied from launcher
+func getLock() error {
+	err := os.MkdirAll(lockDir, 0777)
```
rbradford: 0777?

rbradford: 0777 is too broad.

k8s/kubicle/kubicle.go
```diff
@@ -0,0 +1,213 @@
...
+	case err = <-errCh:
+		cancelFunc()
+	}
+
+	return err
+}
+
+// TODO: copied from launcher
+func getLock() error {
+	err := os.MkdirAll(lockDir, 0777)
```
rbradford: 0777 is too broad.

k8s/kubicle/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+defaults:
+    vcpus: {{.VCPUs}}
+    mem_mb: {{.RAM}}
+cloud_init: "{{.UserDataFile}}"
+disks:
+  - source:
+       service: image
+       id: "{{.ImageUUID}}"
+    size: {{.Disk}}
+    ephemeral: true
```
markdryan: I can add a cli option for this.

testutil/ciao-down/cloudinit.go
```diff
@@ -196,6 +196,9 @@ runcmd:
  - {{download . "https://download.fedoraproject.org/pub/fedora/linux/releases/24/CloudImages/x86_64/images/Fedora-Cloud-Base-24-1.2.x86_64.qcow2" (printf "/home/%%s/local/Fedora-Cloud-Base-24-1.2.x86_64.qcow2" .User)}}
  - {{template "CHECK" .}}
 
+ - curl -X PUT -d "Downloading xenial-server-cloudimg-amd64-disk1.img" 10.0.2.2:{{.HTTPServerPort}}
+ - {{download . "https://cloud-images.ubuntu.com/xenial/current/xenial-server-cloudimg-amd64-disk1.img" (printf "/home/%%s/local/xenial-server-cloudimg-amd64-disk1.img" .User)}}
+
```
markdryan: We could drop the clear image but this is quick to upload so there's not much point.  Dropping Fedora would require changing some of the BAT tests but it would speed up singlevm startup by about 15 seconds.  @kaccardi what do you think?  Shall we ditch fedora?

bat/workload.go
```diff
@@ -172,6 +152,35 @@ func CreatePublicWorkload(ctx context.Context, tenant string, opt WorkloadOption
...
+	var data []byte
+	var err error
+	if public {
+		data, err = RunCIAOCLIAsAdmin(ctx, tenant, args)
+	} else {
+		data, err = RunCIAOCLI(ctx, tenant, args)
+	}
+
+	if err != nil {
+		return "", err
```
markdryan: The error message could be better but it shouldn't be fixed in this PR.  This is not new code, it is existing code that has just be refactored into a function.  Also, the bat package doesn't use the errors package yet.

k8s/kubicle/cloudinit.go
```diff
@@ -0,0 +1,106 @@
...
+ - echo "deb http://apt.kubernetes.io/ kubernetes-xenial main" >/etc/apt/sources.list.d/kubernetes.list
+ - apt-get update
+ - DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true apt-get install -y docker-engine
+ - DEBIAN_FRONTEND=noninteractive DEBCONF_NONINTERACTIVE_SEEN=true apt-get install -y kubelet kubeadm kubectl kubernetes-cni
+ - {{template "PROXIES" .}}no_proxy=` + "`hostname -i`" + ` kubeadm init --pod-network-cidr 10.244.0.0/16 --token {{.Token}} {{if len .ExternalIP}}--apiserver-cert-extra-sans={{.ExternalIP}}{{end}}
+ - cp /etc/kubernetes/admin.conf /home/{{.User}}/
+ - chown {{.User}}:{{.User}} /home/{{.User}}/admin.conf
+ - {{template "PROXIES" .}}wget https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel-rbac.yml
+ - KUBECONFIG=/home/{{.User}}/admin.conf kubectl create -f kube-flannel-rbac.yml
+ - {{template "PROXIES" .}}wget https://raw.githubusercontent.com/coreos/flannel/master/Documentation/kube-flannel.yml
```
markdryan: We did, but I'll do this in a follow on PR.  kubicle is still functional without this change.

k8s/kubicle/kubicle.go
```diff
@@ -0,0 +1,182 @@
...
+type usageError string
+
+func (e usageError) Error() string {
+	return string(e)
+}
+
+type vmOptions struct {
+	vCPUs int
+	mem   int
+	disk  int
```
markdryan: Will do

k8s/kubicle/kubicle.go
```diff
@@ -0,0 +1,182 @@
...
+	workers       int
+	imageUUID     string
+	externalIP    string
+	keep          bool
+}
+
+type baseConfig struct {
+	VCPUs        int
+	RAM          int
+	Disk         int
```
markdryan: Okay

k8s/kubicle/kubicle.go
```diff
@@ -0,0 +1,182 @@
...
+		opts.publicKeyPath = path.Join(home, "local", "testkey.pub")
+	}
+
+	fs := flag.NewFlagSet("create", flag.ExitOnError)
+
+	fs.IntVar(&opts.masterVM.mem, "mmem", opts.masterVM.mem,
+		"Megabytes of RAM allocated to master VM")
+	fs.IntVar(&opts.masterVM.vCPUs, "mcpus", opts.masterVM.vCPUs, "VCPUs assignged to master VM")
+	fs.IntVar(&opts.masterVM.disk, "mdisk", opts.masterVM.disk,
+		"Gigabytes of disk allocated to master VM")
```
markdryan: Okay, I'll update the help.

k8s/kubicle/create.go
markdryan: The goal is to create these two workload files atomically as c.masterPath references c.mc.UserDataFile.  It is useless without it so we delete it even if keep is specified.

k8s/kubicle/create.go
markdryan: This is intentional

k8s/kubicle/create.go
markdryan: This is intentional

k8s/kubicle/create.go
markdryan: Again this is done on purpose to ensure creation of workload files are atomic.

k8s/kubicle/kubicle.go
markdryan: Agreed.  Also I'm thinking of getting rid of this altogether.  It will prevent a race if you try to create two k8s clusters in the same tenant at the same time from the same machine.  But it can't help if two people run kubicle create on the same tenant from two different machines at the same time.  What do you think?  Is it worth keeping?

k8s/kubicle/cloudinit.go
markdryan: What would the use case be here for wanting to preserve these volumes?  They shouldn't contain any user data, just binaries for running k8s.  They can be recreated by re-running kubicle.

testutil/ciao-down/cloudinit.go
kaccardi: @markdryan there were some tests that didn't pass with the clear image, so we can definitely drop that - even though it doesn't take much upload time what's the point in having it if we don't use it? If we can get by with just the ubuntu cloudimg that is fine with me.

bat/external_ip.go
```diff
@@ -0,0 +1,94 @@
...
+
+// ListExternalIPs returns detailed information about all the external ips
+// defined for the given tenant.  The information is retrieved using the
+// ciao-cli external-ip list command.  An error will be returned if the
+// following environment are not set;  CIAO_IDENTITY,  CIAO_CONTROLLER,
+// CIAO_ADMIN_USERNAME, CIAO_ADMIN_PASSWORD.
+func ListExternalIPs(ctx context.Context, tenant string) ([]*ExternalIP, error) {
+	var externalIPs []*ExternalIP
+	args := []string{"external-ip", "list", "-f", "{{tojson .}}"}
+	err := RunCIAOCLIAsAdminJS(ctx, tenant, args, &externalIPs)
```
kaccardi: when calling the function as admin, you'll get a list of every mapped IP for all tenants. Is that what was intended?

testutil/ciao-down/cloudinit.go
markdryan: I guess the one good thing about the clear image is that it's an uefi image so some different code paths are tested by booting it.  But we already test these with the cnci, so as long as we have a clear based CNCI image we're covered.  So I'll remove the clear and fedora images for the time being.

bat/external_ip.go
markdryan: Yes as I also need the instance ids, which are not returned when calling the function as a normal user, e.g.,

```
$ ciao-cli external-ip list -f '{{tojson .}}'
[
	{
		"mapping_id": "52bea4ed-0886-48ae-8d75-15d871ed6b83",
		"external_ip": "198.51.100.2",
		"internal_ip": "172.16.0.2",
		"instance_id": "",
		"tenant_id": "",
		"pool_id": "",
		"pool_name": "",
		"links": [
			{
				"rel": "self",
				"href": "https://singlevm:8889/dd5c1eea2c434e9d89d88806d00394e5/external-ips/52bea4ed-0886-48ae-8d75-15d871ed6b83"
			}
		]
	}
```

Now perhaps this is a bug in ciao-cli, ciao-controller, which could be fixed, but for the time being I need to invoke the function as admin.

testutil/ciao-down/cloudinit.go
markdryan: I've updated the PR so Fedora and Clear are no longer downloaded by default.  I didn't remove them completely though.  There still available via setup.sh's -a option.

PR #1293, opened 2017-05-26
testutil: Update test certificates used for unit tests

markdryan:
The SSNTP unit tests rely on some hardcoded test certificates which had
expired.  This commit replaces the old certificates with some new certs
which will expire sometime in 2037.


markdryan:
I'm just going to merge this as our build are currently broken.

@tpepper If you rebase your patch should pass travis.

tpepper:
Aha!  Duh yes you're right!  I was looking at the dynamically generated certs and seeing they were valid and good relative to the clock.

PR #1296, opened 2017-05-31
Gophercloud revendor

markdryan:
This patch revendors gophercloud to the latest version.  In doing so it switches us from the github.com/rackspace/gophercloud repo to the new github.com/goperhcloud/gophercloud repo.  There have been loads of API breaking changes since we last updated.  I think I've caught all the problems and the BAT tests are passing.  

However, there is one issue that I wasn't able to fix satisfactorily.  The volumes returned by the /v2/{tenant}/volumes/detail endpoint no longer contain CreatedAt timestamps.  The reason is due to a change in the way gophercloud parses timestamps returned from the volume service.  It fails to parse any timestamps that contain timezone information.  We can hack around the problem for ciao-cli volume show but not for ciao-cli volume list.  As far as I can tell from the openstack documentation this field shouldn't be returned for this endpoint anyway.  The openstack cli tool certainly ignores it.  One consequence of all this is that it is no longer possible to sort the results of ciao-cli volume list by creation time.  This commit sorts them instead by name.


markdryan:
The other alternatives I considered to fixing the volume timestamps were:

1. Changing the format of the timestamps produced by the volume service.  This would fix gophercloud and openstack clients but might break other clients.

2. Not use gophercloud to implement the volume list command.  This seemed a bit weird as all the other openstack commands in ciao-cli use gophercloud.

The obvious easy fix would be to patch gophercloud but the maintainers aren't so keen to change it's behaviour.   It's hard for us to justify a patch as the issue isn't reproducible in openstack.

markdryan:
@rbradford, @kaccardi This one needs a good test.  There may be changes not covered by the BAT tests that I've missed.

markdryan:
Let's not merge this one right now.  I'm having one last go and trying to change gophercloud.

ciao-cli/volume.go
```diff
@@ -28,10 +28,20 @@ import (
...
 )
 
+type customVolume volumes.Volume
+type customVolumeExt volumetenants.VolumeExt
+type extendedVolume struct {
+	customVolumeExt
+	customVolume
+	CreatedAt time.Time `json:"created_at"`
+	UpdatedAt time.Time `json:"updated_at"`
+}
```
rbradford: This looks right. I'm saddened how complex it makes our code though.

bat/image.go
```diff
@@ -49,7 +49,7 @@ type ImageOptions struct {
 // Image contains all the meta data for a single image
 type Image struct {
 	ImageOptions
-	SizeBytes   int
+	SizeBytes   int `json:"size"`
```
rbradford: Did you consider renaming to Size? This structure is here to encapsulate the output from the CLI.

PR #1299, opened 2017-06-02
ciao-controller: Avoid raciness when adding tenant for workload creation

rbradford:
This change modifies datastore's AddTenant() to return a newly defined
error if the tenant ID is already in use. This allows the code in
controller to avoid a race between using GetTenant to find out if there
is a tenant AddTenant to add a new one.

This change also ensures that the lock is taken early enough and that
the lock is released on error paths (previously the lock was not
releases on error paths.)


rbradford:
ds.getTenant() to ds.db.getTenant() is needed as ds.getTenant() takes the lock (again, causing a deadlock) and then checks the same map we've already checked before calling out to the database.

rbradford:
I've got a better PR.

PR #1300, opened 2017-06-02
openstack: Fix volume marshaling

markdryan:
The marshalling code for volumes generated json that was unparseable by the
openstack client.  It marshaled fields that were not defined and some
mandatory fields were missing.  This commit addresses these issues.  It's
now possible to do:

openstack volume show

to view a ciao volume.


openstack/block/api.go
```diff
@@ -559,6 +559,9 @@ func showVolumeDetails(bc *Context, w http.ResponseWriter, r *http.Request) (API
 	if err != nil {
 		return errorResponse(err), err
 	}
+	vol.MetaData = map[string]string{"contents": "not junk"}
+	s := "None"
+	vol.VolumeType = &s
```
tpepper: What's the purpose of "not junk" here?

markdryan: This was copied from an openstack example but it turns out it's not needed.  An empty map will suffice. I'll remove it.

PR #1301, opened 2017-06-04
Fix tenant CNCI/workload raciness

rbradford:
This PR addresses a bug found when creating non-public workloads before the CNCI has been activated. This is caused by:

- CreateWorkload() creates a tenant without starting a CNCI
- confirmTenant() checked if the tenant was there or didn't have an IP address. In either case it would call addTenant() which would try and call AddTenant() on the database which after 06085307c8e7b3670c88d712177daca0f6f021b4 will return an error on the duplicate tenant being added.

The solution in this PR is to avoid calling AddTenant() twice.

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -106,29 +106,25 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 	}
 
 	if tenant == nil {
-		if *noNetwork {
-			_, err := c.ds.AddTenant(tenantID)
-			if err != nil {
-				return err
-			}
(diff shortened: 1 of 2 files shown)


rbradford:
@markdryan I've tested this combined with your kubicle PR and I was able to create the required workloads.

rbradford:
@kaccardi Am I correct in my assertion that the special behaviour is no longer required? I didn't see the CNCI going through this code path any longer?

markdryan:
The updated PR fixes the issue for me.

PR #1306, opened 2017-06-07
Delete CNCI when it is no longer in use.

kaccardi:
If we do not need the CNCI because the tenant has deleted all of their instances, delete the CNCI as well in order to keep it from hanging around forever.

Fixes: #1307

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -92,12 +92,7 @@ func (client *ssntpClient) deleteEphemeralStorage(instanceID string) {
 	}
 }
 
-func (client *ssntpClient) releaseResources(instanceID string) error {
-	i, err := client.ctl.ds.GetInstance(instanceID)
-	if err != nil {
-		return errors.Wrapf(err, "error getting instance from datastore")
-	}
-
+func (client *ssntpClient) releaseResources(i *types.Instance) error {
 	// CNCI resources are not quota tracked
 	if i.CNCI {
 		return nil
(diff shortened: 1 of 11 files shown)


markdryan:
The semaphore failure actually looks real.

```
base_bat_test.go:88: Starting an instance should increase instance count on node &{9ced9217-c20c-422e-bdae-afcacabc238b 0001-01-01 00:00:00 +0000 UTC READY 7983 5801 60921 52549 1 4 1 1 0 0} &{9ced9217-c20c-422e-bdae-afcacabc238b 0001-01-01 00:00:00 +0000 UTC READY 7983 6524 60921 52721 1 4 1 1 0 0}
```

ciao-controller/client.go
```diff
@@ -114,26 +109,84 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
...
 	if err != nil {
 		glog.Warningf("Error deleting instance from datastore: %v", err)
 	}
+
+	if i.CNCI {
+		return
+	}
+
+	// check the cnci after 5 minutes has elapsed to see if it's idle.
+	go func() {
```
markdryan: If I do a ciao-cli instance delete -all on a 10000 node cluster will I end up with 10000 of these go routines?  This might not be a critical problem if checkCNCI is re-entrant but you'll probably end up with loads of warnings in the logs.

markdryan: What is the benefit of waiting 5 seconds?  It comes at a price, i.e., making to code more complex.

ciao-controller/client.go
```diff
@@ -114,26 +109,84 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
...
+
+	cncis, err := client.ctl.ds.GetTenantCNCISummary(t.CNCIID)
+	if err != nil {
+		return errors.Wrapf(err, "error getting cnci from datastore")
+	}
+
+	// tenants only have one cnci right now
+	cnci := cncis[0]
+
+	if len(cnci.Subnets) > 0 {
```
markdryan: Is this how we determine whether there are any instances?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -681,7 +681,18 @@ func (ds *Datastore) ReleaseTenantIP(tenantID string, ip string) error {
 	ds.tenantsLock.Lock()
 
 	if ds.tenants[tenantID] != nil {
-		ds.tenants[tenantID].network[int(subnetInt)][int(ipBytes[3])] = false
+		delete(ds.tenants[tenantID].network[int(subnetInt)], int(ipBytes[3]))
+		subnets := ds.tenants[tenantID].subnets
+		network := ds.tenants[tenantID].network
+		i := int(subnetInt)
```
rbradford: Why not use your intermediate variables here:

delete(network[i], int(ipBytes[3]))

ciao-controller/client.go
markdryan: Is this a little bit racy?  Would it be possible for another tenant to call confirmTenantCNCI after this function has confirmed that cnci.Subnets == 0 but before it calls RemoveTenantCNCI?  In this case we'd be removing a CNCI for an instance that needs it.

ciao-controller/client.go
markdryan: Okay, @rbradford has explained this to me.  Without the 5 second wait our BAT tests are going to be a lot slower.

rbradford: Could we only start this goroutine only once the subnet count falls to zero?

ciao-controller/client.go
```diff
@@ -114,26 +109,84 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
...
+		return nil
+	}
+
+	// cache this value.
+	cnciID := t.CNCIID
+
+	// clean up tenant info in datastore.
+	err = client.ctl.ds.RemoveTenantCNCI(t.ID)
```
rbradford: What about moving then len(cnci.Subnets) > 0 check into here and refuse to remove an in-use CNCI. If you do that within the lock used for adding CNCIs then then you should avoid the race Mark is worried about.

ciao-controller/client.go
```diff
@@ -114,26 +109,84 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
...
+	if err != nil {
+		// probably this is ok, it means that something needs
```
rbradford: Let's have a specific error for that case, just silently ignoring errors has been a source of bugs in the past.

ciao-controller/command.go
```diff
@@ -106,12 +106,25 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
...
+		_, err = c.ds.AddTenant(tenantID)
 		if err != nil {
 			return err
 		}
 	}
 
+	return nil
+}
+
+func (c *controller) confirmCNCI(tenantID string) error {
```
rbradford: With this change confirmTenant() will no longer start the CNCI. But this means:

1. The CNCI is only started when starting an instance (see startWorkload() below)
2. All the safety checks done in confirmTenant() to make sure that CNCI starting is not racey are bypassed

ciao-controller/client.go
markdryan: We did discuss using a go-routine to manage this stuff offline.  I was a bit hesitant about doing this when we spoke but now I better understand the problem, having seen the code.  One possible solution would be to have a  go routine per subnet that manages the CNCI lifecycle.   This would make it easy to

1. Protect against starting two CNCIs at one ( currently the job of confirmTenant ) 
2. Protect against removing a CNCI that has just been added.
3. Implement a timer for removing the CNCI when the last instance for that subnet has gone away.
4. Protect against deleting a CNCI that is starting up.  Maybe this isn't a problem.

We need to be careful to handle the controller restart case, i.e., it would need to start the go-routines for all the cncis on startup.

rbradford: > We need to be careful to handle the controller restart case, i.e., it would need to start the go-routines for all the cncis on startup.

And whatever state they might be in (e.g. soon to be deleted.)

kaccardi: I'll prepare an alternate implementation similar to this for comparison.

ciao-controller/command.go
kaccardi: starting the CNCI only when starting instances was a design choice. How do people feel about this? I personally don't believe the optimization of starting the CNCI whenever there's any activity at all is worth the complexity it adds.

rbradford: I endorse starting the cnci only on instance add. 

My concern about the raciness (that the code in confirmTenant()) tried to solve. I think if you adjust the way you split confirmCNCI() out from confirmTenantRaw() then you'd get that (renaming confirmTenant() to confirmCNCI() and then creating a confirmTenant() that only does the adding of the tenant.)

PR #1308, opened 2017-06-09
storage_bat: Re-enable storage tests

markdryan:
By disabling the Fedora workload by default, I inadvertently disabled
some of the storage BAT tests.  The tests were skipped as the fedora
workload could not be found.  This commit updates the storage tests
to use the Ubuntu workload which is always present on SingleVM builds.


markdryan:
The semaphore build is failing on what looks like real test cases failure.  I've also seen this one locally.  I'll need to take a look at this.

PR #1311, opened 2017-06-13
configuration: Move default CNCI configuration values

rbradford:
Move the default values for CNCI resources from controller to the
configuration defaults. This allows code that uses InitDefaults() to get
sane defaults for these values.


rbradford:
@kaccardi  I think your review is required on this change.

PR #1312, opened 2017-06-14
Allow ciao-down VMs mounts and mapped ports to be controlled

markdryan:
The commit makes three main changes to ciao-down.

- The settings, e.g., memory, cpu usage and mounts, specified when VMs are created using prepare or started using start, are now persisted.  These settings are re-used when the VM is started at a later date.
- Users can specify which directories they would like to see mounted in the guest VM and the 9p security model used to mount them.  Multiple mounts can be specified.
- Users can specify which ports can be mapped from the host's localhost to the guest VM.

The two vm types currently supported by ciao include one mounted directory by default, the GOPATH.  It is not possible to remove this mount entirely via the command line, although this can be done by editing the ~/.ciao-down/instance/state.yaml.  However, it is possible to change the security model or the host directory to mount.  This can done as follows

--mount hostgo,none,$HOME/go

The key is to specify the existing hostgo tag.  When ciao-down processes a new mount it checks to see whether the tag, hostgo in this case, already exists.  If it does it updates the mount with the new information.

A similar situation exists with ports.  Each ciao-down VM has one port mapped, port 22, which is mapped to 10022 on the hosts localhost interface.  It is possible to change the host port by specifying a --port option, e.g.,

--port 10023-22

As the guest port 22 has already been defined a new port mapping is not created.  Instead the host port of the existing mapping is updated from 10022 to 10023.

### DIFF
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -26,12 +26,12 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/signal"
 	"path"
-	"path/filepath"
+	"strconv"
+	"strings"
 	"syscall"
 	"time"
 )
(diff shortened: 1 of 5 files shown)


markdryan:
@grahamwhaley here's the PR that should allow you to change the 9p security model used to mount the GOPATH in the guest VM.  Let me know if you have any problems.  You should just be able to do,

./ciao-down stop
pull and build new version of ciao-down from this PR
./ciao-down start --mount hostgo,none,$HOME/go

testutil/ciao-down/ciao_down.go
```diff
@@ -64,10 +64,57 @@ func init() {
...
+}
+
+type ports []portMapping
+
+func (p *ports) String() string {
+	return fmt.Sprint(*p)
+}
+
+func (p *ports) Set(value string) error {
+	components := strings.Split(value, "-")
```
rbradford: My only feedback is that I think <host>:<guest> is a more common notation to refer to port mappings rather than a hyphen.

But that's not enough to hold this PR back :-)

grahamwhaley:
Hi. I can ack that the security model for the 9p can be set with --mount and looks to be persisting.
But... the memory size set on 'start' does not look like it is persisting for me.
I did a `start -mem 6`, got a a 6 on the command line and checked it was in the state.yaml.
Then I did a `stop` and then a `start` with no parameters, and got a default `3G` vm - confirmed in the state.yaml and on the qemu command line through a `ps`.

markdryan:
@grahamwhaley Thanks for the review.  I think I see the problem with mem and cpu.  I'll update the PR shortly with a fix.

markdryan:
@grahamwhaley This should be fixed now.  Could you try again?

grahamwhaley:
Thanks. I can confirm the 9p security model and the memory size are persisting here.
lgtm

PR #1316, opened 2017-06-16
singlevm: Improve singlevm performance by starting keystone earlier

rbradford:
This PR improves the performance of singlevm's ./setup.sh (on my system it goes from 2m06s to 1m45s) by starting the keystone container earlier which means less time is spent blocking waiting for keystone to start and be populated before controller starts.

This PR also makes some other minor cleanups around using absolute path names and removing excessive "cd"s to the same directory.


rbradford:
@markdryan Some more singlevm tweaks which give a good performance improvement for me!

PR #1317, opened 2017-06-16
New deployment tools integrated with singlevm

rbradford:
This PR contains an implmentation of the new deployment tooling integrated into singlevm as a demonstration of how it works.

The commit message in "deploy: Add deployment client and server" explains in detail how the client/server work and are used.


markdryan:
I gave this a try.  The first time it failed with the same error that I see in the semaphore builds.  I wasn't able to reproduce this subsequently.  When it worked the cluster seemed to be functional.  Will test and review some more next week.

rbradford:
@markdryan @kaccardi as discussed yesterday i've removed the WIP label on this to get some more detailed review.

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+
+	downloadedClientPath, err := deploy.DownloadTool(deployServerHost, "ciao-deploy-client")
+	if err != nil {
+		errors.Wrap(err, "Error when downloading client")
+	}
+	defer os.Remove(downloadedClientPath)
+
+	fmt.Println("Downloaded client")
+
+	if _, err := os.Stat(downloadedClientPath); !os.IsNotExist(err) {
```
markdryan: I don't really get what this if statement is checking.  Presumably, the file exists if DownloadTool returns no error, although I haven't looked at the code of this function yet.   Also, I don't really understand what the conditions for entering this block our.  Currently, we come in here, if stat succeeded or if stat failed with some reason other than does not exist.  Is this what we want?

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+}
+
+// Provides path to CA certificate. If not locally available then this is downloaded from the server.
+func obtainCACert() (string, error) {
+	systemCaCertPath := path.Join(systemPkiLocation, "CAcert.pem")
+
+	// TODO: Check if certificate installed is valid
+	if !force {
+		if _, err := os.Stat(systemCaCertPath); !os.IsNotExist(err) {
+			fmt.Printf("CA certificate already installed. Skipping download.\n")
```
markdryan: This doesn't mean that the cert already exists.  It means that the cert already exists or an error is returned by stat which is not IsNotExist.

markdryan: I'm not sure what should be done in such circumstances, but just wanted to note this.

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+
+	f, err := ioutil.TempFile("", "CAcert.pem")
+	if err != nil {
+		return "", errors.Wrap(err, "Error creating CA certificate file")
+	}
+	defer os.Remove(f.Name())
+
+	_, err = io.Copy(f, &certOutput)
+	if err != nil {
+		return "", errors.Wrap(err, "Error copying CA certificate to file")
```
markdryan: We leak f here.

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	if err := f.Chmod(0644); err != nil {
+		return "", errors.Wrap(err, "Error changing mode on certificate")
```
markdryan: We leak f here.

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	r.Header.Set("Content-Type", "text/plain")
+
+	client := &http.Client{}
+	res, err := client.Do(r)
+	if err != nil {
+		return "", errors.Wrap(err, "Error requesting signing of CSR")
+	}
+	defer res.Body.Close()
```
markdryan: No need for defer here.

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+
+	r, err := http.NewRequest("PUT", url, &csrOutput)
+	if err != nil {
+		return "", errors.Wrap(err, "Error requesting signing of CSR")
+	}
+
+	r.Header.Set("Content-Type", "text/plain")
+
+	client := &http.Client{}
+	res, err := client.Do(r)
```
markdryan: As a general comment there are a few long running networky tasks, and invocations of other tasks, launched from this tool, with no way to cancel them.  If one of these hangs will this prevent me from being able to cleanly quit ./setup.sh?

deploy/ciao-deploy-client/main.go
```diff
@@ -0,0 +1,474 @@
...
+
+	f, err := ioutil.TempFile("", deploy.CertName(ssntpRole))
+	if err != nil {
+		return errors.Wrap(err, "Error creating certificate file")
+	}
+	defer os.Remove(f.Name())
+
+	_, err = io.Copy(f, &certOutput)
+	if err != nil {
+		return errors.Wrap(err, "Error copying certificate to file")
```
markdryan: f is leaked here.

rbradford:
Thanks for the first review @markdryan. I've integrated all the actionable feedback you gave!

deploy/utils.go
```diff
@@ -0,0 +1,168 @@
...
+Restart=no
+KillMode=process
+
+[Install]
+WantedBy=multi-user.target
+`
+
+// DownloadTool downloads the desired tool from the server
+func DownloadTool(deployServerHost, tool string) (string, error) {
+	url := fmt.Sprintf("%s/download/%s", deployServerHost, tool)
```
markdryan: Better to use filepath.Join here.

deploy/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	fmt.Printf("Downloading %s: %s\n", tool, url)
+	r, err := http.Get(url)
+
```
markdryan: Any reason for the newline?

deploy/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	return f.Name(), nil
+}
+
+// InstallTool installs a tool to its final destination and manages it via systemd
+func InstallTool(toolPath, tool, caCertPath, certPath string) error {
+	fmt.Printf("Installing %s\n", tool)
+
+	fmt.Printf("Stopping %s\n", tool)
+	cmd := exec.Command("sudo", "systemctl", "stop", tool)
+	cmd.Run()
```
markdryan: We're ignoring an error here.

deploy/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	serviceFilePath := path.Join("/etc/systemd/system", fmt.Sprintf("%s.service", tool))
+
+	f, err := ioutil.TempFile("", fmt.Sprintf("%s.service", tool))
+	if err != nil {
+		return errors.Wrap(err, "Error creating temporary file for service unit")
+	}
+	defer os.Remove(f.Name())
+
+	if _, err := f.Write([]byte(systemdData)); err != nil {
+		return errors.Wrap(err, "Error writing service file data")
```
markdryan: We leak f here.

rbradford: Done.

deploy/utils.go
rbradford: Done.

deploy/utils.go
rbradford: Fixed.

deploy/utils.go
rbradford: It's not a path but a URI starting with http://

deploy/utils.go
```diff
@@ -0,0 +1,168 @@
...
+	"os/exec"
+	"path"
+
+	"github.com/01org/ciao/ssntp"
+	"github.com/pkg/errors"
+)
+
+// CertName gives file name to use for downloaded certificate
+func CertName(role ssntp.Role) string {
+	hs, err := os.Hostname()
```
rbradford: Yeh, done.

deploy/ciao-deploy-server/main.go
```diff
@@ -0,0 +1,476 @@
...
+
+	var err error
+	anchorCertPath, caCertPath, err = createSchedulerCert()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error creating scheduler certificate: %s\n", err)
+		os.Exit(1)
+	}
+	if scheduler {
+		installScheduler(caCertPath, anchorCertPath)
+		if err != nil {
```
rbradford: Yes, you could run the deployment server on a different machine.

deploy/ciao-deploy-server/main.go
```diff
@@ -0,0 +1,476 @@
...
+	flag.Parse()
+
+	var err error
+	anchorCertPath, caCertPath, err = createSchedulerCert()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error creating scheduler certificate: %s\n", err)
+		os.Exit(1)
+	}
+	if scheduler {
+		installScheduler(caCertPath, anchorCertPath)
```
rbradford: I'm hoping that that this part of the tool is idempotent for multiple runs, e.g. by generating files into a temporary location and then copying over.

In particular the client is designed to be run in two phases, the first to generate the CSR and upload. The second to download the signed certficate.

Cleaning up needs to be done carefully, if you're running the server only to offer up upgrades and then you delete the scheduler certs because of some failure to e.g. bind to the interface then the whole cluster needs to be recreated.

In this particular case I don't think failing to install scheduler should delete the certificates as they may have been already there before this invocation.

deploy/ciao-deploy-server/main.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	f, err := ioutil.TempFile("", "configuration.yaml")
+	if err != nil {
+		return "", errors.Wrap(err, "Error creating temporary file")
+	}
+
+	_, err = f.Write(data)
+	if err != nil {
+		return "", errors.Wrap(err, "Error writing data to temporary file")
```
rbradford: Done.

deploy/ciao-deploy-server/main.go
```diff
@@ -0,0 +1,476 @@
...
+	requests = make(map[string]*signingRequest)
+}
+
+func sign(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "PUT" {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+
+	vars := mux.Vars(r)
+	if vars["token"] != secretToken {
```
rbradford: So with this set of tools the token is pre-shared. But the next level of tooling will generate that token so reusing that code will be beneficial.

rbradford:
@markdryan can you review the new http server / signal handling bits?

PR #1320, opened 2017-06-27
sqlite3db: Remove old database population code

rbradford:
This PR removes some residual database code that is no longer used after the removal of default workload population from CSV files.


rbradford:
@markdryan can you take a look? As I mentioned offline this is also allows us to restart controller now (previously the the init would fail due to clashes on database table population)

PR #1325, opened 2017-07-05
ciao-controller: add fatal logging for controller HTTP services

rbradford:
Fatally log if controller HTTP services return an error. These are run
inside a goroutine as ListenAndServeTLS blocks so it is necessary to
report these errors inside an anonymous function run via a goroutine.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -204,13 +204,25 @@ func main() {
 	}
 
 	wg.Add(1)
-	go ctl.startComputeService()
+	go func() {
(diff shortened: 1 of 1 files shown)


rbradford:
@markdryan: as we discussed yesterday.

CC: @kaccardi 

The use of the sync.WaitGroup / shutdown problems are discussed in #1326

rbradford:
@markdryan Acknowledged. I hope we can resolve it by the approach outlined in #1326

Issue #1324, opened 2017-07-04
Controller creates a CNCI for tenant "public"


rbradford:
The "public" tenant label is only intended as a means to mark workloads as available to all tenants. However a CNCI is being started for this and as such it is being treated as a real tenant.

PR #1327, opened 2017-07-05
ciao-controller: Don't start a CNCI for "public" tenant

rbradford:
The tenant with ID "public" exists in the database to satisfy the
requirement that workloads are associated with a tenant. This
pseudo tenant is thus used to signify that this workload is public.

However as no instances are going to be created with this tenant it is
not necessary to launch a CNCI for the tenant.

Fixes: #1324

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -112,6 +112,12 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


rbradford:
Long term i'd like to see if we can remove this pseudo tenant and replace it with a field in the database to indicate that this is a public workload.

rbradford:
verify.sh & BAT tests pass. Verified that no CNCI running after creating public workloads.

Issue #1303, opened 2017-06-06
ciao-cli external-ip list returns limited information for non admin users


markdryan:
For example,

```
$ ciao-cli external-ip list -f '{{tojson .}}'
[
	{
		"mapping_id": "52bea4ed-0886-48ae-8d75-15d871ed6b83",
		"external_ip": "198.51.100.2",
		"internal_ip": "172.16.0.2",
		"instance_id": "",
		"tenant_id": "",
		"pool_id": "",
		"pool_name": "",
		"links": [
			{
				"rel": "self",
				"href": "https://singlevm:8889/dd5c1eea2c434e9d89d88806d00394e5/external-ips/52bea4ed-0886-48ae-8d75-15d871ed6b83"
			}
		]
	}
```

There's no instance_id, tenant_id, pool_id or pool_name reported.  It's not clear to me whether this is intentional or there is a good reason to exclude this information for non-admin users.

kaccardi:
the return struct for a non-admin user is MappedIPShort:
// MappedIPShort is a summary version of a MappedIP.
type MappedIPShort struct {
        ID         string `json:"mapping_id"`
        ExternalIP string `json:"external_ip"`
        InternalIP string `json:"internal_ip"`
        Links      []Link `json:"links"`
}

It's limited in info mainly to restrict the info to what the user actually needs to see and nothing more. For example, the instance ID is already tied to the user's internal IP, so adding that doesn't add any information. The tenant ID is your own tenant ID, so you don't need that. the pool id and name aren't relevant to the user since they can't query the pool for info (this is restricted to admin).

markdryan:
I see.

However, from a BAT point of view it would be useful to have instanceID.  This would help us to map external-ips to instance structures.  When we call bat.GetAllInstances we get a map indexed by instance-ids.

Also, having the internal-ip is not quite the same as having the instance-id as the internal-ips are re-used.   For example, let's suppose a user retrieves a list of external-ips and then a list of instances and tries to match them up.   By the time he's retrieved the list of instances, the external-ip could have been unmapped, the original instance destroyed and a new one created which re-uses the old internal-ip.  In such a scenario the caller would think that the external-ip is mapped to an instance which has never had any external-ip associated with it.

rbradford:
We should modify MappedIPShort to add the InstanceID to avoid the raciness @markdryan highlights.

PR #1329, opened 2017-07-05
Expose instance ID to all users of external-ip list

rbradford:
Previously this was only exposed to admin users but there is no harm in exposing it to all users (and potential benefits)

Fixes: #1303


rbradford:
CC @kaccardi

Issue #1289, opened 2017-05-19
Showing a pool as normal user gives confusing results


rbradford:
```
rob@singlevm:~/go/src/github.com/01org/ciao/testutil/singlevm$ ciao-cli pool show -name k8s-pool
	UUID: 
	Name: 
	Free IPs: 0
	Total IPs: 0

```

Should give a 403.

PR #1331, opened 2017-07-05
Minor pool bug fixes

rbradford:
Whilst I was looking into #1289 I found a few other minor issues which are fixed in this PR along with a fix for #1289.


rbradford:
Only slightly controversial bit is the removal of dumpPool() i'm going to argue that it's not useful for debugging in its current form as it's only used on show/list where you already get the pool details.

PR #1333, opened 2017-07-05
[DONOTMERGE] Support a CNCI per subnet per tenant in ciao-controller

kaccardi:
This PR implements multiple CNCIs per tenant in the controller, one for each subnet. When a subnet is no longer in use, the CNCI for that subnet will be deleted after a 5 minute settle period. It creates a CNCI controller interface per tenant. This controller is responsible for managing the CNCI lifecycle.  With this code, we will now store CNCI instance information in the datastore with the tenant instances so that we can delete CNCI instances when we no longer need them.

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -158,9 +158,23 @@ func (client *ssntpClient) concentratorInstanceAdded(payload []byte) {
 		return
 	}
 	newCNCI := event.CNCIAdded
-	err = client.ctl.ds.AddCNCIIP(newCNCI.ConcentratorMAC, newCNCI.ConcentratorIP)
+	i, err := client.ctl.ds.GetInstance(newCNCI.InstanceUUID)
 	if err != nil {
-		glog.Warningf("Error adding CNCI IP to datastore: %v", err)
+		glog.Warningf("Error getting instance: %v", err)
+		return
+	}
+
+	i.IPAddress = newCNCI.ConcentratorIP
+	i.MACAddress = newCNCI.ConcentratorMAC
+
+	err = client.ctl.ds.UpdateInstance(i)
+	if err != nil {
+		glog.Warningf("Error updating CNCI Info: %v", err)
+	}
+
+	err = client.ctl.ds.CNCIAdded(i.TenantID)
+	if err != nil {
+		glog.Warningf("Error adding CNCI: %v", err)
 	}
 }
 
@@ -511,6 +525,11 @@ func (client *ssntpClient) RestartInstance(i *types.Instance, w *types.Workload,
 		Hostname: hostname,
 	}
 
+	cnci, err := client.ctl.ds.GetInstance(t.CNCIID)
+	if err != nil {
+		return err
+	}
+
 	restartCmd := payloads.StartCmd{
 		TenantUUID:          i.TenantID,
 		InstanceUUID:        i.ID,
@@ -522,7 +541,7 @@ func (client *ssntpClient) RestartInstance(i *types.Instance, w *types.Workload,
 			VnicMAC:          i.MACAddress,
 			VnicUUID:         i.VnicUUID,
 			ConcentratorUUID: t.CNCIID,
-			ConcentratorIP:   t.CNCIIP,
+			ConcentratorIP:   cnci.IPAddress,
 			Subnet:           i.Subnet,
 			PrivateIP:        i.IPAddress,
 		},
@@ -647,6 +666,11 @@ func (client *ssntpClient) Disconnect() {
 }
 
 func (client *ssntpClient) mapExternalIP(t types.Tenant, m types.MappedIP) error {
+	i, err := client.ctl.ds.GetInstance(t.CNCIID)
+	if err != nil {
+		return err
+	}
+
 	payload := payloads.CommandAssignPublicIP{
 		AssignIP: payloads.PublicIPCommand{
 			ConcentratorUUID: t.CNCIID,
@@ -654,7 +678,7 @@ func (client *ssntpClient) mapExternalIP(t types.Tenant, m types.MappedIP) error
 			InstanceUUID:     m.InstanceID,
 			PublicIP:         m.ExternalIP,
 			PrivateIP:        m.InternalIP,
-			VnicMAC:          t.CNCIMAC,
+			VnicMAC:          i.MACAddress,
 		},
 	}
 
@@ -671,6 +695,11 @@ func (client *ssntpClient) mapExternalIP(t types.Tenant, m types.MappedIP) error
 }
 
 func (client *ssntpClient) unMapExternalIP(t types.Tenant, m types.MappedIP) error {
+	i, err := client.ctl.ds.GetInstance(t.CNCIID)
+	if err != nil {
+		return err
+	}
+
 	payload := payloads.CommandReleasePublicIP{
 		ReleaseIP: payloads.PublicIPCommand{
 			ConcentratorUUID: t.CNCIID,
@@ -678,7 +707,7 @@ func (client *ssntpClient) unMapExternalIP(t types.Tenant, m types.MappedIP) err
 			InstanceUUID:     m.InstanceID,
 			PublicIP:         m.ExternalIP,
 			PrivateIP:        m.InternalIP,
-			VnicMAC:          t.CNCIMAC,
+			VnicMAC:          i.MACAddress,
 		},
 	}
 
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -92,7 +92,12 @@ func (client *ssntpClient) deleteEphemeralStorage(instanceID string) {
 	}
 }
 
-func (client *ssntpClient) releaseResources(i *types.Instance) error {
+func (client *ssntpClient) releaseResources(instanceID string) error {
+	i, err := client.ctl.ds.GetInstance(instanceID)
+	if err != nil {
+		return errors.Wrapf(err, "error getting instance from datastore")
+	}
+
 	// CNCI resources are not quota tracked
 	if i.CNCI {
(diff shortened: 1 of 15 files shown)


ciao-controller/client.go
```diff
@@ -162,6 +179,25 @@ func (client *ssntpClient) instanceAdded(payload []byte) {
...
+	}
+
+	tenant, err := client.ctl.ds.GetTenant(i.TenantID)
+	if err != nil || tenant == nil {
+		glog.Warningf("Error getting tenant: %v", err)
+		return
+	}
+
+	if tenant.CNCIctrl == nil {
+		fmt.Println("No CNCI controller saved")
```
kaccardi: stray debug left in here.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,397 @@
...
+	cnci, ok := c.cncis[ID]
+	if !ok {
+		return false
+	}
+
+	return cnci.active()
+}
+
+func (c *CNCIManager) launch() (string, error) {
+	glog.Info("launching cnci")
```
kaccardi: remove stray debug

ciao-controller/internal/datastore/datastore.go
```diff
@@ -772,6 +795,24 @@ func (ds *Datastore) AllocateTenantIP(tenantID string) (net.IP, error) {
...
+		err := mgr.AddSubnet(int(subnetInt))
+		if err != nil {
+			glog.Warningf("Error adding subnet (%v)\n", err)
+		}
+	}
+
+	// wait for cnci to become ready before moving on.
+	if mgr != nil {
+		err := mgr.WaitForActive(int(subnetInt))
+		if err != nil {
```
rbradford: Or, after talking to @markdryan if you're just wanting to block then AddSubnet() can just return a channel, that might already be closed and <- on that. You can then use the same channel for all requests into AddSubnet()

ciao-controller/cnci.go
```diff
@@ -0,0 +1,397 @@
...
+
+	cnci.ID = ""
+
+	cnci.transitionState(failed)
+
+	return nil
+}
+
+// WaitForActive will block until a cnci for a subnet is active
+func (c *CNCIManager) WaitForActive(subnet int) error {
```
rbradford: I think this function can go away if the same channel is used to signify that the initial cnci launch vs a later "wait"

ciao-controller/client.go
```diff
@@ -158,10 +175,33 @@ func (client *ssntpClient) instanceAdded(payload []byte) {
...
+	i, err := client.ctl.ds.GetInstance(newCNCI.InstanceUUID)
 	if err != nil {
-		glog.Warningf("Error adding CNCI IP to datastore: %v", err)
+		glog.Warningf("Error getting instance: %v", err)
+		return
+	}
+
+	// update the instance mac, ip, and subnet (TBD)
+	i.IPAddress = newCNCI.ConcentratorIP
+	i.MACAddress = newCNCI.ConcentratorMAC
```
mcastelino: @kaccardi will there be another PR for CNCI to subnet tracking

ciao-controller/instance.go
```diff
@@ -93,6 +95,11 @@ func newInstance(ctl *controller, tenantID string, workload *types.Workload,
 		MACAddress: config.mac,
 		CreateTime: time.Now(),
 		Name:       name,
+		StateLock:  &sync.RWMutex{},
+	}
+
+	if subnet != "" {
```
mcastelino: when will subnet be ""

ciao-controller/instance.go
```diff
@@ -285,6 +321,91 @@ func controllerStorageResourceFromPayload(volume payloads.StorageResource) (s ty
...
+	// send in CIDR notation?
+	networking.PrivateIP = ipAddress.String()
+	mask := net.IPv4Mask(255, 255, 255, 0)
+	ipnet := net.IPNet{
+		IP:   ipAddress.Mask(mask),
+		Mask: mask,
+	}
+	networking.Subnet = ipnet.String()
+
+	cnciInstance, err := tenant.CNCIctrl.GetSubnetCNCI(networking.Subnet)
```
mcastelino: when will the CNCI get created when you end up on a different subnet

ciao-controller/cnci.go
```diff
@@ -0,0 +1,459 @@
...
+// CNCIManager is a structure which defines a manager for CNCI instances
+// TBD: managing multiple CNCI instances.
+type CNCIManager struct {
+	tenant string
+	ctrl   *controller
+
+	// there's no reason to have separate lock for each map.
+	cnciLock *sync.RWMutex
+	cncis    map[string]*CNCI
+	subnets  map[int]*CNCI
```
mcastelino: may be helpful to indicate what the map keys are in each case.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,459 @@
...
+	cnci.transitionState(failed)
+
+	// we should probably not do this, and instead we should
+	// delete from the map?
+	cnci.instance = nil
+
+	return nil
+}
+
+func (c *CNCIManager) waitForActive(subnet int) error {
```
rbradford: Please rethink your use of channels. To do the multiple reader on the same channel you have two options:

a.) Close the channel. All readers will unblock and move forward. Removing the need for all the polling stuff below. We have code just like this in Ciao already at https://github.com/01org/ciao/blob/master/ciao-controller/command.go#L140

Obviously by closing a channel you can only convey completion.

b.) If you want to send the information down the channel then store a slice of channels in the struct (protected by a lock) and have everyone who wants to listen put their channel in the struct and then they can just wait for the answer on their own channel.

c.) Or using a condition variable: https://golang.org/pkg/sync/#Cond
No need for polling in either case.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,459 @@
...
+		cnci.instance = i
+
+		// convert cnci instance string to int for map
+		subnetInt, err := subnetStringToInt(i.Subnet)
+		if err != nil {
+			return nil, err
+		}
+
+		cnci.subnet = subnetInt
+
```
rbradford: Some extra whitespace here and below.

ciao-controller/compute_test.go
```diff
@@ -862,7 +862,8 @@ func testListNodeServers(t *testing.T, httpExpectedStatus int, validToken bool)
 		}
 
 		if result.TotalServers != len(instances) {
-			t.Fatal("Incorrect number of servers")
+			fmt.Printf("result: %v\n", result)
```
rbradford: Left over debugging?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -681,11 +562,52 @@ func (ds *Datastore) ReleaseTenantIP(tenantID string, ip string) error {
...
+			ds.tenantsLock.RUnlock()
+
+			if removeSubnet {
+				// handle error?
+				err := ds.tenants[tenantID].CNCIctrl.RemoveSubnet(i)
+				if err != nil {
+					glog.Warningf("Error removing subnet: (%v)\n", err)
+				}
+			}
+		}()
```
rbradford: In general when you start a goroutine you need to handle tidying up (especially for a long lived one). For this one, create a context.WithCancel() and save the cancel function into a map on the datastore (or in the tenant struct.)

In the goroutine select {} on both a time.After and the Done() channel on the context. On the former do your deletion on the latter just return and terminate the goroutine.

In the datastore shutdown call all the cancel functions  which will safely terminate all the running goroutines. You can also call the cancel if you get a new instance for a pending closed subnet removing the need to even check if its in use.

ciao-controller/client.go
kaccardi: This PR includes cnci -> subnet mapping.

ciao-controller/client.go
```diff
@@ -114,26 +109,48 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
 	return nil
 }
 
-func (client *ssntpClient) removeInstance(instanceID string) {
-	err := client.releaseResources(instanceID)
+func (client *ssntpClient) removeInstance(i *types.Instance) {
```
kaccardi: @rbradford I reviewed the change again and think that this change was made in an earlier revision of the patch, and is no longer relevant. I'll revert it back to the original implementation.

kaccardi: possibly I was also trying to avoid multiple calls to GetInstance in the other routines.

ciao-controller/instance.go
kaccardi: whenever this is a standard tenant instance and not a CNCI.

ciao-controller/instance.go
kaccardi: the CNCI is created when we allocate the tenant IP address down in the datastore. It will know whether this is a new/different subnet at that time and do the launch if needed.

ciao-controller/cnci.go
kaccardi: ok, I'll improve the comments.

ciao-controller/client.go
```diff
@@ -114,26 +109,48 @@ func (client *ssntpClient) releaseResources(instanceID string) error {
...
+	tenant, err := client.ctl.ds.GetTenant(i.TenantID)
+	if err != nil {
+		glog.Warningf("Error retrieving tenant %v", err)
+		return
+	}
+
+	err = client.releaseResources(i)
 	if err != nil {
 		glog.Warningf("Error when releasing resources for deleted instance: %v", err)
```
kaccardi: @rbradford going back and reviewing the original code here, it looks like it was intentionally not returning in order to attempt to try to complete as much of the removal as could be completed. I think that makes sense to keep.

PR #1336, opened 2017-07-06
ciao-down: Add workloads support

markdryan:
The user data definitions have been moved out of the ciao-down source code into separate files and stored in the workloads directory.  The result being that ciao-down is no longer restricted to creating  ciao and clear-container VMs.

Workload definitions must be stored in either

- $GOPATH/src/github.com/01org/ciao/testutil/ciao-down/workloads
- ~/.ciao_down/workloads

In addition to being able to specify the cloud-init file used to bootstrap a new VM, the workload yaml files now contain two new sections that allow the user to specify the resources that will be consumed by the VM, mounts and port mappings and the underlying image upon which the VM will be based.  ciao-down is no longer tied to xenial.

There are some new helper functions that make it easier to write workload definitions, updated documentation and some new workloads for base fedora25 and xenial workloads and a fedora25 based ciao workload.

The command line hasn't changed.  You still do

./ciao-down prepare

for ciao

and 

./ciao-down prepare --vmtype=clearcontainers

for clear containers.  A ciao fedora workload can be created using

./ciao-down prepare --vmtype=ciao-fedora25


markdryan:
@mcastelino , @grahamwhaley  could you give this a test to make sure cc still works for you?

mcastelino:
@markdryan on it now.. Will also try with some custom yaml and cloud-init files.

At this point if this works I would like to delete --vmtype=clearcontainers and just publish yaml and cloud-init.

rbradford:
Interestingly it looks like go is missing from the path inside semaphore:

Installing it seems fine:

> Downloading Go : [OK]
> Unpacking Go : [OK]

but setup.sh fails to find "go"

Could something have changed with the way the PATH is setup?

rbradford:
I've tested this with a few combinations and it works for me. I'll look more closely at the code next.

testutil/ciao-down/workloads/ciao.yaml
```diff
@@ -227,6 +224,9 @@ runcmd:
  - {{endTaskCheck .}}
 {{end}}
 
+ - echo "export GOPATH={{template "GOPATH" . }}" >> /home/{{.User}}/.bashrc
+ - echo "export PATH=$PATH:{{template "GOPATH" . }}/bin:/usr/local/go/bin:/usr/local/nodejs/bin" >> /home/{{.User}}/.bashrc
```
mcastelino: @markdryan putting gopath in /etc/environment may still be needed for the sudo invocation of go tests. The networking unit tests use sudo go test... . Worst case we need to ensure our tests give the full path to like
sudo `which go` test

mcastelino:
@markdryan It may be good to reflect vmtype in ciao-down status. So that the user knows which type of instance is currently setup.

mcastelino:
@markdryan Passes basic sanity tests for me

testutil/ciao-down/instance.go
```diff
@@ -0,0 +1,302 @@
...
+		{
+			Tag:           "hostgo",
+			SecurityModel: "passthrough",
+			Path:          ws.GoPath,
+		},
+	}
+
+	in.PortMappings = []portMapping{
+		{
+			Host:  10022,
```
markdryan: Multiple instances are not yet supported and port mapping is one of the biggest reasons why not.  I was thinking that a pre-cursor to implementing multiple instances would be to move away from SLIRP to a more ciao-like compute node networking, which would solve the port mapping issue as each VM would have it's own IP address.  I'd like to discuss this with you in  a bit more detail first though.

Note that even in this version of ciao-down the host port numbers are not fixed.  You can override them in the .yaml file, or when issuing a prepare and start command via the ports option.

testutil/ciao-down/instance.go
```diff
@@ -0,0 +1,302 @@
...
+	data, err := ioutil.ReadFile(path.Join(ws.instanceDir, "state.yaml"))
+	if err != nil {
+		return fmt.Errorf("Unable to read instance state : %v", err)
+	}
+
+	return in.unmarshall(data)
+}
+
+func (in *instance) save(ws *workspace) error {
+	data, err := yaml.Marshal(in)
```
markdryan: I see.  You're worried that an upgrade to ciao-down might update the yaml files, thereby preventing you from re-creating the same VM that you had before.

There's already a solution for this.  You can copy the .yaml file into ~/.ciao-down/workloads.  The prepare command will still be able to find it and this file will not be touched when upgrading ciao-down.  I was planning to add some additional commands,

ciao-workload list
ciao-workload clone

which would make the copying easier.  However, it would still be a manually process.  If you'd like I could automatically copy the workload yaml file into ~/.ciao-down/workloads during prepare if it doesn't already exist.

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+{{.}}
+   path: /etc/environment
+{{end}}
 ```
-$GOPATH/bin/ciao-down connect
+
+#### download
+
+Download should be used to download files from inside the guest.  Download is preferable
+to direclty downloading files inside the cloudinit document using curl or wget for
```
markdryan: Thanks, I'll fix this.

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+download takes three parameters
+
+- The workspace object
+- The URL of the file to download
+- The location on the guest where the downloaded file should be stored
+
+An example of its usage is
+
 ```
+- {{download . "https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz" "/tmp/go1.8.linux-amd64.tar.gz"}}
```
markdryan: The download command already does this.  When you use download the file is downloaded by the host, placed in ~/.ciao-down/cache and then served up to the guest via HTTP.  The file is then downloaded from the host and save at the path given above on the guest.  The path given here, /tmp/go1.8.linux-amd64.tar.gz is a guest local path and not a host path.   Needless to say, if the file already exists on the host, it doesn't need to be downloaded from the Internet.

The support for host side caching of files is another factor that will complicate multiple instance support.

markdryan:
@mcastelino Some additional responses
> At this point if this works I would like to delete --vmtype=clearcontainers and just publish yaml and cloud-init.

This is what this PR does.  The ciao and cc cloud-init files are no longer hardcoded in ciao-down and there's almost no ciao specific code left in ciao-down apart from the code that migrates the state of old ciao VM instances to the new format.  The value passed to the vmtype option is just the name of the yaml workload you want to create.  workload would be a better name for this option but I deliberately avoided changing the command line to avoid needing to update cc scripts and documentation.

> It may be good to reflect vmtype in ciao-down status. So that the user knows which type of instance is currently setup.

Sure, I can do this.  Not such a huge issue now but going forward when we have multiple instance support this will be vital.

markdryan:
> Could something have changed with the way the PATH is setup?

Interesting indeed.  I don't think it's the PATH as the full path is used to invoke Go.  Perhaps there's something wrong with the proxy functions.  I'll check.

testutil/ciao-down/instance.go
markdryan: The problem with automatically copying files to .workload though is that they shadow the workloads in the source directory.  This would mean that you'd never get any updates to a workload after you'd created a single instance, which isn't good.  I can modify the code to do what you suggest.  The entire workload would be then saved in the instance directory, but only the middle section would be modified by ciao-down.

testutil/ciao-down/prepare.go
```diff
@@ -110,6 +119,26 @@ func hostSupportsNestedKVM() bool {
...
 
+func createWorkload(ws *workspace, vmType string) (*workload, error) {
+	localPath := filepath.Join(ws.Home, ".ciao-down", "workloads",
+		fmt.Sprintf("%s.yaml", vmType))
+	ud, err := ioutil.ReadFile(localPath)
+	if err == nil {
+		return &workload{userData: string(ud)}, nil
+	}
+
+	p, err := build.Default.Import(ciaoDownPkg, "", build.FindOnly)
```
rbradford: This use of build.Default.Import() is cool. I think if we ever tried to package Ciao into debs/rpms this code would need to be modified to look into a system location too.

testutil/ciao-down/ciao_down.go
```diff
@@ -137,41 +137,46 @@ func checkDirectory(dir string) error {
...
 	fs := flag.NewFlagSet("prepare", flag.ExitOnError)
 	vmFlags(fs, &memGiB, &CPUs, &m, &p)
 	fs.BoolVar(&debug, "debug", false, "Enables debug mode")
-	fs.StringVar(&runCmd, "runcmd", "", "Path to a file containing additional commands to execute when preparing the VM")
-	fs.StringVar(&vmType, "vmtype", CIAO, "Type of VM to launch. "+CIAO+" or "+CLEARCONTAINERS)
+	fs.StringVar(&vmType, "vmtype", CIAO, "Type of VM to launch.")
 
 	if err := fs.Parse(flag.Args()[1:]); err != nil {
-		return nil, false, "", "", err
+		return nil, nil, false, err
```
rbradford: How about using some errors.Wrap() here?

testutil/ciao-down/instance.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	in.PortMappings = []portMapping{
+		{
+			Host:  10022,
+			Guest: 22,
+		},
+	}
+
+	if vmType == CIAO {
```
rbradford: Could you put a comment to say what this is for.

testutil/ciao-down/prepare.go
```diff
@@ -234,9 +234,40 @@ func downloadFN(ws *workspace, URL, location string) string {
 	return fmt.Sprintf("wget %s -O %s", url.String(), location)
 }
 
+func beginTaskFN(ws *workspace, message string) string {
```
rbradford: I'm probably being naive here. What does the FN suffix stand for?

testutil/ciao-down/prepare.go
```diff
@@ -260,8 +260,66 @@ func finishedFN(ws *workspace) string {
 	return fmt.Sprintf(finishedStr, ws.HTTPServerPort)
 }
 
+func proxyVarsFN(ws *workspace) string {
+	var buf bytes.Buffer
+	if ws.NoProxy != "" {
+		buf.WriteString("no_proxy=")
+		buf.WriteString(ws.NoProxy)
+		buf.WriteString(" ")
```
rbradford: Are you sure that the software that expects HTTP_PROXY (i.e. capitalised) doesn't also expect NO_PROXY too?

testutil/singlevm/setup_images.sh
```diff
@@ -141,4 +141,4 @@ if [ -f $ubuntu_cloud_image ]; then
         --name "Ubuntu Server 16.04" \
 	--visibility public
 fi
-popd
\ No newline at end of file
+popd
```
rbradford: Good catch!

testutil/ciao-down/ciao_down.go
```diff
@@ -136,11 +150,14 @@ func prepareFlags(ws *workspace) (*workload, bool, error) {
...
 	var p ports
 	var memGiB, CPUs int
+	var update packageUpgrade
 
 	fs := flag.NewFlagSet("prepare", flag.ExitOnError)
 	vmFlags(fs, &memGiB, &CPUs, &m, &p)
 	fs.BoolVar(&debug, "debug", false, "Enables debug mode")
 	fs.StringVar(&vmType, "vmtype", CIAO, "Type of VM to launch.")
+	fs.Var(&update, "package-upgrade",
+		"Hint to enable or disable update of VM packages.  Should be true or false")
```
rbradford: What's the motivation over a simpler BoolVar e.g. --no-package-upgrade with default false?

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+packages.
+
+Once it's finished you'll be able to connect to the the VM via SSH
+using the following command.
+
+```
+ciao-down connect
+```
+
+The command above assumes that either $GOPATH/bin or ~/go/bin is on
```
rbradford: s/is on/is in/

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+your PATH.
+
+ciao-down will cache the Ubuntu image locally so the next time you
+create another xenial based you won't have to wait very long. It also
```
rbradford: s/xenial based/xenial based instance/

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+knows about HTTP proxies and will mirror your host computer's proxy
+settings inside the VMs it creates.
+
+You can delete the VM you've just created by calling,
```
rbradford: nit: I think you call functions but you run commands...

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+
+ciao-down ships with a number of workloads for creating VMs based on standard images,
+such as Ubuntu 16.04 and Fedora 25.  Users are also free to create their own workloads.
+Standard workloads are stored in $GOPATH/src/github.com/01org/ciao/testutil/ciao-down/workloads.
+User created workloads are stored in ~/.ciao-down/workloads.  ciao-down always checks the
+~/.ciao-down/workloads directory first so if a workload exists in both directories
+with the same name, ciao-down will use the workload in ~/.ciao-down/workloads.
+
+You can tell ciao-down which workload to use when creating the VM with the
+prepare command.  This is done via the --vmtype option.  You specify the file name
```
rbradford: I'm proposing a change from --vmtype to --workload as that will make the terminology more consistent

testutil/ciao-down/README.md
```diff
@@ -10,39 +12,336 @@ Then simply type
...
+of the workload without the .yaml extension.  The workload must be present in either
+of the two directories mentioned above.  For example, the prepare command in the
+introduction section used the option --vmtype xenial.  This caused ciao-down to
+load the pre-shipped workload defintion in
```
rbradford: nit: It's not pre-shipped. It's either predefined or shipped :-)

testutil/ciao-down/prepare.go
markdryan: No.  I'll add NO_PROXY

testutil/ciao-down/prepare.go
markdryan: function

testutil/ciao-down/ciao_down.go
markdryan: I guess cause it's not used anywhere else in ciao-down right now.  Migrating to a errors wrap would need to be a separate PR.

testutil/ciao-down/ciao_down.go
markdryan: Because we need tree values:

- ""
- "false"
- "true"

"" means, use the value specified in the workload definition and is the default.

testutil/ciao-down/instance.go
markdryan: Actually, the function name is misleading.  It's called newDefaultInstance, but it should be called newLegacyInstance.  I'll rename it and add a comment.

testutil/ciao-down/README.md
markdryan: Yes, I agree.  The reason it's still called vmtype is that I didn't want to break ciao-down for existing users and scripts.  But I could add a workload option as well which would do the same thing as vmtype.

testutil/ciao-down/README.md
markdryan: okay.  I agree with all your edits and will update the docs.  Thanks for taking the time to read them.

testutil/ciao-down/instance.go
markdryan: All, done.  The entire workload specification is now stored in .ciao-down/instance.  The contents of the second document might differ from the original if the user has overriden the default values for cpus and memory, but the 1st and 3rd (userdata) documents will be the same,

testutil/ciao-down/README.md
markdryan: Fixed

testutil/ciao-down/prepare.go
markdryan: Done

testutil/ciao-down/instance.go
markdryan: Done

testutil/ciao-down/README.md
markdryan: Done

testutil/ciao-down/README.md
markdryan: Done

testutil/ciao-down/README.md
markdryan: Done

testutil/ciao-down/README.md
markdryan: Done

markdryan:
> @markdryan It may be good to reflect vmtype in ciao-down status. So that the user knows which type of instance is currently setup.

@mcastelino This is done now.

markdryan:
Not sure why the semaphore build didn't start on the new set of PRs.  I seem to be unable to trigger it manually.  verify.sh and bat pass for me on ciao-down.

rbradford:
The only feedback I have but you don't have to fix in this version would be for a way to get the list of vmtypes. Maybe in the --usage?

markdryan:
> The only feedback I have but you don't have to fix in this version would be for a way to get the list of vmtypes. Maybe in the --usage?

Totally agree.  I'm planning

ciao-cli workload list
ciao-cli workload delete
ciao-cli workload clone

PR #1338, opened 2017-07-07
Enable graceful shutdown of controller 

rbradford:
This PR modifies controller to handle SIGINT and SIGTERM starting a shutdown process when they are received. To do this the new http.Server.Shutdown() added in Go 1.8 is used. This requires bumping the Go requirement to 1.8 which this PR does.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -22,7 +22,10 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
+	"time"
 
 	"github.com/01org/ciao/ciao-controller/api"
 	"github.com/01org/ciao/ciao-controller/internal/datastore"
@@ -55,6 +58,7 @@ type controller struct {
 	tenantReadiness     map[string]*tenantConfirmMemo
 	tenantReadinessLock sync.Mutex
 	qs                  *quotas.Quotas
+	httpServers         []*http.Server
 }
 
 var cert = flag.String("cert", "", "Client certificate")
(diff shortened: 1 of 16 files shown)


rbradford:
@markdryan Please review
CC @kaccardi

rbradford:
I used `pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)` to reveal the running goroutines on exit and did not see any that shouldn't have been there (with the exception of a sqlite one that I fix in the first commit.)

ciao-controller/openstack_compute.go
```diff
@@ -667,5 +668,13 @@ func (c *controller) startComputeService() error {
...
 
-	return http.ListenAndServeTLS(service, httpsCAcert, httpsKey, r)
+	glog.Infof("Starting compute API on port %d", computeAPIPort)
+
+	server := &http.Server{
+		Handler: r,
+		Addr:    service,
+	}
+
+	c.httpServers = append(c.httpServers, server)
```
markdryan: I think this might be racy as all the methods that start a service are run in their own go routines.

ciao-controller/main.go
```diff
@@ -273,7 +289,23 @@ func (c *controller) startCiaoService() error {
...
+	}
 
-	return http.ListenAndServeTLS(service, httpsCAcert, httpsKey, r)
+	c.httpServers = append(c.httpServers, server)
+	return server.ListenAndServeTLS(httpsCAcert, httpsKey)
+}
+
+func (c *controller) ShutdownHTTPServers() {
+	glog.Warning("Shutting down HTTP servers")
+	for _, server := range c.httpServers {
```
markdryan: This is racy too I think.  Might be easiest create the httpServers in advance and pass them into the goroutines.  They could then be local variables of main.

ciao-controller/main.go
```diff
@@ -273,7 +289,23 @@ func (c *controller) startCiaoService() error {
...
+		err := server.Shutdown(context.TODO())
```
markdryan: We should use a timeout context here.  There's no guarantee that shutdown will ever return.  We could call shutdown in parallel with a time out of say 1 second.

ciao-controller/main.go
```diff
@@ -268,12 +289,27 @@ func (c *controller) startCiaoService() error {
...
+	}
+
+	return server, nil
+}
 
-	return http.ListenAndServeTLS(service, httpsCAcert, httpsKey, r)
+func (c *controller) ShutdownHTTPServers() {
+	glog.Warning("Shutting down HTTP servers")
+	for _, server := range c.httpServers {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
```
markdryan: Maybe we should call shutdown of each server in parallel.  Otherwise, we could end up waiting up to 15 seconds for controller to shutdown in the worst case.

rbradford:
@markdryan all feedback addressed.

PR #1340, opened 2017-07-10
datastore: Return an error when a subnet has no usable IP addresses ￼…

rbradford:
If a /31 or /32 subnet is provided there are not enough IP addresses
available to be used. If a user wants to specify a single IP address
they should not use the /32 notation under the subnet code path but
instead the address code path.

Fixes: #1290


rbradford:
@kaccardi Could you review please?

PR #1343, opened 2017-07-11
ciao-cli: Pool creation/deletion/addition/removal are privileged

rbradford:
These commands are privileged so check in ciao-cli and fatal out if the
user is not privileged. This also avoids the problem of trying to reach
non-existant API entry points generated by the automatic url generation
based on privilege.

Fixes: #1288


rbradford:
Hi @kaccardi would you like to review?

PR #1344, opened 2017-07-11
Complete removal of workload Image ID

rbradford:
The image ID in the workload has been non-functional since we moved to specifying the image ID as source for a disk in the workload. This PR cleans up the remaining references to image ID by removing the field from the database, payloads and the ciao-cli reporting.

Fixes: #988 
Fixes: #1232


markdryan:
Looks okay to me.  I didn't realise there was so much imageid code left.  @kaccardi Is this okay for you?  There are a fair few controller changes here.

PR #1348, opened 2017-07-12
Add more linting coverage to ciao

rbradford:
This adds the following linters to the linters run against ciao on travis: varcheck, structcheck and unused and fixes all the issues identified by running those linters.


templateutils/example_test.go
```diff
@@ -50,7 +50,6 @@ func ExampleGenerateUsageUndecorated() {
 	i := struct {
 		X       int
 		Y       string
-		hidden  float64
```
markdryan: We're going to need to skip this file.  This field is hidden for a reason.  To demonstrate that it is excluded from the auto generated documentation.

rbradford: Yeah, and I think this is overzealous with fields that are actually read by code that is excluded (i.e. in vendor) and is breaking BAT. So i'll refine those removed by "unused" and also not add it to the travis builds.

rbradford:
BAT passes for me under singlevm with this PR now. @markdryan do you want to continue to review?

ciao-controller/controller_test.go
```diff
@@ -1101,109 +1087,6 @@ func testStartWorkload(t *testing.T, num int, fail bool, reason payloads.StartFa
...
-
-	if len(instances) != num {
-		t.Fatalf("Wrong number of instances, expected %d, got %d", num, len(instances))
-	}
-
-	instanceCh <- instances
-}
-
-// NOTE: the caller is responsible for calling Shutdown() on the *SsntpTestClient
-func testStartWorkloadLaunchCNCI(t *testing.T, num int) (*testutil.SsntpTestClient, []*types.Instance) {
```
markdryan: @kaccardi,@rbradford I'm wondering whether we actually want this code, but that it is not being run as it is named incorrectly, i.e., it starts with a lower case t.  It does sort of look like a unit test.

markdryan:
@rbradford I've just got one query about that controller test.  Otherwise it looks good to me.

ciao-controller/controller_test.go
rbradford: The refactoring in 2fc83a87053fd35a94eda24bd828b68b08e0fd05 deleted the caller of this function. From that commit I can't infer whether it was intentional or not. We need @kaccardi to confirm.

kaccardi: @rbradford I actually have started using it again in my in progress PR for cnci launching, so I don't want to delete it, although you are correct that on master it's not being used.

rbradford:
@kaccardi Cool. i've updated the PR not to delete any of the unused functions in controller_test.go. After you're done with the bulk of the work on your PR it would be great if you would run "unused" on the tree to see if any of those functions are now suitable candidates for deletion.

@markdryan After Travis passes this is good to go then.

PR #1358, opened 2017-07-13
SingleVM: Setup the NAT rules dynamically within ciao-down

mcastelino:
The single VM script had a hard coded ubuntu specific interface
naming convention. Now that ciao-down supports all types of
linux distributions, remove this assumption and discover the
primary outgoing interface name dynamically.

### DIFF
--- a/testutil/singlevm/cleanup.sh
+++ b/testutil/singlevm/cleanup.sh
@@ -4,6 +4,7 @@
 
 ciao_gobin="$GOPATH"/bin
 ciao_host=$(hostname)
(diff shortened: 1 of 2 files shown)


mcastelino:
@markdryan this was a latent bug due to a hardcoded interface naming assumption. With this external network access is working for fedora based ciao-down.

Please note that ping will not work due to the fact that qemu SLIRP networking does not allow this
http://wiki.qemu.org/Documentation/Networking#User_Networking_.28SLIRP.29 unless you explicitly configure your host to allow for it.

rbradford:
Tested Fedora & Kubicle. LGTM!

PR #1362, opened 2017-07-18
ciao-down prepare

dlespiau:
As discussed with Mark and Rob, sadly offline, `ciao-down create` would be a nicer name. Also we want to detach ciao-down from ciao a bit and have explicit workload names instead of defaulting to the ciao workload.

### DIFF
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -46,8 +46,8 @@ const (
 func init() {
 	flag.Usage = func() {
 		fmt.Fprintf(os.Stderr, "Usage of %s:\n\n", os.Args[0])
-		fmt.Fprintf(os.Stderr, "%s [prepare|start|stop|quit|status|connect|delete]\n\n", os.Args[0])
-		fmt.Fprintln(os.Stderr, "- prepare : creates a new VM for ciao or clear containers")
+		fmt.Fprintf(os.Stderr, "%s [create|start|stop|quit|status|connect|delete]\n\n", os.Args[0])
+		fmt.Fprintln(os.Stderr, "- create : creates a new VM")
 		fmt.Fprintln(os.Stderr, "- start : boots a stopped VM")
 		fmt.Fprintln(os.Stderr, "- stop : cleanly powers down a running VM")
 		fmt.Fprintln(os.Stderr, "- quit : quits a running VM")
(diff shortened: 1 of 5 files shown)


testutil/ciao-down/ciao_down.go
```diff
@@ -144,32 +144,41 @@ func checkDirectory(dir string) error {
...
 	fs.Var(&update, "package-upgrade",
 		"Hint to enable or disable update of VM packages.  Should be true or false")
 
 	if err := fs.Parse(flag.Args()[1:]); err != nil {
 		return nil, false, err
 	}
 
+	if fs.NArg() != 1 {
+		fs.Usage()
+		os.Exit(1)
```
markdryan: Better to return an error here rather than exit straight away as by exiting we're by-passing ciao-down's normal shutdown sequence.  The problem with exiting is that your defers don't get called so someone adding a defer to this function or a function further up in the callstack would be surprised to discover that their deferred function does not get called.

markdryan:
@grahamwhaley , @mcastelino I'd like your feedback on this as this is an API breaking change for ciao-down.   Once it's merged, for CC you do

ciao-down create clear-containers

rather than

ciao-down prepare --vm-type clear-containers

Will this break anything for you?  Is there any documentation you need to update?

markdryan:
We'll need to update semaphore ci once this is merged and also the https://github.com/01org/ciao/blob/master/DeveloperQuickStart.md

dlespiau:
PR updated:

- Don't os.exit() in createFlags() directly,
- Add a missing Fixes tag.

markdryan:
@dlespiau regarding the Fixes tag, we usually specify the bug number preceeded by a #.  Does github automatically close the issues if you provide the URL directly?   If so, great.

dlespiau:
yes it does! it has the added benefit that one can click on the link directly from `git log`,

grahamwhaley:
Hi @markdryan It won't affect me personally as I generally use `ciao-down` in a hand-run mode. The only place I'm aware of for CC where we will have to edit is https://github.com/01org/cc-oci-runtime/blob/master/documentation/Quickstart-Clear-Containers-in-a-Ubuntu-VM.md#ciao-down

markdryan:
@dlespiau I think we can merge this.  Could you create a patch for the cc documentation @grahamwhaley links to?  I'll merge and fix semaphore so the build passes.

PR #1367, opened 2017-07-18
ciao-down: add support to specify a workload file as workload 

### LINKED ISSUE #1366: ciao-down: create instances from workload files
would be nice to have an option to specify a custom workload file

devimc:
With this patch create command is able to create instances from
workload files specified as workload argument.
The custom workload file can be provided using 3 different schemas:
* http
* https
* file

or throght its absolute or relative path

* ../../relative/path
* /absolute/path

for example to create an instance from a file hosted in a remote server:

```
$ ciao-down create https://raw.githubusercontent.com/01org/ciao/master/testutil/ciao-down/workloads/fedora25.yaml
```

fixes #1366

### DIFF
--- a/testutil/ciao-down/README.md
+++ b/testutil/ciao-down/README.md
@@ -65,11 +65,29 @@ User created workloads are stored in ~/.ciao-down/workloads.  ciao-down always c
 with the same name, ciao-down will use the workload in ~/.ciao-down/workloads.
 
 Which workload to use is specified when creating the VM with the create
-command.  This is the name the workload file without the .yaml extension.  The
-workload must be present in either of the two directories mentioned above.  For
-example, the create command in the introduction section creates an Ubuntu 16.04
+command. The workload can be a file without the .yaml extension or a URI pointing to
+a local or remote file. If the workload is a file without the .yaml extension then it
+must be present in either of the two directories mentioned above. For example,
+the create command in the introduction section creates an Ubuntu 16.04
 workload (xenial).  This caused ciao-down to load the workload definition in
 $GOPATH/src/github.com/01org/ciao/testutil/ciao-down/workloads/xenial.yaml.
+In the case of a remote file the supported schemes are http, https and file. Be aware
+of the remote file will not be saved hence each time this option is used, ciao-down
+will try to get it from the remote location. For example, to create a workload using the
+file /home/x/workload.yaml we have two options.
+
+Using the file scheme:
+
+```
+ciao-down create file:///home/ciao-down/workload.yaml
+```
+
+or its absolute path:
+
+```
+ciao-down create  /home/ciao-down/workload.yaml
+```
+
 
 As the majority of the ciao-down workloads are cloud-init files, ciao-down only
 works with images that are designed to run cloud-init on their first boot.  Typically,
--- a/testutil/ciao-down/workload.go
+++ b/testutil/ciao-down/workload.go
@@ -19,12 +19,16 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"go/build"
 	"io/ioutil"
+	"net/url"
+	"os"
 	"path"
 	"path/filepath"
 	"regexp"
+	"time"
 
 	yaml "gopkg.in/yaml.v2"
 )
@@ -75,10 +79,55 @@ func (wkld *workload) save(ws *workspace) error {
 	return nil
 }
 
-func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
+func workloadFromURL(ctx context.Context, u url.URL) ([]byte, error) {
+	var workloadPath string
+
+	switch u.Scheme {
+	case "http", "https":
+		workloadFile, err := ioutil.TempFile("", ".workload")
+		if err != nil {
+			return nil, fmt.Errorf("Failed to create a temporal file: %s", err)
+		}
+
+		workloadPath = workloadFile.Name()
+		defer func() { _ = os.Remove(workloadPath) }()
+
+		// 60 seconds should be enough to download the workload file
+		ctx, cancelFunc := context.WithTimeout(ctx, 60*time.Second)
+		err = getFile(ctx, u.String(), workloadFile, downloadProgress)
+		cancelFunc()
+		if err != nil {
+			return nil, fmt.Errorf("Unable download workload file from %s: %v", u.String(), err)
+		}
+	case "file":
+		workloadPath = u.Path
+
+	default:
+		return nil, fmt.Errorf("Unable download workload file %s: unsupported scheme", u.String())
+	}
+
+	return ioutil.ReadFile(workloadPath)
+}
+
+func loadWorkloadData(ctx context.Context, ws *workspace, workloadName string) ([]byte, error) {
+	u, err := url.Parse(workloadName)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse workload name %s: %s", workloadName, err)
+	}
+
+	// Absolute means that it has a non-empty scheme
+	if u.IsAbs() {
+		return workloadFromURL(ctx, *u)
+	}
+
+	wkld, err := ioutil.ReadFile(workloadName)
+	if err == nil {
+		return wkld, nil
+	}
+
 	localPath := filepath.Join(ws.Home, ".ciao-down", "workloads",
 		fmt.Sprintf("%s.yaml", workloadName))
-	wkld, err := ioutil.ReadFile(localPath)
+	wkld, err = ioutil.ReadFile(localPath)
 	if err == nil {
 		return wkld, nil
 	}
@@ -113,8 +162,8 @@ func unmarshalWorkload(ws *workspace, wkld *workload, insSpec, insData,
 	return nil
 }
 
-func createWorkload(ws *workspace, workloadName string) (*workload, error) {
-	data, err := loadWorkloadData(ws, workloadName)
+func createWorkload(ctx context.Context, ws *workspace, workloadName string) (*workload, error) {
+	data, err := loadWorkloadData(ctx, ws, workloadName)
 	if err != nil {
 		return nil, err
 	}
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -145,7 +145,7 @@ func checkDirectory(dir string) error {
 	return nil
 }
 
-func createFlags(ws *workspace) (*workload, bool, error) {
+func createFlags(ctx context.Context, ws *workspace) (*workload, bool, error) {
 	var debug bool
 	var m mounts
 	var p ports
@@ -179,7 +179,7 @@ func createFlags(ws *workspace) (*workload, bool, error) {
 		}
 	}
 
-	wkl, err := createWorkload(ws, workloadName)
+	wkl, err := createWorkload(ctx, ws, workloadName)
 	if err != nil {
 		return nil, false, err
 	}
@@ -246,7 +246,7 @@ func create(ctx context.Context, errCh chan error) {
 		return
 	}
 
-	wkld, debug, err := createFlags(ws)
+	wkld, debug, err := createFlags(ctx, ws)
 	if err != nil {
 		return
 	}


rbradford:
@markdryan quick refinement, I think absolute paths (i.e. beginning with /) are much nicer than file URIs from a UX perspective for local paths.

markdryan:
> @markdryan quick refinement, I think absolute paths (i.e. beginning with /) are much nicer than file URIs from a UX perspective for local paths.

Yes, agreed.  The problem is that if we allow normal paths this creates an ambiguity if no path is specified at all.  For example,

ciao-cli create test

Does test refer to a file in one of the workloads directories or a file in the current directory?  We would expect the former but users may expect the latter.  With URLs there is no ambiguity.  We can work around this but we need to decide what the correct behaviour is in this case and document it.  Perhaps if there's no path, we look in the current directory first and then in the workload directories.

devimc:
@markdryan  done

rbradford:
@markdryan  that's why I said *absolute paths* they by definition remove the ambiguity as they start with a /

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
 }
 
 func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
+	var absPath string
+	u, err := url.Parse(workloadName)
+	if err == nil {
+		switch u.Scheme {
+		case "http", "https":
+			workloadFile, err := ioutil.TempFile("/tmp", ".workload")
```
rbradford: nit: prefer "" to "/tmp"

rbradford: Does this file need to be closed?

markdryan:
> @markdryan that's why I said absolute paths they by definition remove the ambiguity as they start with a /

But the problem is that once you start accepting absolute paths users may logically expect to use relative paths as they can with any other command.  This is where the confusion and ambiguity sneak in.  So we need to come up with some plan to address this.

testutil/ciao-down/workload.go
markdryan: Yes

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+			if err != nil {
+				break
+			}
+
+			absPath = workloadFile.Name()
+			defer os.Remove(absPath)
+
+			// 10 seconds should be enough to download the workload file
```
markdryan: Not sure.  10 seconds might be too little.  Let's bump this up a bit to 1 minute maybe

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
 }
 
 func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
```
markdryan: This function and the function that call it need to take a context parameter.  The context parameter will be passed in to createWorkload from the createFlags function.  It must be passed to the context.WithTimeout call below.  If you do this, hitting CTRL-C at the console will cancel the context you pass to getFile which is what we want.

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+			workloadFile, err := ioutil.TempFile("/tmp", ".workload")
+			if err != nil {
+				break
+			}
+
+			absPath = workloadFile.Name()
+			defer os.Remove(absPath)
+
+			// 10 seconds should be enough to download the workload file
+			ctx, cancelFunc := context.WithTimeout(context.Background(), 10*time.Second)
```
markdryan: context.Background() is not the correct context, see above.

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+			defer cancelFunc()
```
markdryan: There's no need to use defer here.

Just do

```
err = getFile(ctx, workloadName, workloadFile, downloadProgress)
cancelFunc()
if err ==

File: testutil/ciao-down/workload.go [hunk 6]
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
 }
 
 func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
+	var absPath string
+	u, err := url.Parse(workloadName)
```
markdryan: Can we split all this out into a separate function?

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+				return nil, fmt.Errorf("Unable download workload file from %s: %v", workloadName, err)
+			}
+		case "file":
+			absPath = path.Join("/", u.Host, u.Path)
+		default:
+			absPath, _ = filepath.Abs(u.Path)
+		}
+	}
+
+	if strings.Compare(absPath, "") != 0 {
```
markdryan: Use == rather than strings.Compare

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+		wkld, err := ioutil.ReadFile(absPath)
```
markdryan: So here's a question.  When we've downloaded the file or read it from another location on our disk, do we want to store the file in our ~/.ciao-down/.workloads directory so we can use it to create other instances without having to download it again?  This is not what the code currently does, and maybe that is fine, but I wanted to raise the question.

@dlespiau, @rbradford any preferences here?

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+		if err == nil {
+			return wkld, nil
```
markdryan: Don't we need to return here if we get an error as well.  There's no point in falling through to the existing code.

testutil/ciao-down/workload.go
```diff
@@ -76,6 +81,40 @@ func (wkld *workload) save(ws *workspace) error {
...
+	if err == nil {
+		switch u.Scheme {
+		case "http", "https":
+			workloadFile, err := ioutil.TempFile("/tmp", ".workload")
+			if err != nil {
+				break
+			}
+
+			absPath = workloadFile.Name()
+			defer os.Remove(absPath)
```
markdryan: Could you call defer like this

defer func() { _ = os.Remove(absPath) }()

This is the idiom we use elsewhere in the code an the reason for doing this is that it allows us to explicitly discard errors from deferred functions.  This is important if we want a clean report from errcheck.

markdryan:
@devimc Sorry, I should have started a review rather than making lots of small comments.  Overall, the approach looks good but I have lots of small comments.  Also, I have one design question?  Should we place downloaded workloads in ~/.ciao-down/workloads.

testutil/ciao-down/workload.go
devimc: this file will be closed by ```getFile``` 
https://github.com/01org/ciao/blob/master/testutil/ciao-down/download.go#L77

devimc:
> Also, I have one design question? Should we place downloaded workloads in ~/.ciao-down/workloads.

I think that is not needed, doing that we could override different workloads with the same name and I wouldn't like that ciao-down overrides my custom workloads

testutil/ciao-down/workload.go
devimc: ohh ok, make sense, fixing
thanks

testutil/ciao-down/workload.go
devimc: sure I can change it

testutil/ciao-down/workload.go
devimc: yup

testutil/ciao-down/workload.go
devimc: ehmm, Go recommends to use ```strings.Compare``` since it has a better performance that ```==```

> Compare is included only for symmetry with package bytes. It is usually clearer and always faster to use the built-in string comparison operators ==, <, >, and so on.

testutil/ciao-down/workload.go
devimc: override my custom workloads (~/.ciao-down/.workloads)?, ehhmm nop, probably I just want to test a remote workload file, if I like it then I can download it in ~/.ciao-down/.workloads

testutil/ciao-down/workload.go
devimc: fixing, thanks

testutil/ciao-down/workload.go
devimc: I wouldn't like to have functions used by only one function

testutil/ciao-down/workload.go
devimc: ok, fixing

testutil/ciao-down/workload.go
devimc: good catch, fixing thanks

testutil/ciao-down/workload.go
devimc: ehmm nop, we should check the workloads dirs

testutil/ciao-down/workload.go
markdryan: Actually, it's recommending the opposite.  The text you've quoted says use "==" as it's faster.

testutil/ciao-down/workload.go
markdryan: Okay, good point.  We'd have to cope with a pre-existing file with the same name.  Let's leave it the way it is.  Maybe we could add an import-workload command in the future that downloaded to the workload directory.

testutil/ciao-down/workload.go
markdryan: Isn't that bound to fail though?  

The variable localPath would get initialised to something weird

```
localPath := filepath.Join(ws.Home, ".ciao-down", "workloads",
		fmt.Sprintf("%s.yaml", workloadName))
```

localPath would end up being "http://myurl.com/workload.yaml" which can't be opened by ReadFile

testutil/ciao-down/workload.go
markdryan: There's no problems with having one function that is called just be one other function.  Splitting code up into functions is not just about code re-use.  It's also about readability.  When functions get too long they get hard to read and hard to maintain.  The piece of code you're adding here is fairly self contained and naturally fits in its own function.  Please split.

testutil/ciao-down/workload.go
markdryan: Yes, good point.  Forgotten how my own code works.

markdryan:
We'll need to update the [README](https://github.com/01org/ciao/blob/master/testutil/ciao-down/README.md) as well.

testutil/ciao-down/workload.go
```diff
@@ -75,7 +80,41 @@ func (wkld *workload) save(ws *workspace) error {
...
+			ctx, cancelFunc := context.WithTimeout(ctx, 60*time.Second)
+			cancelFunc()
+			err = getFile(ctx, workloadName, workloadFile, downloadProgress)
+			if err != nil {
+				return nil, fmt.Errorf("Unable download workload file from %s: %v", workloadName, err)
+			}
+		case "file":
+			absPath = path.Join("/", u.Host, u.Path)
+		default:
+			absPath, _ = filepath.Abs(u.Path)
```
markdryan: Maybe we should just return an error here.  For example, if someone were to specify

 ftp://user:password@host:port/path 

as a URL we should not be trying to access an absolute path on the local machine.

Also, what does

url.Parse("ciao") return?  Does this error out?

testutil/ciao-down/workload.go
devimc: ohh crap, sorry

testutil/ciao-down/workload.go
```diff
@@ -75,7 +79,47 @@ func (wkld *workload) save(ws *workspace) error {
 	return nil
 }
 
-func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
+func loadWorkloadData(ctx context.Context, ws *workspace, workloadName string) ([]byte, error) {
+	var absPath string
+	u, err := url.Parse(workloadName)
+	if err == nil {
```
markdryan: Why don't we just return here if we have an error?

url.Parse("ciao") does not return an error so what is the point of falling through to the workload directory scanning code.

testutil/ciao-down/workload.go
```diff
@@ -75,7 +79,47 @@ func (wkld *workload) save(ws *workspace) error {
...
+
+			// 60 seconds should be enough to download the workload file
+			ctx, cancelFunc := context.WithTimeout(ctx, 60*time.Second)
+			cancelFunc()
+			err = getFile(ctx, workloadName, workloadFile, downloadProgress)
+			if err != nil {
+				return nil, fmt.Errorf("Unable download workload file from %s: %v", workloadName, err)
+			}
+		case "file":
+			absPath = path.Join("/", u.Host, u.Path)
```
markdryan: Won't u.Host be "" if this is a file:///?

testutil/ciao-down/workload.go
```diff
@@ -75,7 +79,47 @@ func (wkld *workload) save(ws *workspace) error {
 	return nil
 }
 
-func loadWorkloadData(ws *workspace, workloadName string) ([]byte, error) {
+func loadWorkloadData(ctx context.Context, ws *workspace, workloadName string) ([]byte, error) {
+	var absPath string
+	u, err := url.Parse(workloadName)
+	if err == nil {
+		switch u.Scheme {
```
markdryan: This code needs to move into a separate function.  In loadWorkloadData we need  (comments are optional)

```
// Check to see whether workloadName might refer to something in our workload directories
// If this condition fails we know that workloadName cannot be an absolute path.
if filepath.Base(workloadName) != workloadName {
   return workloadFromURL(u, ctx)
}

// Fall through to the existing code.

```

Put all the new URL handling code in workloadFromURL.  It will figure out the correct path, downloading the file if necessary and return either an error or the bytes of the workload.

testutil/ciao-down/README.md
```diff
@@ -65,11 +65,29 @@ User created workloads are stored in ~/.ciao-down/workloads.  ciao-down always c
 with the same name, ciao-down will use the workload in ~/.ciao-down/workloads.
 
 Which workload to use is specified when creating the VM with the create
-command.  This is the name the workload file without the .yaml extension.  The
-workload must be present in either of the two directories mentioned above.  For
-example, the create command in the introduction section creates an Ubuntu 16.04
+command. The workload can be a file without the .yaml extension or a URI pointing to
+a local or remote file. If the workload is a file without the .yaml extension then it
+must be present in in either of the two directories mentioned above. For example,
```
devimc: good catch, fixing

testutil/ciao-down/workload.go
```diff
@@ -75,7 +79,47 @@ func (wkld *workload) save(ws *workspace) error {
...
+			if err != nil {
+				break
+			}
+
+			absPath = workloadFile.Name()
+			defer func() { _ = os.Remove(absPath) }()
+
+			// 60 seconds should be enough to download the workload file
+			ctx, cancelFunc := context.WithTimeout(ctx, 60*time.Second)
+			cancelFunc()
```
devimc: yep, fixing

testutil/ciao-down/workload.go
devimc: yes, you're right

devimc:
@markdryan changes applied, thanks for your review

testutil/ciao-down/workload.go
```diff
@@ -75,10 +79,55 @@ func (wkld *workload) save(ws *workspace) error {
...
+
+func loadWorkloadData(ctx context.Context, ws *workspace, workloadName string) ([]byte, error) {
+	u, err := url.Parse(workloadName)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse workload name %s: %s", workloadName, err)
+	}
+
+	// Absolute means that it has a non-empty scheme
+	if u.IsAbs() {
+		return workloadFromURL(ctx, *u)
```
markdryan: Why not just pass the pointer?

devimc: because the function does not modify the object

markdryan: Fair enough.  Let me give this a quick test.

PR #1376, opened 2017-07-19
Create GetTenantInstance

kaccardi:
Make sure that any instance that a tenant requests to assign an external IP to is actually owned by the tenant by creating a new method "GetTenantInstance(tenantID, instanceID)". This method will only return the instance if it is owned by the tenant. In the case of an admin user, mapping external IPs to any valid instance will still be allowed, so the original GetInstance method will still be used.

Finally, use the new GetTenantInstance method to retrieve instances for verification of ownership in our apis.

### DIFF
--- a/ciao-controller/api.go
+++ b/ciao-controller/api.go
@@ -347,15 +347,11 @@ func serversAction(c *controller, w http.ResponseWriter, r *http.Request) (APIRe
 	if len(servers.ServerIDs) > 0 {
 		for _, instanceID := range servers.ServerIDs {
 			// make sure the instance belongs to the tenant
-			instance, err := c.ds.GetInstance(instanceID)
-
+			_, err := c.ds.GetTenantInstance(tenant, instanceID)
 			if err != nil {
 				return errorResponse(err), err
 			}
 
-			if instance.TenantID != tenant {
-				return errorResponse(err), err
-			}
 			actionFunc(instanceID)
 		}
 	} else {
--- a/ciao-controller/api/api.go
+++ b/ciao-controller/api/api.go
@@ -396,6 +396,7 @@ func listMappedIPs(c *Context, w http.ResponseWriter, r *http.Request) (Response
 }
 
 func mapExternalIP(c *Context, w http.ResponseWriter, r *http.Request) (Response, error) {
+	vars := mux.Vars(r)
 	var req types.MapIPRequest
 
 	body, err := ioutil.ReadAll(r.Body)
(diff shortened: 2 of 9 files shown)


ciao-controller/api.go
```diff
@@ -347,15 +347,12 @@ func serversAction(c *controller, w http.ResponseWriter, r *http.Request) (APIRe
 	if len(servers.ServerIDs) > 0 {
 		for _, instanceID := range servers.ServerIDs {
 			// make sure the instance belongs to the tenant
-			instance, err := c.ds.GetInstance(instanceID)
+			_, err := c.ds.GetTenantInstance(tenant, instanceID)
```
rbradford: nit: Whilst you're editing this bit can you kill the whitespace between the line that generates the error and where it is tested.

ciao-controller/api/api.go
```diff
@@ -408,7 +409,12 @@ func mapExternalIP(c *Context, w http.ResponseWriter, r *http.Request) (Response
 		return errorResponse(err), err
 	}
 
-	err = c.MapAddress(req.PoolName, req.InstanceID)
+	tenantID, ok := vars["tenant"]
+	if !ok {
+		tenantID = "public"
```
rbradford: I'd rather not overload tenantID == "public" to mean admin for the MapAddress() case. I'd rather go with "" or something like "admin". But see my other comment below...

ciao-controller/external_ip.go
```diff
@@ -180,10 +180,16 @@ func (c *controller) ListMappedAddresses(tenant *string) []types.MappedIP {
...
 }
 
-func (c *controller) MapAddress(poolName *string, instanceID string) (err error) {
+func (c *controller) MapAddress(tenantID string, poolName *string, instanceID string) (err error) {
 	var m types.MappedIP
+	var i *types.Instance
 
-	i, err := c.ds.GetInstance(instanceID)
+	if tenantID == "public" {
+		// we allow the admin to map anyone's instance
```
rbradford: Do we generally let admins do activity on user resources?

kaccardi: we are inconsistent. We do in many cases, such as deleting an instance. We don't in some cases, such as deleting a workload.

kaccardi: in the case of forcibly unmapping an external IP, I think it makes sense. In the case of mapping the external IP for a user, I think it could also make sense but the argument isn't as strong.

ciao-controller/api.go
kaccardi: sure

ciao-controller/api/api.go
kaccardi: @rbradford I switched it to use "" to indicate an admin map.

PR #1377, opened 2017-07-20
ciao-down: Make workload and state files be only 2 YAML documents 

dlespiau:
A bit of simplification in how workloads are written by folding the second document into a `vm` object in the workload description (1st document).

I kept compatibility with the current (and past) workloads by being able to load 1, 2 and 3 YAML documents description and state files. A "migration" happens when saving the state file.

The 2 first commits are part of PR https://github.com/01org/ciao/pull/1372 that should be merged first, but included them here so the two PRs don't conflict.


dlespiau:
Note that the various renamings are debatable :) Not entirely convinved this makes the most sense yet.

PR #1382, opened 2017-07-21
Add ciao-deploy create-bat-workloads and use in SingleVM

rbradford:
This PR introduces a new tool, ciao-deploy with commands called "create-bat-workloads" and "create-cnci" which will create the BAT workload on a running cluster and setup the CNCI respectively.

For CNCI this command downloads the image and customises it and then uploads it to the cluster

For BAT this encompasses all steps from downloading the image from the remote location, uploading the image to controller and creating the workload derived from it.

This PR should be ready for review immediately.

It introduces a dependency on Cobra (and vendors it) which is the command line tool used by k8s and provides a good framework for managing commands & subcommands, etc. This dependency was discussed with @markdryan in person and it seems like this would also be a good solution for ciao-cli in the future.


rbradford:
Semaphore passed first time \o/

rbradford:
@mcastelino  @kaccardi  Now with CNCI setup too. I had thought you might want to test and land this separately before the full cluster setup is in place. However that should be happening within the next week so i'm easy either way. Unfortunately the PR is only going get bigger however it doesn't touch very much outside ciao-deploy.

PR #1383, opened 2017-07-24
ciao-controller: Improve error handling in datastore

rbradford:
This PR works to try and ensure that the cache maintained by datastore is consistent with the database. It also works towards being able to enable errcheck for controller.

CNCI / IP related code is not touched to avoid too badly conflicting with #1333

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -1086,12 +1086,22 @@ func (ds *Datastore) DeleteInstance(instanceID string) error {
 	}
 
 	msg := fmt.Sprintf("Deleted Instance %s", instanceID)
-	ds.db.logEvent(tenantID, string(userInfo), msg)
-
-	return nil
+	return errors.Wrap(ds.db.logEvent(tenantID, string(userInfo), msg), "Error logging event")
 }
 
 func (ds *Datastore) updateInstanceStatus(status, instanceID string) error {
+	stats := []payloads.InstanceStat{
+		{
+			InstanceUUID: instanceID,
+			State:        status,
+		},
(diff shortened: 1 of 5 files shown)


rbradford:
@kaccardi please can you review. This is what we were chatting about earlier.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -845,6 +845,12 @@ func (ds *Datastore) GetAllInstancesByNode(nodeID string) ([]*types.Instance, er
...
 // The instance will be updated both in the cache and in the database
 func (ds *Datastore) AddInstance(instance *types.Instance) error {
+	err := ds.db.addInstance(instance)
+
+	if err != nil {
+		return errors.Wrap(err, "Error adding instance to database")
+	}
+
 	// add to cache
 	ds.instancesLock.Lock()
```
kaccardi: do you think we should lock the cache, update the persistent store, update the cache, unlock the cache? Can't decide if this is paranoid overkill.

kaccardi:
As I mentioned too - I'd be interested in discovering the extent of the performance hit we'll encounter with this change, although this isn't a blocker of course, just something we can be informed about.

rbradford:
@kaccardi Great question re, locking. And something I thought about too. I don't think it buys us anything as 
a.) the datastore locks are somewhat orthogonal to the database tables themselves,
b.) we have very few cross table dependencies (and even then we don't update the multiple tables together in a single transaction)

But also it does come at a cost for reads in other goroutines for commonly used caches like instances or tenants. As such I think it's best if we don't extend the lock to cover the database access as I think it might do more harm than good.

PR #1387, opened 2017-07-28
ciao-controller: datastore: implement GetNodeSummary from cache

kaccardi:
Rather than retrieving the node summary from the sqlite datastore,
just retrieve it faster from the cache. Remove getNodeSummary from
sqllitedb because that query is no longer used.


ciao-controller/internal/datastore/datastore.go
```diff
@@ -1490,9 +1489,45 @@ func (ds *Datastore) GetCNCIWorkloadID() (string, error) {
...
-	return ds.db.getNodeSummary()
+	var nodes []*types.NodeSummary
+
+	ds.nodesLock.RLock()
+
+	for _, n := range ds.nodes {
+		var total []*types.Instance
+		var running []*types.Instance
+		var pending []*types.Instance
+		var paused []*types.Instance
```
rbradford: As we don't actually care about the instances themselves can we we just use some ints here? Or even better just increment the values in the summary struct:

```
 summary.TotalInstances++
 switch i.State {
 	case payloads.Pending:
 	        // Do nothing?
 	case payloads.Running:
 		summary.TotalRunnningInstances++
 	case payloads.Exited:
 		summary.TotalPausedInstances++
}
```
WDYT?

kaccardi: fine with me. will send revision.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1490,9 +1489,39 @@ func (ds *Datastore) GetCNCIWorkloadID() (string, error) {
...
+		for _, i := range n.instances {
+			if i.CNCI == true {
+				continue
+			}
+
+			summary.TotalInstances++
+
+			switch i.State {
+			case payloads.Pending:
+				summary.TotalPausedInstances++
```
rbradford: summary.TotalPendingInstances++ ?

kaccardi: sorry for the noise - coding while distracted, will try to shape up.

PR #1391, opened 2017-08-04
qemu: Update blockdev-add qmp command to support newer qemu versions

amshinde:
With qemu 2.9, the qmp block-dev command was updated  from:
{ "execute": "blockdev-add", "arguments": { "options": { ... } } }

to:
{ "execute": "blockdev-add", "arguments": { ... } }

Also, instead of id, blockdev-add now requires a node-name for the
root node(https://wiki.qemu.org/index.php/ChangeLog/2.9)

Store the version information with QMPStart and use that to issue
qmp command for adding block devices in the correct format.

Signed-off-by: Archana Shinde <archana.m.shinde@intel.com>

### DIFF
--- a/qemu/qmp.go
+++ b/qemu/qmp.go
@@ -127,6 +127,7 @@ type QMP struct {
 	cfg            QMPConfig
 	connectedCh    chan<- *QMPVersion
 	disconnectedCh chan struct{}
+	version        *QMPVersion
 }
 
 // QMPVersion contains the version number and the capabailities of a QEMU
@@ -533,7 +534,6 @@ func QMPStart(ctx context.Context, socket string, cfg QMPConfig, disconnectedCh 
 
 	connectedCh := make(chan *QMPVersion)
 
-	var version *QMPVersion
 	q := startQMPLoop(conn, cfg, connectedCh, disconnectedCh)
 	select {
 	case <-ctx.Done():
@@ -542,13 +542,13 @@ func QMPStart(ctx context.Context, socket string, cfg QMPConfig, disconnectedCh 
 		return nil, nil, fmt.Errorf("Canceled by caller")
 	case <-disconnectedCh:
 		return nil, nil, fmt.Errorf("Lost connection to VM")
-	case version = <-connectedCh:
-		if version == nil {
+	case q.version = <-connectedCh:
+		if q.version == nil {
 			return nil, nil, fmt.Errorf("Failed to find QMP version information")
 		}
 	}
 
-	return q, version, nil
+	return q, q.version, nil
 }
 
 // Shutdown closes the domain socket used to monitor a QEMU instance and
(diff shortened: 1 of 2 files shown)


rbradford:
@amshinde Thanks for the patch. Do we need to do something to support multiple versions of qemu? Ciao expects to work on the system version of qemu in Ubuntu 16.04 and Fedora 25/26.

qemu/qmp.go
```diff
@@ -542,13 +543,13 @@ func QMPStart(ctx context.Context, socket string, cfg QMPConfig, disconnectedCh
 		return nil, nil, fmt.Errorf("Canceled by caller")
 	case <-disconnectedCh:
 		return nil, nil, fmt.Errorf("Lost connection to VM")
-	case version = <-connectedCh:
+	case q.version = <-connectedCh:
 		if version == nil {
```
markdryan: This should be if q.version == nil.

Also you need to delete the var version *QMPVersion above which is no longer used.

qemu/qmp.go
```diff
@@ -602,16 +603,30 @@ func (q *QMP) ExecuteQuit(ctx context.Context) error {
 // used to name the device.  As this identifier will be passed directly to QMP,
 // it must obey QMP's naming rules, e,g., it must start with a letter.
 func (q *QMP) ExecuteBlockdevAdd(ctx context.Context, device, blockdevID string) error {
-	args := map[string]interface{}{
-		"options": map[string]interface{}{
+	var args map[string]interface{}
+
+	if q.version.Minor <= 8 {
```
markdryan: Shouldn't we check the major version number as well?

qemu/qmp.go
```diff
@@ -602,16 +603,30 @@ func (q *QMP) ExecuteQuit(ctx context.Context) error {
...
+		args = map[string]interface{}{
```
markdryan: The only difference between the two maps is the key used for the blockdevID.  It would be better to initialise the map without specifying the blockdevID.  The blockdevID could then be set appropriately in the if statement.  This would save us about 6 lines of code.

markdryan:
>  Do we need to do something to support multiple versions of qemu? Ciao expects to work on the system version of qemu in Ubuntu 16.04 and Fedora 25/26.

Ubuntu 16.04 only has qemu 2.5.  Assuming we continue supporting Ubuntu 16.04, we will need the version check if we want to support new versions of qemu.

qemu/qmp.go
markdryan: Okay, having gone back and read the commit description I see there's another difference, "options".  It should still be possible to factorize the code a little though.

amshinde: @markdryan I have refactored this and pushed.

qemu/qmp.go
amshinde: Yes, I have added a check for major number as well.

qemu/qmp.go
amshinde: Fixed.

qemu/qmp.go
```diff
@@ -602,16 +602,26 @@ func (q *QMP) ExecuteQuit(ctx context.Context) error {
...
+
+	blockdevArgs := map[string]interface{}{
+		"driver": "raw",
+		"file": map[string]interface{}{
+			"driver":   "file",
+			"filename": device,
 		},
 	}
+
+	if q.version.Major == 2 && q.version.Minor >= 9 {
```
markdryan: This should probably be 

if q.version.Major > 2 || (q.version.Major == 2 && q.version.Minor >= 9)

Otherwise, we'll fall back to the old protocol when qemu 3.0 comes out which is probably not what we want.

markdryan: Apart from this the PR looks good.

amshinde: @markdryan Yes you are right,  I have fixed this.

PR #1393, opened 2017-08-10
ciao-controller: store CNCI instance with tenant instances

kaccardi:
Start storing CNCI instances in the datastore, and treat CNCI
as a special tenant instance. This will allow us to use normal
controller commands (such as restart and delete) in the future to
operate on the CNCI.

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -118,20 +118,12 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 		return nil
 	}
 
-	if tenant.CNCIIP == "" {
-		err := c.launchCNCI(tenantID)
-		if err != nil {
-			return err
-		}
-
-		tenant, err = c.ds.GetTenant(tenantID)
-		if err != nil {
-			return err
-		}
-
-		if tenant.CNCIIP == "" {
-			return errors.New("Unable to Launch Tenant CNCI")
-		}
+	if tenant.CNCIID != "" {
+		return nil
+	}
+	err = c.launchCNCI(tenantID)
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/ciao-controller/compute_test.go
+++ b/ciao-controller/compute_test.go
@@ -35,6 +35,21 @@ import (
 	"github.com/01org/ciao/testutil"
 )
 
+// ByTenantID is used to sort CNCI instances by Tenant ID.
+type ByTenantID []types.CiaoCNCI
+
+func (c ByTenantID) Len() int {
+	return len(c)
+}
+
+func (c ByTenantID) Swap(i int, j int) {
+	c[i], c[j] = c[j], c[i]
+}
+
+func (c ByTenantID) Less(i int, j int) bool {
+	return c[i].TenantID < c[j].TenantID
+}
+
 func testHTTPRequest(t *testing.T, method string, URL string, expectedResponse int, data []byte, validToken bool) []byte {
 	req, err := http.NewRequest(method, URL, bytes.NewBuffer(data))
 	if err != nil {
@@ -989,9 +1004,9 @@ func testListCNCIs(t *testing.T, httpExpectedStatus int, validToken bool) {
 		t.Fatal(err)
 	}
 
-	var subnets []types.CiaoCNCISubnet
-
 	for _, cnci := range cncis {
+		var subnets []types.CiaoCNCISubnet
+
 		if cnci.InstanceID == "" {
 			continue
 		}
@@ -1014,6 +1029,8 @@ func testListCNCIs(t *testing.T, httpExpectedStatus int, validToken bool) {
 		)
 	}
 
+	sort.Sort(ByTenantID(expected.CNCIs))
+
 	url := testutil.ComputeURL + "/v2.1/cncis"
 
 	body := testHTTPRequest(t, "GET", url, httpExpectedStatus, nil, validToken)
@@ -1028,6 +1045,8 @@ func testListCNCIs(t *testing.T, httpExpectedStatus int, validToken bool) {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	sort.Sort(ByTenantID(result.CNCIs))
 
 	if reflect.DeepEqual(expected, result) == false {
 		t.Fatalf("expected: \n%+v\n result: \n%+v\n", expected, result)
--- a/ciao-controller/instance.go
+++ b/ciao-controller/instance.go
@@ -17,6 +17,7 @@
 package main
 
 import (
+	"crypto/rand"
 	"encoding/json"
 	"fmt"
 	"net"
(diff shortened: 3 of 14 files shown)


rbradford:
Hi @kaccardi I see that this failed on Travis on TestListCNCIs. I've noticed that this test seems to be flaking on other builds too. I **think** this has started happening since 47b4c8cc0fc2f0fcf4aeb71c3081b67751eb684c

kaccardi:
@rbradford it fails because the array returned is sometimes in a different order. I've put a patch to sort the array prior to returning which should address the issue.

ciao-controller/command.go
```diff
@@ -129,7 +136,12 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 			return err
 		}
 
-		if tenant.CNCIIP == "" {
+		i, err := c.ds.GetInstance(tenant.CNCIID)
```
rbradford: it looks like c.launchCNCI is supposed to guarantee that the datastore contains the instance for the CNCI. As that is happening synchronously it would be better for c.launchCNCI to return that instance rather than require the side effect of pulling it from the datastore.

ciao-controller/instance.go
```diff
@@ -107,14 +108,20 @@ func newInstance(ctl *controller, tenantID string, workload *types.Workload,
...
-	}
+	err = ds.AddInstance(&i.Instance)
 	if err != nil {
 		return errors.Wrapf(err, "Error creating instance in datastore")
 	}
+
+	if i.CNCI == true {
+		t, err := ds.GetTenant(i.TenantID)
+		if err != nil {
+			return errors.Wrap(err, "Error creating instance")
```
rbradford: The error is getting the tenant.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -823,6 +764,16 @@ func (ds *Datastore) GetAllInstancesFromTenant(tenantID string) ([]*types.Instan
 	return nil, nil
 }
 
+// GetAllInstancesFromTenant will retrieve all instances belonging to a specific tenant
```
rbradford: I think the comment should be explicit that this excludes CNCIs.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -832,7 +783,9 @@ func (ds *Datastore) GetAllInstancesByNode(nodeID string) ([]*types.Instance, er
 	n, ok := ds.nodes[nodeID]
 	if ok {
 		for _, val := range n.instances {
-			instances = append(instances, val)
+			if val.CNCI == false {
```
rbradford: Why should the CNCI be excluded from he node listing?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1239,7 +1183,15 @@ func (ds *Datastore) GetInstanceLastStats(nodeID string) types.CiaoServersStats
 		if instance.NodeID != nodeID {
 			continue
 		}
-		serversStats.Servers = append(serversStats.Servers, instance)
+
+		i, err := ds.GetInstance(instance.ID)
+		if err != nil {
+			continue
```
rbradford: Is there an expected situation where this can fail (maybe stats message is received after instance deleted?) Should we at least log this error.

ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -43,6 +45,28 @@ func newTenantHardwareAddr(ip net.IP) (hw net.HardwareAddr) {
 	return
 }
 
+func newHardwareAddr() (net.HardwareAddr, error) {
```
rbradford: This is duplicated. Can you put it somewhere so that it can be reused?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1453,10 +1405,23 @@ func (ds *Datastore) GetTenantCNCISummary(cnci string) ([]types.TenantCNCI, erro
 			continue
 		}
 
+		var IPAddress string
+		var MACAddress string
+
+		if t.CNCIID != "" {
```
rbradford: If t.CNCIID=="" then an empty TenantCNCI will be returned. Is there value in the that? Shouldn't we error out or not include the CNCI in the slice returned?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1439,6 +1454,8 @@ func (ds *Datastore) GetTenantCNCISummary(cnci string) ([]types.TenantCNCI, erro
 
 	ds.tenantsLock.RUnlock()
 
+	sort.Sort(ByTenantID(cncis))
```
kaccardi: ok - I put it there because I thought the api might want it sorted as well, however, probably that's better left to the client now that I think of it.

ciao-controller/internal/datastore/datastore.go
kaccardi: we always have, so if we don't exclude we've changed behavior. The caller of this is an api which is meant to return all the "servers" on a node. This is defined by open stack, and this doesn't include CNCI, which is supposed to be sort of invisible.

ciao-controller/internal/datastore/datastore.go
kaccardi: It is highly unlikely this would ever happen. I don't know if there's much value in logging it, but we can certainly do that.

ciao-controller/internal/datastore/datastore.go
kaccardi: seems better to just continue. I update this.

ciao-controller/instance.go
kaccardi: the result of the err is that there is an error creating an instance. I will make the message more verbose.

ciao-controller/command.go
```diff
@@ -118,7 +118,14 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 		return nil
 	}
 
-	if tenant.CNCIIP == "" {
+	i, err := c.ds.GetInstance(tenant.CNCIID)
+	if err != nil {
+		if tenant.CNCIID != "" {
```
kaccardi: I'm removing a bunch of this error checking - it's not really needed anymore.

ciao-controller/internal/datastore/datastore.go
rbradford: Ack.

ciao-controller/types/types.go
```diff
@@ -714,3 +716,34 @@ type QuotaUpdateRequest struct {
 type QuotaListResponse struct {
 	Quotas []QuotaDetails `json:"quotas"`
 }
+
+func NewTenantHardwareAddr(ip net.IP) net.HardwareAddr {
```
rbradford: I don't think types.go is the right place. Currently only the only "code" in this file for manipulating the types defined in it. I think putting it in the datastore package as datastore/utils.go or putting a utils.go in the ciao-controller main package would be better.

rbradford: Also you need comments of the form // NewTenantHardwareAddr .... (failed on go linting)

ciao-controller/compute_test.go
```diff
@@ -35,6 +35,21 @@ import (
 	"github.com/01org/ciao/testutil"
 )
 
+// ByTenantID is used to sort CNCI instances by Tenant ID.
```
rbradford: github doesn't let you leave comments on the commit message :-( But I think the commit message needs updating to reflect that the sorting is happening in the unit test.

kaccardi: ok.

ciao-controller/types/types.go
kaccardi: ok, will create new "utils" package for ciao-controller.

ciao-controller/instance.go
```diff
@@ -107,14 +108,20 @@ func newInstance(ctl *controller, tenantID string, workload *types.Workload,
 func (i *instance) Add() error {
 	ds := i.ctl.ds
 	var err error
-	if i.CNCI == false {
-		err = ds.AddInstance(&i.Instance)
-	} else {
-		err = ds.AddTenantCNCI(i.TenantID, i.ID, i.MACAddress)
-	}
+	err = ds.AddInstance(&i.Instance)
```
rbradford: Nothing do with the code you've modified but if something later fails we've got an inconsistent state with an instance in the datastore but returned an error.

Issue #1384, opened 2017-07-25
kubicle fails without error message


kaccardi:
kubicle can fail silently with an empty admin.conf and no explanation. If the VMs fail to init properly (due to errors encountered during cloud-init for example), there is no notification to the users.

markdryan:
@kaccardi Is this happening a lot for you?  The only time I've seen it really fail is on k8s 1.7 or when I didn't give the ciao-down VM enough memory.  It needs about 6 GB.

Detecting an error on the master node should be easy enough, and is probably all we need to do, but this is not currently done.  If you do end up with an empty admin.conf file the only thing you can do at the moment is to ssh into the master node, 

```
ssh 198.51.100.2 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i ~/local/testkey
```

and have a look in /var/log/cloud-init-output.log

#1380 should probably go into the kubicle doc as well.  The short term work around for this one is to remove lines [6 to 11)[https://github.com/01org/ciao/blob/master/testutil/ciao-down/workloads/ciao.yaml#L6] in the ciao workload.

kaccardi:
@markdryan I only tried it once, using the memory recommendations in the kubicle tutorial. I deleted the kubernetes cluster and reran kubicle and it worked after that. I'm going to experiment some to see if this is a consistent problem, or if I just got unlucky.

PR #1395, opened 2017-08-11
kubicle: Detect failure to setup master node

markdryan:
This commit makes two changes to kubicle:

1. It modifies the cloud-init file of both the workers and the master
   to retry certain commands that are known to be a little flaky.
   This should improve the success rate of setting up a k8s cluster
   with kubicle.
2. Kubicle now detects whether the commands that setup the master
   node in the cloud-init file fail or not.  If they fail kubicle
   reports the error to the user and downloads the
   cloud-init-output.log file.

Fixes https://github.com/01org/ciao/issues/1384

### DIFF
--- a/k8s/kubicle/create.go
+++ b/k8s/kubicle/create.go
@@ -505,30 +505,35 @@ func (c *creator) cleanup() {
 	}
 }
 
-func startHTTPServer(adminPath string, listener net.Listener, errCh chan error) {
+func writeFile(path string, r *http.Request) error {
+	d, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


k8s/kubicle/create.go
```diff
@@ -505,30 +505,35 @@ func (c *creator) cleanup() {
...
-		err = ioutil.WriteFile(adminPath, d, 0600)
-		if err != nil {
-			return
-		}
+func writeFile(path string, r *http.Request) error {
+	d, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(path, d, 0600)
```
rbradford: I would have used io.CopyBuffer() here.

PR #1396, opened 2017-08-16
qemu: Add disable-modern option for all devices

sboeuf:
For some cases, we need to disable the fast MMIO support, by disabling
virtio 1.0. The reason for this is that we want to be able to nest our
qemu VM inside a VM run by an hypervisor with no support for fast MMIO.

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -202,6 +202,9 @@ type FSDevice struct {
 
 	// SecurityModel is the security model for this filesystem device.
 	SecurityModel SecurityModelType
+
+	// DisableModern prevents qemu from relying on fast MMIO.
+	DisableModern bool
 }
 
 // Valid returns true if the FSDevice structure is valid and complete.
@@ -220,6 +223,9 @@ func (fsdev FSDevice) QemuParams(config *Config) []string {
 	var qemuParams []string
 
 	deviceParams = append(deviceParams, fmt.Sprintf("%s", fsdev.Driver))
+	if fsdev.DisableModern {
+		deviceParams = append(deviceParams, ",disable-modern=true")
+	}
 	deviceParams = append(deviceParams, fmt.Sprintf(",fsdev=%s", fsdev.ID))
 	deviceParams = append(deviceParams, fmt.Sprintf(",mount_tag=%s", fsdev.MountTag))
 
@@ -276,6 +282,9 @@ type CharDevice struct {
 	ID   string
 	Path string
 	Name string
+
+	// DisableModern prevents qemu from relying on fast MMIO.
+	DisableModern bool
 }
 
 // Valid returns true if the CharDevice structure is valid and complete.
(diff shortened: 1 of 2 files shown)


sboeuf:
@kaccardi could you re-trigger the semaphore build for this one please ?

qemu/qemu_test.go
```diff
@@ -89,22 +89,23 @@ func TestAppendEmptyMachine(t *testing.T) {
 	testAppend(machine, "", t)
 }
 
-var deviceNVDIMMString = "-device nvdimm,id=nv0,memdev=mem0 -object memory-backend-file,id=mem0,mem-path=/root,size=65536"
+var deviceNVDIMMString = "-device nvdimm,disable-modern=true,id=nv0,memdev=mem0 -object memory-backend-file,id=mem0,mem-path=/root,size=65536"
```
mcastelino: @sboeuf You do not need disable-modern for NMDIMM. We need it only for virt-io pci devices. It may not even work

qemu/qemu_test.go
```diff
@@ -114,12 +115,13 @@ func TestAppendDeviceFS(t *testing.T) {
...
 		MountTag:      "rootfs",
 		SecurityModel: None,
+		DisableModern: true,
 	}
 
 	testAppend(fsdev, deviceFSString, t)
 }
 
-var deviceNetworkString = "-device virtio-net,netdev=tap0,mac=01:02:de:ad:be:ef -netdev tap,id=tap0,ifname=ceth0,downscript=no,script=no,fds=3:4,vhost=on"
+var deviceNetworkString = "-device virtio-net,disable-modern=true,netdev=tap0,mac=01:02:de:ad:be:ef -netdev tap,id=tap0,ifname=ceth0,downscript=no,script=no,fds=3:4,vhost=on"
```
mcastelino: This will not work either. The disable modern is for the --driver
```
-driver=virtio-net-pci,disable-modern=true,netdev=%s,mac=%s
```

qemu/qemu.go
```diff
@@ -131,6 +134,9 @@ func (object Object) QemuParams(config *Config) []string {
 	var qemuParams []string
 
 	deviceParams = append(deviceParams, string(object.Driver))
+	if object.DisableModern {
```
mcastelino: Maybe I am mistaken. I do not think we can add it blindly. We need to do this on a per device type.

qemu/qemu.go
```diff
@@ -391,6 +412,9 @@ func (netdev NetDevice) QemuParams(config *Config) []string {
 	var qemuParams []string
 
 	deviceParams = append(deviceParams, fmt.Sprintf("%s", netdev.Driver))
+	if netdev.DisableModern {
```
mcastelino: We need to change the way we do network devices.
We need to switch to the new format
``
"driver=virtio-net-pci,netdev=%s,mac=%s
```

qemu/qemu_test.go
sboeuf: Yep it does not, it was just for the unit test. But I can remove it :)

qemu/qemu.go
sboeuf: Ok that sounds good, because Object is too generic, I agree.

sboeuf:
@mcastelino I have reworked the change and all the tests have passed. Please merge it if that's okay for you.

PR #1397, opened 2017-08-16
qemu: Update virtio-net-pci command line

sboeuf:
In case of a network device, and specifically virtio-net-pci, we need
to update to what is expected by qemu. In this case, the driver name
should be prefixed with "driver=".


sboeuf:
@kaccardi sorry to bother you again, but this one is a new PR and Semaphore is hanging... (this is not the same changes, the PR #1396 has passed now).

PR #1399, opened 2017-08-18
Remove gophercloud API use

rbradford:
This is the first stage in decoupling ourselves from our keystone requirement by pushing all HTTP requests through the ciao-cli sendHttpRequest() call where it will be possible to use HTTP basic authentication.

### DIFF
--- a/ciao-cli/main.go
+++ b/ciao-cli/main.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/01org/ciao/ciao-controller/api"
 	"github.com/01org/ciao/ciao-controller/types"
+	"github.com/01org/ciao/openstack/block"
 	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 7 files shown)


ciao-cli/main.go
```diff
@@ -175,6 +176,11 @@ func buildCiaoURL(format string, args ...interface{}) string {
 	return fmt.Sprintf(prefix+format, args...)
 }
 
+func buildBlockURL(format string, args ...interface{}) string {
+	prefix := fmt.Sprintf("https://%s:%d/v2/", *controllerURL, block.APIPort)
+	return fmt.Sprintf(prefix+format, args...)
+}
+
```
kaccardi: I'm not really happy with this - we shouldn't be hardcoding our apis in our client in general, and it instead we should be discovering the URL. For an openstack api, you'd get the service endpoint from keystone, then build it from there. For a ciao api, we'd discover it from a "GET" on "/". However, these old openstack apis are really problematic from the get go, and we know they will eventually be phased out. I'd prefer an intermediate solution where we abstract the URL building from the rest of the client code, but knowing that @tmarcu is creating a new client, I think I'm ok with just doing the simplest fastest thing that gets us where we want to go in the short term. However I'm strongly against this approach as a long term thing.

rbradford:
@tmarcu ping?

PR #1404, opened 2017-08-23
ciao-controller: add a CNCI controller to manage CNCI lifecycle

kaccardi:
This commit adds a CNCI controller to help manage the CNCI lifecyle,
and allows for multiple CNCIs per tenant (1 per subnet).

Create and implement CNCI controller interface which supports starting
a CNCI per tenant subnet. When the subnet is no longer in use (the last
host IP is deleted from this subnet), delete the CNCI after a 5 minute
delay.

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -120,9 +120,29 @@ func (client *ssntpClient) removeInstance(instanceID string) {
 		glog.Warningf("Error when releasing resources for deleted instance: %v", err)
 	}
 	client.deleteEphemeralStorage(instanceID)
+
+	i, err := client.ctl.ds.GetInstance(instanceID)
+	if err != nil {
+		glog.Warningf("Error getting instance from datastore: %v", err)
+		return
+	}
+
 	err = client.ctl.ds.DeleteInstance(instanceID)
 	if err != nil {
 		glog.Warningf("Error deleting instance from datastore: %v", err)
+	}
+
+	if i.CNCI {
+		tenant, err := client.ctl.ds.GetTenant(i.TenantID)
+		if err != nil {
+			glog.Warningf("Error retrieving tenant %v", err)
+			return
+		}
+
+		err = tenant.CNCIctrl.CNCIRemoved(i.ID)
+		if err != nil {
+			glog.Warningf("Error removing CNCI: %v", err)
+		}
 	}
 }
 
@@ -172,10 +192,13 @@ func (client *ssntpClient) concentratorInstanceAdded(payload []byte) {
 		glog.Warningf("Error updating CNCI Info: %v", err)
 	}
 
-	err = client.ctl.ds.CNCIAdded(i.TenantID)
-	if err != nil {
-		glog.Warningf("Error adding CNCI: %v", err)
+	tenant, err := client.ctl.ds.GetTenant(i.TenantID)
+	if err != nil || tenant == nil {
+		glog.Warningf("Error getting tenant: %v", err)
+		return
 	}
+
+	tenant.CNCIctrl.CNCIAdded(newCNCI.InstanceUUID)
 }
 
 func (client *ssntpClient) traceReport(payload []byte) {
@@ -308,9 +331,29 @@ func (client *ssntpClient) startFailure(payload []byte) {
 			glog.Warningf("Error when releasing resources for start failed instance: %v", err)
 		}
 	}
+
+	i, err := client.ctl.ds.GetInstance(failure.InstanceUUID)
+	if err != nil {
+		glog.Warningf("Error getting instance: %v", err)
+		return
+	}
+
+	cnci := i.CNCI
+	tenantID := i.TenantID
+
 	err = client.ctl.ds.StartFailure(failure.InstanceUUID, failure.Reason, failure.Restart)
 	if err != nil {
 		glog.Warningf("Error adding StartFailure to datastore: %v", err)
+	}
+
+	if cnci {
+		tenant, err := client.ctl.ds.GetTenant(tenantID)
+		if err != nil {
+			glog.Warningf("Unable to send start failure event: Error getting tenant %v", err)
+			return
+		}
+
+		tenant.CNCIctrl.StartFailure(failure.InstanceUUID)
 	}
 }
 
@@ -515,6 +558,12 @@ func (client *ssntpClient) RestartInstance(i *types.Instance, w *types.Workload,
 		return errors.Wrapf(err, "Unable to update instance state before restarting")
 	}
 
+	// get the CNCI for this instance
+	cnci, err := t.CNCIctrl.GetInstanceCNCI(i.ID)
+	if err != nil {
+		return err
+	}
+
 	hostname := i.ID
 	if i.Name != "" {
 		hostname = i.Name
@@ -523,11 +572,6 @@ func (client *ssntpClient) RestartInstance(i *types.Instance, w *types.Workload,
 	metaData := userData{
 		UUID:     i.ID,
 		Hostname: hostname,
-	}
-
-	cnci, err := client.ctl.ds.GetInstance(t.CNCIID)
-	if err != nil {
-		return err
 	}
 
 	restartCmd := payloads.StartCmd{
@@ -540,7 +584,7 @@ func (client *ssntpClient) RestartInstance(i *types.Instance, w *types.Workload,
 		Networking: payloads.NetworkResources{
 			VnicMAC:          i.MACAddress,
 			VnicUUID:         i.VnicUUID,
-			ConcentratorUUID: t.CNCIID,
+			ConcentratorUUID: cnci.ID,
 			ConcentratorIP:   cnci.IPAddress,
 			Subnet:           i.Subnet,
 			PrivateIP:        i.IPAddress,
@@ -666,14 +710,15 @@ func (client *ssntpClient) Disconnect() {
 }
 
 func (client *ssntpClient) mapExternalIP(t types.Tenant, m types.MappedIP) error {
-	i, err := client.ctl.ds.GetInstance(t.CNCIID)
+	// get the CNCI for this instance
+	i, err := t.CNCIctrl.GetInstanceCNCI(m.InstanceID)
 	if err != nil {
 		return err
 	}
 
 	payload := payloads.CommandAssignPublicIP{
 		AssignIP: payloads.PublicIPCommand{
-			ConcentratorUUID: t.CNCIID,
+			ConcentratorUUID: i.ID,
 			TenantUUID:       m.TenantID,
 			InstanceUUID:     m.InstanceID,
 			PublicIP:         m.ExternalIP,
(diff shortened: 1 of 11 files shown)


ciao-controller/cnci.go
```diff
@@ -0,0 +1,588 @@
...
+
+// GetInstanceCNCI will return the CNCI Instance for a specific tenant Instance
+func (c *CNCIManager) GetInstanceCNCI(ID string) (*types.Instance, error) {
+	// figure out what subnet we are looking for.
+	instance, err := c.ctrl.ds.GetInstance(ID)
+	if err != nil {
+		return nil, err
+	}
+
+	// convert subnet string to int
```
rbradford: Can you use the function subnetStringToInt() below for this?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,588 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		cnci.subnet = subnetInt
+
+		mgr.cncis[i.ID] = &cnci
+
+		mgr.subnets[subnetInt] = &cnci
+
```
rbradford: nit: ^ bit too much whitespace up here I think.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,588 @@
...
+		if recv != e {
+			return fmt.Errorf("expecting %v got %v", e, recv)
+		}
+		return nil
+	case <-time.After(timeout):
+		return fmt.Errorf("timeout waiting for event %v", e)
+	}
+}
+
+func (c *CNCI) transitionState(to CNCIState) {
```
rbradford: I wonder if long term we should have an event mechanism on types.Instance which would be able to cover this use case but also allow monitoring of activity on more than just CNCI instances.

ciao-controller/types/types.go
```diff
@@ -714,3 +717,17 @@ type QuotaUpdateRequest struct {
 type QuotaListResponse struct {
 	Quotas []QuotaDetails `json:"quotas"`
 }
+
+// CNCIController is the interface for the cnci controller associated with each tenant
+type CNCIController interface {
+	ConcentratorInstanceAdded(ID string) error
```
rbradford: Can we rename these s/Concentrator/CNCI/

ciao-controller/client.go
```diff
@@ -308,10 +331,29 @@ func (client *ssntpClient) startFailure(payload []byte) {
...
+
 	err = client.ctl.ds.StartFailure(failure.InstanceUUID, failure.Reason, failure.Restart)
 	if err != nil {
 		glog.Warningf("Error adding StartFailure to datastore: %v", err)
 	}
+
+	if cnci {
+		tenant, err := client.ctl.ds.GetTenant(tenantID)
+		if err != nil {
+			glog.Warningf("Error adding StartFailure to datastore: %v", err)
```
rbradford: Error getting tenant?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,588 @@
...
+
+type event string
+
+var (
+	added        event = "concentrator added"
+	startFailure event = "cnci start failure"
+	removed      event = "concentrator removed"
+)
+
+var cnciEventTimeout = (5 * time.Minute)
```
rbradford: 5 minutes? This seems quite a long time for a user visible goroutine to be blocked.

ciao-controller/instance.go
```diff
@@ -63,17 +64,23 @@ func isCNCIWorkload(workload *types.Workload) bool {
...
 		if err != nil {
 			return nil, errors.Wrap(err, "error trying to resolve name")
 		}
 
-		if existingID != "" {
-			return nil, fmt.Errorf("Instance name already in use: %s", name)
+		if instance != nil {
+			// allow repeat names on stopping instances
+			instance.StateLock.RLock()
+			if instance.State != payloads.Stopping {
```
rbradford: This looks like some tedious special casing. Is there any other way we can do that?

ciao-controller/internal/datastore/datastore.go
```diff
@@ -1398,50 +1368,33 @@ func (ds *Datastore) addInstanceStats(stats []payloads.InstanceStat, nodeID stri
...
-		for _, subnet := range t.subnets {
-			binary.BigEndian.PutUint16(subnetBytes, (uint16)(subnet))
-			cn.Subnets = append(cn.Subnets, fmt.Sprintf("Subnet 172.%d.%d.0/8", subnetBytes[0], subnetBytes[1]))
-		}
+		cn.Subnets = append(cn.Subnets, i.Subnet)
 
 		cncis = append(cncis, cn)
 
-		if cnci != "" && cnci == t.CNCIID {
+		if cnci != "" {
```
rbradford: rename cnci to cnciID and then you can rename cn to cnci to be consistent

ciao-controller/cnci.go
kaccardi: yes, I have been thinking that too - I think in general it might be nice to have a manager for each instance - we can migrate the cnci manager to this eventually.

ciao-controller/cnci.go
kaccardi: @rbradford true - I was worried on a busy/large cluster it might take some time to hear back from launcher.

ciao-controller/cnci.go
kaccardi: yes.

ciao-controller/instance.go
kaccardi: @rbradford I don't think it's appropriate to do the check inside ResolveInstance, and so the only other thing we could do is make some kind of random name for the cnci. However, I think it's perfectly fine to have duplicate names when an instance is about to stop anyway.

rbradford: I like the idea of generating a pseudorandom name for the cnci if possible.

kaccardi: isn't it nicer though to have a name with some meaning?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+		return fmt.Errorf("timeout waiting for event %v", e)
+	}
+}
+
+func (c *CNCI) transitionState(to CNCIState) {
+	transitionInstanceState(c.instance, (string(to)))
+
+	// some state changes cause events
+	c.eventChLock.Lock()
+	ch := c.eventCh
```
markdryan: Is there a reason to use a pointer for eventCh?  nil is a valid value for channels.

The locks are needed I guess as a state transition might not be always be initiated by one of the methods in CNCIManager.  Is this correct?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+		eventCh:     &ch,
+		eventChLock: sync.Mutex{},
+		subnet:      subnet,
+	}
+
+	// we initialized the eventCh because we are going to wait for
+	// an event. Close and delete at the conclusion of this function.
+	defer func() {
+		cnci.eventChLock.Lock()
+		close(ch)
```
markdryan: No need to close the channel as far as I can tell.  By the time we get here there will be nobody reading from it.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+		cnci.eventCh = nil
+		cnci.eventChLock.Unlock()
+	}()
+
+	c.cnciLock.Unlock()
+
+	err = waitForEventTimeout(ch, removed, cnciEventTimeout)
```
markdryan: You could just have a return here.

ciao-controller/cnci.go
markdryan: Okay so maybe I'm wrong.  If we have two parallel calls to waitForTimeout we need the close.

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+			count++
+		}
+	}
+
+	return count, nil
+}
+
+// Shutdown cleans up a CNCIManager in anticipation of a shutdown.
+func (c *CNCIManager) Shutdown() {
+	// the only thing we need to do right now at shutdown time
```
markdryan: Are we missing a lock here?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+		}
+	}
+}
+
+func newCNCIManager(ctrl *controller, tenant string) (*CNCIManager, error) {
+	mgr := CNCIManager{
+		tenant: tenant,
+		ctrl:   ctrl,
+
+		cnciLock: &sync.RWMutex{},
```
markdryan: Is there a reason to use a pointer here?

ciao-controller/cnci.go
```diff
@@ -0,0 +1,579 @@
...
+		return nil, err
+	}
+
+	// you need to see if this cnci instance is actually needed
+	// anymore.
+
+	for _, i := range instances {
+		cnci := CNCI{
+			ctrl:        ctrl,
+			eventChLock: sync.Mutex{},
```
markdryan: Shouldn't need to explicitly initialise this although I guess it doesn't hurt.

ciao-controller/client.go
```diff
@@ -172,10 +192,13 @@ func (client *ssntpClient) concentratorInstanceAdded(payload []byte) {
 		glog.Warningf("Error updating CNCI Info: %v", err)
 	}
 
-	err = client.ctl.ds.CNCIAdded(i.TenantID)
-	if err != nil {
-		glog.Warningf("Error adding CNCI: %v", err)
+	tenant, err := client.ctl.ds.GetTenant(i.TenantID)
+	if err != nil || tenant == nil {
+		glog.Warningf("Error getting tenant: %v", err)
```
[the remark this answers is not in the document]
kaccardi: without the tenant I've got no way to find a channel to signal success/failure. So, after the timeout waiting for the event, the user will get an error indicating their was a problem. We would still have a CNCI active though for this tenant. So, the next time the user tried to launch anything no cnci would actually be launched again, and everything would seem normal. Except of course that if this error occurred, it means there's something seriously wrong with the datastore or controller and likely everything has crashed and burned.

ciao-controller/cnci.go
kaccardi: no, I thought I needed them because multiple go routines within cnci.go might modify the eventCh variable value. For example, a new channel is created when you are waiting to remove a subnet, as well as when you are waiting to add one. Can I remove the locks?

kaccardi: maybe its sufficient to just use the cnciLock lock.

markdryan: No I think you're correct.  You need the locks as you're creating and deleting channels.

ciao-controller/instance.go
rbradford: could still have that "cnci-<tenant-ID>-<subnet>-<random suffix>" or something

kaccardi: hah! after all my worrying about the name making sense, the cnci name isn't user visible anyway....
will just make some random thing.

ciao-controller/cnci.go
kaccardi: fixed.

ciao-controller/cnci.go
kaccardi: no.

kaccardi:
This should be ready for merge I think....

PR #1405, opened 2017-08-24
ciao-controller: Only setup openstack identity handler once

rbradford:
In 10a927a379e4834866b32b7d6580107f89959aef an inefficiency was
introduced where HTTP handlers would get wrapped in the openstack
identity handler multiple times. This change removes that excessive
wrapping by expanding the list of services acceptable for the ciao
end point to be the union of all the others allowing all the others to
be removed.


rbradford:
@kaccardi I found I needed to do this anyway and so might as well get it in.

PR #1410, opened 2017-08-30
Use client certificates for authentication

rbradford:
This PR enables authentication without Keystone by using client certificates inside singlevm and for physical clusters including passing BAT. 

In order to create new users "ciao-deploy auth create" is used. If tenant IDs are provided with the username then that user certificate is granted access to those tenants. If no tenant ID is provided a new tenant ID will be created and it will be created in controller on first use.


openstack/image/api.go
```diff
@@ -505,11 +505,15 @@ func Routes(config APIConfig, r *mux.Router) *mux.Router {
 		r = mux.NewRouter()
 	}
 
+	r.Handle("/v2/{tenant}/images", APIHandler{context, createImage}).Methods("POST")
+	r.Handle("/v2/{tenant}/images/{image_id:"+uuid.UUIDRegex+"}/file", APIHandler{context, uploadImage}).Methods("PUT")
+	r.Handle("/v2/{tenant}/images", APIHandler{context, listImages}).Methods("GET")
+	r.Handle("/v2/{tenant}/images/{image_id:"+uuid.UUIDRegex+"}", APIHandler{context, getImage}).Methods("GET")
+	r.Handle("/v2/{tenant}/images/{image_id:"+uuid.UUIDRegex+"}", APIHandler{context, deleteImage}).Methods("DELETE")
 	r.Handle("/v2/images", APIHandler{context, createImage}).Methods("POST")
```
kaccardi: oh god - I hate this. Now into our fake open stack compatibility we add fake openstack-ish but still incompatible endpoints. when does it stop???

is it not possible to get the tenant ID from the cert and then populate the tenant context var and use that?

rbradford: @kaccardi Yeh it sucks. Every other endpoint whether they be openstack based or our own have the tenant embedded in the URL as the mechanism for the controller to know what tenant the user wants to work on. As users can have access to multiple tenants (or now the admin can work on a tenant's behalf) you need to know what tenant the action is for.

The image API previously extracted the tenant from the keystone token. However without that and no tenant embedded in the url we have no way to know what tenant the action should be performed on.

Some thoughts:

a.) We could do as a do I here and embed the tenant in the URL like we have in all our other APIs.
b.) We could put the tenant ID inside a HTTP header (but then for all the other entry points we'd have duplication of the same information.)
c.) Mandate that there is only one tenant per certificate however then we would lack the feature of allowing admin to work on any tenant's images which is something they can do for all other resource types.

My preference is a.) and then (longer term) auditing all our APIs so that they're sensible!

ciao-cli/main.go
```diff
@@ -329,72 +326,104 @@ func limitToString(limit int) string {
...
 
-	if *identityURL == "" {
-		fatal += "Missing required identity URL\n"
+	for {
+		p, data = pem.Decode(data)
+		if p == nil {
+			break
+		}
+		if p.Type == "CERTIFICATE" {
+			certBlock = p
```
markdryan: Are we missing a break here, or are we looking for the last cert in the file, assuming there can be more than one?

markdryan:
Great work by the way.  ./setup.sh now takes just 45 seconds for me.

ciao-cli/main.go
rbradford: There will only be one PEM CERTIFICATE block in the file. If there are multiple x509 certificates (because there is a chain with an intermediate CA) they are still only encoded into a single PEM block.

However for safety i'll modify this to error out if p != nil when we try and make the assignment thus enforcing that behaviour.

rbradford:
@markdryan Updated now to remove openstack-client from ciao-down ciao workloads.

markdryan:
@kaccardi This is good for me.  Are you okay to merge?

openstack/image/api.go
kaccardi: ok - a) for now and then making better APIs long term.

PR #1413, opened 2017-08-30
Make ciao work with Go 1.9 and enable Go 1.9 in ciao ciao-down workloads

markdryan:
This PR fixes the slow CNCI startup times and verify.sh failures seen when compiling ciao with Go 1.9. Changes in Go 1.9 meant that the cnci_agent was unable to connect to the scheduler for up to 1 minute after it had started.  The agent's TLS session was blocking waiting for /dev/urandom's entropy pool to be initialsed.  Inside a newly created VM booting for the first time this could take up to a minute, even when using virtio-rng-pci.  The issue is fixed by specifying that the cnci_agent's ssntp's tls session should read from /dev/urandom directly, rather than calling getrandom.

There are a couple of other fixes here including enabling SSNTP logging in the scheduler and increasing the ping time out in verify.sh for the CNCI.

virtio-rng-pci is also enabled for ciao-down

### DIFF
--- a/ciao-scheduler/scheduler.go
+++ b/ciao-scheduler/scheduler.go
@@ -1248,6 +1248,7 @@ func configSchedulerServer() (sched *ssntpSchedulerServer) {
 		CAcert:    *cacert,
 		Cert:      *cert,
 		ConfigURI: *configURI,
+		Log:       ssntp.Log,
 	}
 
 	setSSNTPForwardRules(sched)
--- a/networking/ciao-cnci-agent/client.go
+++ b/networking/ciao-cnci-agent/client.go
@@ -19,6 +19,7 @@ package main
 import (
 	"encoding/json"
 	"flag"
+	"io"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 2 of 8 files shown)


markdryan:
This PR is a little rushed but it fixes the issue for me.

I'll need to submit a new PR that prevents virtio-rng-pci stealing all of ciao-down's host's entropy.

I decided to put the random number generator in the cnci_agent instead of somewhere more generic as we shouldn't encourage its use.

@mcastelino The ping timeout increase in verify.sh I think is probably only a work around.  I've seen instances where it takes the CNCI a really long time, up to a minute to configure the firewall after a reboot.  I don't know why this is so slow.  It doesn't happen all the time.

mcastelino:
> I'll need to submit a new PR that prevents virtio-rng-pci stealing all of ciao-down's host's entropy.

@markdryan how do you plan to solve this?

rbradford:
@mcastelino  https://wiki.qemu.org/index.php/Features/VirtIORNG

> Optional parameters to limit the rate of data sent to the guest are inlucded:
> -device virtio-rng-pci,max-bytes=1024,period=1000

rbradford:
@mcastelino Let's open a bug about specialisation for singlevm vs physical case and not hold up this PR.

Issue #1414, opened 2017-08-31
verify.sh fails with CNCI ping timeout after restart


kaccardi:
FATAL ERROR exiting:  Unable to ping CNCI after restart 1

kaccardi:
This happened within ciao-down.

kaccardi:
Increasing the ping timeout has no effect - the CNCI is not responding to network traffic.

kaccardi:
you can duplicate this outside of verify.sh by starting a single instance, finding the cnci ip with ciao-cli, then ssh'ing to the cnci and telling it to reboot now. it never returns.

kaccardi:
git revert -m 1 730416be70d7031bf18e6730d66f30453c3691a2 works.

rbradford:
@kaccardi I've seen that test flake often before we switched to go 1.9. After the switch to 1.9 @markdryan saw it consistently. After 0a6a18d630070bddada10572ea2a481345cf5e26 it *feels* like the flake happens less frequently. However I did see it on a couple of semaphore builds yesterday.

e.g. Here is a example of it happening before 1.9: https://semaphoreci.com/kaccardi/ciao/branches/pull-request-1410/builds/10

rbradford:
Another datapoint on this. When I connect to the console of the CNCI and then trigger a reboot i see it shutdown, then the "screen" changes colour (I think that's from the OVMF firmware) and then it doesn't come back.

If I reboot the Ubuntu VM I do see it come back when connected to the console via nc.

rbradford:
So I think wasn't being patient enough with the CNCI reboot. Sometimes it can take a long time but it does eventually happen.

This reboot in particular took over two minutes:

```
 Sep 01 15:12:11 cnci-a0833e20f6294e2a9a5c09d3feacec7d-b00addb9 systemd-journald[137]: Journal stopped
 -- Reboot --
 Sep 01 15:14:23 cnci-a0833e20f6294e2a9a5c09d3feacec7d-b00addb9 systemd-journald[146]: Runtime journal (/run/log/journal/) is currently using
  724.0K.
```

rbradford:
By contrast this one only took 11s!

```
Sep 01 15:43:54 cnci-a0833e20f6294e2a9a5c09d3feacec7d-b00addb9 systemd-journald[146]: Journal stopped
-- Reboot --
[...]

PR #1415, opened 2017-09-01
Revert "ciao-launcher: Expose host /dev/random to instances"

rbradford:
Reverts 01org/ciao#1409

My observations are that with this this change our CNCI reboot time varies substantially, from 9s to over 2 minutes. With this change reverted it's sub 10s.

Fixes: #1414

### DIFF
--- a/ciao-launcher/qemu.go
+++ b/ciao-launcher/qemu.go
@@ -336,7 +336,6 @@ func generateQEMULaunchParams(cfg *vmConfig, isoPath, instanceDir string,
 	qmpSocket := path.Join(instanceDir, "socket")
 	qmpParam := fmt.Sprintf("unix:%s,server,nowait", qmpSocket)
 	params = append(params, "-qmp", qmpParam)
-	params = append(params, "-device", "virtio-rng-pci")
 
 	if cfg.Mem > 0 {
 		memoryParam := fmt.Sprintf("%d", cfg.Mem)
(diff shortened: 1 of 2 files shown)


kaccardi:
we were unable to prove that this wasn't a coincidence, so probably don't merge till we spend more time looking at this.

markdryan:
Also, if we revert this should we also revert https://github.com/01org/ciao/commit/e0335e0af4b13155bfbec3a7346155ac40b1e913?

rbradford:
@markdryan Not necessarily. The ciao-down boot time is not so critical as the CNCI start time.

markdryan:
I can't make verify.sh fail on my setup with #1409 reverted.  I've tried 15 times.  With #1409 merged I get a failure rate of 1 in 2.  So for me, #1409 does seem to be responsible.  The thing is that #1409's changes are probably needed on a real cluster.  We'd need to test on a real cluster to see whether it causes any issues with verify.sh.  In the short term though it's probably best to revert this change.

markdryan:
> @markdryan Not necessarily. The ciao-down boot time is not so critical as the CNCI start time.

Agreed. I haven't seen any negative affects on ciao-down every since I added this.

markdryan:
Seeing as how the original bug is also addressed by https://github.com/01org/ciao/commit/30ddabb9e201a7985100750e64172ae4b518d1e6 (and so this patch is no longer needed to make ciao work with Go 1.9) and this patch does seem to be causing the issue for both @rbradford and myself, I'm going to merge this PR.

PR #1418, opened 2017-09-05
Restart cnci and allows failed CNCIs to be restarted

markdryan:
Here's a first attempt at implementing migration of the CNCIs.  It's pretty basic but it does improve on what we currently  have.  The PR makes the following changes.

1. The CNCI manager will now restart a CNCI instance that exits or crashes.
2. It is now possible to restart failed CNCIs.

### DIFF
--- a/ciao-controller/client.go
+++ b/ciao-controller/client.go
@@ -163,10 +163,30 @@ func (client *ssntpClient) instanceStopped(payload []byte) {
 		glog.Warning("Error unmarshalling InstanceStopped: %v")
 		return
 	}
-	glog.Infof("Stopped instance %s", event.InstanceStopped.InstanceUUID)
-	err = client.ctl.ds.InstanceStopped(event.InstanceStopped.InstanceUUID)
+	instanceID := event.InstanceStopped.InstanceUUID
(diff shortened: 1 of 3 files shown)


markdryan:
This was really a learning PR to help me understand what CNCI migration actually means.  The changes I've made are just a stepping stone but are useful so I've submitted what I have.  There's a bunch of stuff missing.

1. CNCI rootfs volumes are leaked when the CNCI fails to start (and also perhaps when they're deleted).  Maybe we should just make them ephemeral.  This shouldn't prevent them from being migrated.

2.  There's no generic solution for specifying restart policies in workloads yet.

3. We need to decide what happens if an attempt to restart a CNCI fails.  Maybe this ties into the policy mentioned in step 2.  I think right now the CNCI will be deleted.  It's not clear to me what that means for a tenant.  The CNCI will get restarted the next time one is needed by a command issued by the user but the old CNCI state will be lost.  Is this a scenario we actually want to support and if so how do we handle it?

4. Try to figure out whether any of this can be tested by BAT.

markdryan:
I have a follow on commit that removes instance.Attachments.  I'll submit this in a separate PR.

Issue #1295, opened 2017-05-31
Drop elliptic choice from ssntp cert


rbradford:
Currently we conditionally support RSA vs ECDSA. It would be nice to just use one as it would simplify our code paths (in particular in the new deployment code.)

```
func generatePrivateKey(ell bool) (interface{}, error) {
	if ell == false {
		return rsa.GenerateKey(rand.Reader, 2048)
	}
	return ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
}
```

@sameo @tpepper thoughts?

PR #1420, opened 2017-09-05
ciao-cert: Drop option to choose elliptic curve certs

rbradford:
Having the ability to choose different certificate types makes the
ability to programmatic creation of certificates, e.g. ciao-deploy
problematic as the type of signing key must be known to to manage these
keys. Instead remove the option to simplify the certificate workflow.

Fixes: #1295


rbradford:
Also, I should add ECDSA based cryptography is much slower especially with the key size we're using.

Issue #1202, opened 2017-02-23
Volumes for CNCI are reported via quota system and user visible


rbradford:
Currently the ephemeral volume used for the CNCI is tracked in the quota system appearing as a resource consumed and will also be checked.

It also appears in ciao-cli volume list. We have agreed that the resources used for the CNCI should be outside the quota. As the CNCI instance is also invisible to the user I feel that it's volume should also not be visible.

Find some way to not track the CNCI volume and hide it in ciao-cli volume list

PR #1421, opened 2017-09-05
Hide CNCI volume from user and quotas

rbradford:
Hide the CNCI volume from the output of "ciao-cli volume list" and also make sure that it is not included in the quotas for the user. Do this by establishing an Internal field on types.BlockData that is populated from types.StorageResource (which is statically specified for the CNCI) which can be tested in conditional statements.

Fixes: #1202

### DIFF
--- a/ciao-controller/instance.go
+++ b/ciao-controller/instance.go
@@ -218,6 +218,7 @@ func addBlockDevice(c *controller, tenant string, instanceID string, device stor
 		TenantID:    tenant,
 		Name:        fmt.Sprintf("Storage for instance: %s", instanceID),
 		Description: s.Tag,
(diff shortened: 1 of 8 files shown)


ciao-controller/instance.go
```diff
@@ -218,6 +218,7 @@ func addBlockDevice(c *controller, tenant string, instanceID string, device stor
 		TenantID:    tenant,
 		Name:        fmt.Sprintf("Storage for instance: %s", instanceID),
 		Description: s.Tag,
+		Internal:    s.Internal,
```
rbradford: @markdryan FYI this is where you want to set Ephemeral: true.

Issue #235, opened 2016-06-09
testing: add -race to .travis.yml and test-cases


tpepper:
We can't simply add -race to .travis.yml as test-cases must pass it through.

And if we add it, we surely have some currently undetected data races.  So travis would fail and nothing would merge for a while.

I will experiment with this, find and fix issues, get us to a clean state, then enable the test support and submit a PR on a patchset.

amyleeland:
Tried and failed here, moving this to backlog to revisit

tpepper:
Issue #868 is an instance of a -race detected issue.

PR #1422, opened 2017-09-05
travis: Enable race detection for unit tests

rbradford:
We've made quite a few improvements to the unit tests so we should now
try and run with race detection turned on.

Fixes: #235

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ go:
     - 1.9
     - tip
 
-env:
-    - COVERALLS_TOKEN=mwTn1pOFqEOUT13vylZNHq53NanoMznO7
 matrix:
   allow_failures:
   - go: tip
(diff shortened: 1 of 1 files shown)


markdryan:
I'm a bit worried about this one as it became a merge blocker in the past.  I'll pull this to my fork and test my fork builds with race enabled for a few days.  If there are no issues I'll merge, otherwise I'll enter and fix any issues I find.

rbradford:
@markdryan Agree that just one clean built is not sufficient. One thought I had was to use travis's matrix to add -race versions of say 1.8 & 1.9 and then mark those as allowable failures.

The other thing to note is that -race approximately doubles our travis cycle time (to around ~12 minutes now.)

rbradford:
@markdryan updated to make -race an allowed failure. Notice that I also remove the coveralls token. The goveralls documentation says this is not necessary for public repos: https://github.com/mattn/goveralls

Coveralls not updating s not related to this PR, it just seems to be having issues.

Issue #660, opened 2016-10-11
ciao-cli: node show -node-id doesn't work.


kaccardi:
$ ./ciao-cli node show -node-id 40c75c6b-9fc9-4711-808d-887c6acd5081
usage: ciao-cli [options] node show

Show info about a node

The show flags are:
  -cnci
        Show info about a cnci node
  -node-id string
        Node ID

I would have expected instead of the above to see information about the node.

PR #1423, opened 2017-09-05
ciao-cli node show fix

rbradford:
Make "ciao-cli node show" work when given a node-id and also fix the error message for when --cnci is provided but no --node-id is given.

Fixes: #660

### DIFF
--- a/ciao-cli/node.go
+++ b/ciao-cli/node.go
@@ -334,7 +334,7 @@ func showNode(cmd *nodeShowCommand) error {
 
 func showCNCINode(cmd *nodeShowCommand) error {
 	if cmd.nodeID == "" {
-		fatalf("Missing required -cnci-id parameter")
+		fatalf("Missing required -node-id parameter")
 	}
 
 	var cnci types.CiaoCNCI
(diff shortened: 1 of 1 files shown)


ciao-cli/node.go
```diff
@@ -285,14 +285,59 @@ func (cmd *nodeShowCommand) run(args []string) error {
...
+	var nodes types.CiaoNodes
+	err = unmarshalHTTPResponse(resp, &nodes)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	var node *types.CiaoNode
+	for i := range nodes.Nodes {
+		if nodes.Nodes[i].ID == cmd.nodeID {
+			node = &nodes.Nodes[i]
```
markdryan: We should break here.

ciao-cli/node.go
```diff
@@ -285,14 +285,59 @@ func (cmd *nodeShowCommand) run(args []string) error {
...
+	if node == nil {
+		fatalf("Node not found: %s", cmd.nodeID)
+	}
+
+	if cmd.template != "" {
+		return templateutils.OutputToTemplate(os.Stdout, "node-show", cmd.template,
+			node, nil)
+	}
+
+	fmt.Printf("\tUUID: %s\n", node.ID)
```
markdryan: This code is duplicated with dumpNode.  We could factorize.

rbradford:
@markdryan feedback addressed.

PR #1428, opened 2017-09-12
qemu: Add support for memory pre-allocation

mcastelino:
Add support for pre-allocating all of the RAM.
This increases the memory footprint of QEMU and should be used
only when needed.


sameo:
LGTM

PR #1430, opened 2017-09-13
QEMU: Enable realtime options

mcastelino:
Enable realtime options in QEMU. Also add support to control memory
locking. Turning realtime on with memory locking disabled allows
memory to be swapped out, potentially increasing density of VMs.

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -744,6 +744,13 @@ type Knobs struct {
 
 	// MemPrealloc will allocate all the RAM upfront
 	MemPrealloc bool
+
+	// Mlock will control locking of memory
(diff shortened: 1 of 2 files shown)


markdryan:
The new unit test seems to be failing

```
Logs for qemu.TestAppendKnobsAllTrue
	qemu_test.go:71: Failed to append parameters [-no-user-config -nodefaults -nographic -daemonize -realtime mlock=on] != [-no-user-config -nodefaults -nographic -daemonize]
```

mcastelino:
@markdryan the unit test should be fixed now. Once it is done can you merge.

/cc @sameo

Issue #1407, opened 2017-08-25
make subnet size configurable


kaccardi:
Add some config option to change the size of the tenant subnet.

kaccardi:
The host will still be represented by one byte, however, the subnet size will dictate how many hosts we will allocate before we need to move to the next subnet.

PR #1431, opened 2017-09-13
make subnet bits configurable

kaccardi:
This PR makes the number of bits in the subnet mask a configurable attribute of the tenant. It also allows the tenant name to be configurable.

Fixes: #1407

### DIFF
--- a/ciao-cli/tenant.go
+++ b/ciao-cli/tenant.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/01org/ciao/ciao-controller/api"
 	"github.com/01org/ciao/ciao-controller/types"
+	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/intel/tfortools"
 )
 
@@ -109,6 +110,8 @@ func putCiaoTenantConfig(ID string, name string, bits int) error {
 
 	err = unmarshalHTTPResponse(resp, &config)
 
+	oldconfig := config
+
 	if name != "" {
 		config.Name = name
 	}
@@ -117,14 +120,26 @@ func putCiaoTenantConfig(ID string, name string, bits int) error {
 		config.SubnetBits = bits
 	}
 
+	a, err := json.Marshal(oldconfig)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
 	b, err := json.Marshal(config)
 	if err != nil {
 		fatalf(err.Error())
 	}
 
-	body := bytes.NewReader(b)
+	merge, err := jsonpatch.CreateMergePatch(a, b)
+	if err != nil {
+		fatalf(err.Error())
+	}
 
-	_, err = sendCiaoRequest("PUT", url, nil, body, api.TenantsV1)
+	fmt.Println(string(merge))
+
+	body := bytes.NewReader(merge)
+
+	_, err = sendCiaoRequest("PATCH", url, nil, body, "merge-patch+json")
 
 	return err
 }
--- a/ciao-cli/tenant.go
+++ b/ciao-cli/tenant.go
@@ -17,15 +17,117 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"os"
 	"text/template"
 	"time"
 
+	"github.com/01org/ciao/ciao-controller/api"
 	"github.com/01org/ciao/ciao-controller/types"
 	"github.com/intel/tfortools"
 )
+
+func getCiaoTenantsResource() (string, error) {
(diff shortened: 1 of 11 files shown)


ciao-cli/tenant.go
```diff
@@ -17,16 +17,118 @@
...
+		name:  "id",
+		value: ID,
+	}
+
+	url, err := getCiaoTenantsResource()
+	if err != nil {
+		return "", err
+	}
+
+	if !checkPrivilege() {
```
markdryan: Shouldn't this check go first?  Is there any point in calling getCiaoTenantsResource() if the user does not have the correct rights.  Also, we should return and error here.  This code will always return nil for the error.

markdryan: Maybe this check shouldn't even be here.  Wouldn't it prevent an unprivileged user from doing a 

ciao-cli tenant list --config

ciao-cli/tenant.go
```diff
@@ -210,3 +388,61 @@ func listTenantResources(t *template.Template) error {
 
 	return nil
 }
+
+func listTenantConfig(t *template.Template, tenantID string) error {
+	if tenantID == "" {
+		fatalf("Missing required -for-tenant parameter")
```
markdryan: or --tenant-id when the user is not privileged.

ciao-cli/tenant.go
```diff
@@ -210,3 +388,61 @@ func listTenantResources(t *template.Template) error {
...
+	}
+
+	resp, err := sendCiaoRequest("GET", url, nil, nil, api.TenantsV1)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	err = unmarshalHTTPResponse(resp, &tenants)
+	if err != nil {
+		return err
```
markdryan: Any reason not to use fatalf here?  I guess it doesn't really make much difference and this code is all going to have to change but it's a little inconsistent.

ciao-cli/tenant.go
```diff
@@ -17,16 +17,118 @@
...
+
+func putCiaoTenantConfig(ID string, name string, bits int) error {
+	var config types.TenantConfig
+
+	url, err := getCiaoTenantRef(ID)
+	if err != nil {
+		return err
+	}
+
+	resp, err := sendCiaoRequest("GET", url, nil, nil, api.TenantsV1)
```
markdryan: This is a little racy.  If two clients update the tenant information at the same time we could have unexpected results.

If a client does 

ciao-cli tenant update --subnebits

he would expect to just update the subnetbits and not the tenant name.

But if someone else updated the name after this call but before the second call our user makes to sendCiaoRequest, he'd end up inadvertently overwriting their change.

ciao-cli/tenant.go
kaccardi: @markdryan an unprivileged user is allowed to view their own config (but not modify it). This function is used for that purpose, as well as for privileged users. The check ensures that for a non privilege user the ref it gets is for a non-privileged endpoint, while for a privileged user the ref is for the privileged endpoint.

kaccardi: The non-privileged user is prevented from doing ciao-cli tenant list --config for anyone except their own ID both programatically in ciao cli, and also in controller via a cert check. 
        if cmd.config {
                if checkPrivilege() == false {
                        return listTenantConfig(t, *tenantID)
                }

                return listTenantConfig(t, cmd.tenantID)
        }

ciao-cli/tenant.go
kaccardi: I'll change that.

ciao-cli/tenant.go
kaccardi: I'll update the error message, thanks.

ciao-cli/tenant.go
markdryan: Okay, I think I see now.  If if the user isn't privileged we actually want to return the url we got from getCiaoTenantsResource.

ciao-cli/tenant.go
kaccardi: @markdryan you are right. This is due to the fact that a PUT is supposed to completely replace the resource. I could change this verb to be a PATCH and that would allow you to only update a single attribute if you wanted to.

kaccardi:
@markdryan I think your feedback is addressed. Added issue for missing bat tests: https://github.com/01org/ciao/issues/1435

kaccardi:
we also need to vendor github.com/evanphx/json-patch at some point.

markdryan:
Tested on the real cluster and everything seems okay.  I'll merge this now and then submit a separate patch to vendor json-patch.

PR #1432, opened 2017-09-14
Fix a number issues with --hard-reset

markdryan:
--hard-reset in launcher was not working very well for a number of reasons.

1. There was a long standing bug about the fact that it could hang.
2. It didn't wipe the docker plugin DB.  This caused problems with ciao-deploy installed clusters.
3. It was unable to cleanly destroy docker containers as the docker plugin DB initialization code was broken.
4. It was unable to cleanly destroy docker containers as it was not passed information about the management and compute nets.

All of these issues are addressed by this PR.  Issues 3 and 4 meant that running a ./cleanup.sh on singlevm was really slow if there were some container instances up and running.

### DIFF
--- a/networking/libsnnet/docker_plugin.go
+++ b/networking/libsnnet/docker_plugin.go
@@ -171,7 +171,7 @@ func (d *DockerEpMap) NewTable() {
 
 //Name provides the name of the map
 func (d *DockerEpMap) Name() string {
-	return tableNetworkMap
+	return tableEndPointMap
 }
 
 //NewElement allocates and returns an endpoint value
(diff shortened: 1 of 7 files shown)


networking/libsnnet/docker_plugin.go
```diff
@@ -171,7 +171,7 @@ func (d *DockerEpMap) NewTable() {
 
 //Name provides the name of the map
 func (d *DockerEpMap) Name() string {
-	return tableNetworkMap
+	return tableEndPointMap
```
mcastelino: wow.. that was quite a big bug at my end. I wonder how things worked in spite of this bug.

testutil/singlevm/cleanup.sh
```diff
@@ -2,14 +2,16 @@
 
 . ~/local/demo.sh
 
+ciao_vlan_ip=198.51.100.1
+ciao_vlan_subnet=${ciao_vlan_ip}/24
```
mcastelino: @markdryan should these move to demo.sh just for consistency?

PR #1433, opened 2017-09-14
ciao-down: Add timeout to connect

tmarcu:
Instead of looping forever, add a 30 second timeout to the connect call and
quit with a nicer error message. In cases where the VM simply has not been
started, it could me misleading to continually print waiting to boot.

Signed-off-by: Tudor Marcu <tudor.marcu@intel.com>

### DIFF
--- a/testutil/ciao-down/ciao_down.go
+++ b/testutil/ciao-down/ciao_down.go
@@ -460,6 +460,7 @@ func status(ctx context.Context, errCh chan error) {
 }
 
 func connect(ctx context.Context, errCh chan error) {
+	timeout := time.NewTimer(time.Second * 30)
 	ws, err := prepareEnv(ctx)
 	if err != nil {
 		errCh <- err
@@ -494,6 +495,9 @@ func connect(ctx context.Context, errCh chan error) {
 			case <-ctx.Done():
 				errCh <- fmt.Errorf("Cancelled")
 				return
+			case <-timeout.C:
+				fmt.Println("\nCannot connect to VM, please make sure it has been started:\n\t$ ciao-down start")
+				break DONE
 			}
 
 			if sshReady(ctx, sshPort) {


tmarcu:
RFC - Instead of trying to figure out the real status of the VM and what's happening with it, just add a timeout so the user isn't stuck waiting forever.
Is 30 seconds reasonable? I'm not sure what other kinds of errors might be holding up connecting, so it may or may not be.

testutil/ciao-down/ciao_down.go
```diff
@@ -494,6 +495,9 @@ func connect(ctx context.Context, errCh chan error) {
 			case <-ctx.Done():
 				errCh <- fmt.Errorf("Cancelled")
 				return
+			case <-timeout.C:
+				fmt.Println("\nCannot connect to VM, please make sure it has been started:\n\t$ ciao-down start")
+				break DONE
```
markdryan: This will still try to ssh into the instance which is unlikely to succeed.  We should probably just return here sending an error down the errCh.

testutil/ciao-down/ciao_down.go
```diff
@@ -460,6 +460,7 @@ func status(ctx context.Context, errCh chan error) {
 }
 
 func connect(ctx context.Context, errCh chan error) {
+	timeout := time.NewTimer(time.Second * 30)
```
markdryan: You could use the existing context to do the time out, e.g.

```
ctx, cancel = context.WithTimeout(ctx, time.Second * 30)
defer cancel()
```

Then you don't need to modify the select statement at all.  If you wanted to distinguish between a cancel and a timeout you could compare ctx.Error() to context. DeadlineExceeded

testutil/ciao-down/ciao_down.go
tmarcu: AH, I meant to return at this point... you're right it will still try to ssh in, and if you put the timeout low enough it can hit the case where it prints out "Cannot connect" and then ssh's in successfully.

testutil/ciao-down/ciao_down.go
tmarcu: I like this idea.

PR #1434, opened 2017-09-15
Knobs: Modify the behaviour of the Mlock knob.

mcastelino:
The Mlock knob is unfortunately tied to realtime. Now enable Mlock knob to implicitly enable realtime to get the desired swapping behavior.

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -1015,9 +1015,20 @@ func (config *Config) appendKnobs() {
 
 	if config.Knobs.Realtime == true {
 		config.qemuParams = append(config.qemuParams, "-realtime")
+		// This path is redundant as the default behaviour is locked memory
+		// Realtime today does not control any other feature even though
+		// other features may be added in the future
+		// https://lists.gnu.org/archive/html/qemu-devel/2012-12/msg03330.html
 		if config.Knobs.Mlock == true {
 			config.qemuParams = append(config.qemuParams, "mlock=on")
 		} else {
+			config.qemuParams = append(config.qemuParams, "mlock=off")
+		}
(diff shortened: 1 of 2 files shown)


qemu/qemu_test.go
```diff
@@ -240,6 +239,7 @@ func TestAppendKnobsAllTrue(t *testing.T) {
 }
 
 func TestAppendKnobsAllFalse(t *testing.T) {
+	var knobsString = "-realtime mlock=off"
```
kaccardi: so you cannot explicitly turn off mlock without the -realtime option if I am understanding my quick google of the option. It appears that -realitme on it's own does nothing, but if you don't explicitly do -mlock=off, the default is mlock=on, which you want to make configurable. Right?

qemu/qemu.go
```diff
@@ -1020,6 +1020,13 @@ func (config *Config) appendKnobs() {
 		} else {
 			config.qemuParams = append(config.qemuParams, "mlock=off")
 		}
+	} else {
+		// QEMU always locks memory anyway if realtime is not defined
+		if config.Knobs.Mlock == false {
+			//Enable realtime anyway just to get the right swapping behaviour
```
kaccardi: its a nit, but I would word this comment differently. I would say "in order to turn mlock off we need the -realtime option as well"

qemu/qemu_test.go
mcastelino: Agree. Today it does nothing else.. I will reword accordingly

mcastelino:
@kaccardi reworded both the commit and comments int the code. Hopefully this will make things clearer.

Issue #1435, opened 2017-09-15
Add BAT tests for tenants list/show/update


kaccardi:
There are no BAT tests for tenants list/show/update. Make some.

PR #1438, opened 2017-09-20
Add tenant create/delete and BAT tests

kaccardi:
Add tenant create and delete support. Tenant create will create a new tenant with a given config. Tenant delete will remove a tenant and all it's associated objects, including instances, workloads, images, storage, and external IPs.

add some BAT tests to exercise our new tenant APIs. 

Fixes: #1435

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -107,7 +107,13 @@ func (c *controller) confirmTenantRaw(tenantID string) error {
 		return nil
 	}
 
-	tenant, err = c.ds.AddTenant(tenantID)
+	// if we are adding tenant this way, we need to use defaults
+	config := types.TenantConfig{
+		Name:       "",
+		SubnetBits: 24,
+	}
+
+	tenant, err = c.ds.AddTenant(tenantID, config)
 	if err != nil {
 		return err
 	}
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/ciao-project/ciao/ciao-controller/types"
 	"github.com/ciao-project/ciao/payloads"
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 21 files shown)


ciao-controller/command.go
```diff
@@ -73,6 +74,41 @@ func (c *controller) stopInstance(instanceID string) error {
...
+	}
+
+	err = c.deleteInstance(instanceID)
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		i.StateChange.L.Lock()
+		for i.StateLock.RLock(); i.State != payloads.Deleted && i.State != payloads.Hung; {
```
markdryan: I'm not sure if this loop is safe.  If i.State == payload.Deleted when the loop is first entered, i.StateLock will get locked, but we'll never enter the loop body and so it will not get unlocked.

markdryan: Also, if you go around the loop twice iStateLock will not be locked on the second iteration.

markdryan: The same comments applies to i.StateChange.  It may never get unlocked if we don't enter the loop and it may be unlocked when we call Wait

ciao-controller/tenants.go
```diff
@@ -69,3 +72,155 @@ func (c *controller) PatchTenant(tenantID string, patch []byte) error {
...
+	instances, err := c.ds.GetAllInstancesFromTenant(tenantID)
+	if err != nil {
+		return errors.Wrap(err, "Unable to remove tenant")
+	}
+
+	for _, i := range instances {
+		err := c.deleteInstanceSync(i.ID)
+		if err != nil {
+			// remove directly.
+			c.client.RemoveInstance(i.ID)
```
markdryan: Do we have a test that checks that controller gracefully handles InstanceDeleted events that are received after it has decided and that an instance has hung and has removed it directly, i.e., that are received 2 minutes after the delete request was sent?

ciao-controller/tenants.go
```diff
@@ -69,3 +72,155 @@ func (c *controller) PatchTenant(tenantID string, patch []byte) error {
...
+	}
+
+	// delete all this tenant's instances.
+	instances, err := c.ds.GetAllInstancesFromTenant(tenantID)
+	if err != nil {
+		return errors.Wrap(err, "Unable to remove tenant")
+	}
+
+	for _, i := range instances {
+		err := c.deleteInstanceSync(i.ID)
```
markdryan: You might want to do this in parallel.  Otherwise in a worst case scenario, a node hosting 100 instances from the tenant you're deleting drops off the network, it will take 3 hours and 20 minutes, to delete all the instances.

ciao-controller/tenants.go
kaccardi: no test - although I did review the code and confirm that it will work just fine.

rbradford:
lgtm. I'd like to see the fixup commits merged in to the appropriate places.

PR #1441, opened 2017-09-22
README: Update badges

markdryan:
This commit updates the badges in the README.md file to reflect the change
of the project repo.  It also removes the waffle badge as we aren't currently
using waffle.


markdryan:
The coverage badge isn't working yet but at least the URL is correct.

PR #1443, opened 2017-09-22
Docs: Remove out of date or duplicated documentation

markdryan:
This PR removes documentation that does really serve any real purpose,
is out of date or is duplicated elsewhere such as in godoc or on
ciao-project.github.io.

The content of the remaining README.md files will need to be either
migrated elsewhere, deleted or perhaps be given leave to remain if
there is no other suitable place for these files.


markdryan:
This is just the start.  There are still some README,md files left in the repo, but there content is useful and so will need to be migrated elsewhere before they are deleted.

PR #1449, opened 2017-09-25
QEMU: Knobs: Huge Page Support: Add support for huge pages

mcastelino:
Add support to launch virtual machines where the RAM is
allocated using huge pages. This is useful for running
with a user mode networking stack, and for custom setups
which require high performance and low latency.


qemu/qemu.go
```diff
@@ -771,6 +771,9 @@ type Knobs struct {
 	// MemPrealloc will allocate all the RAM upfront
 	MemPrealloc bool
 
+	// HugePages will pre-allocate all the RAM from huge pages
+	HugePages bool
```
markdryan: Maybe we should add a note here, and to MemPrealloc, to say that if you set HugePages you should also set Memory.Size otherwise setting HugePages will have no effect.

mcastelino: @markdryan will do. Also do we even work if the caller does not set Memory.Size? Do we choose a default anywhere. Maybe I missed that.

mcastelino:
@markdryan addressed your review comments.

qemu/qemu.go
mcastelino: Nevermind.. we allow QEMU to pick the default. Added a comment, that calls that out.

qemu/qemu.go
```diff
@@ -1025,7 +1028,25 @@ func (config *Config) appendKnobs() {
 		config.qemuParams = append(config.qemuParams, "-daemonize")
 	}
 
-	if config.Knobs.MemPrealloc == true {
+	// Both HugePages and MemPrealloc require the memory size of the VM
```
markdryan: @mcastelino Sorry to be pedantic, but it would be better to have the comment in the Knobs structure as this will then appear in the godoc.

https://godoc.org/github.com/01org/ciao/qemu#Knobs

mcastelino: @markdryan that should be addressed now. Showing up in godoc is required for the externally visible options.

qemu/qemu.go
```diff
@@ -768,6 +768,17 @@ type Knobs struct {
 	// Daemonize will turn the qemu process into a daemon
 	Daemonize bool
 
+	// Both HugePages and MemPrealloc require the Memory.Size of the VM
+	// to be set, as they need to reserve the memory upfront in order
+	// for the VM to boot without errors.
+
```
markdryan: You need to delete these new lines.  Otherwise this information won't show up in godoc.  These comments are not seen to be associated with any fields.

qemu/qemu.go
```diff
@@ -768,6 +768,17 @@ type Knobs struct {
...
+	// HugePages always results in memory pre-allocation.
+	// However the setup is different from normal pre-allocation.
+	// Hence HugePages has precedence over MemPrealloc
+
```
markdryan: This newline needs to go as well.

mcastelino: @markdryan should be fixed. Sorry for the thrash.

PR #1450, opened 2017-09-26
Remove use of gophercloud

rbradford:
Switch ciao-cli volume and ciao-cli image over to using structures from ciao's API implementation. Also remove ciao-cli volume update command as that uses an endpoint that does not exist. This PR also removes gophercloud from the vendored dependencies.


rbradford:
@kaccardi  or @markdryan  bye-bye gophercloud \o/

Issue #1379, opened 2017-07-21
test-cases doesn't seem to run example tests


markdryan:
Which isn't great.  I need to take a look and see why.  For example

```
cd $GOPATH/github.com/01org/ciao/templateutils

$ test-cases
Package       Test Case                 Time Taken Result 
templateutils TestTemplateGenerateUsage 0.00s      PASS   
templateutils TestOptions               0.00s      PASS   
templateutils TestOptAllFns             0.00s      PASS   
templateutils TestBadCols               0.00s      PASS   
```

Only a fraction of the unit tests are being run as the example test cases are being skipped.

PR #1453, opened 2017-09-26
test-cases: Run Example tests

rbradford:
If there are Example...() tests defined run them in test-cases.

Fixes: #1379


rbradford:
@markdryan good bug!

rbradford:
@markdryan thoughts on what to do? go test won't run those tests due to the requirements you told me about today (e.g. the comment.)

markdryan:
Yeah, I saw this.  We'd need to look at the go test code to figure out how it decides not to run them.  Or we could do a hack an ignore NOT RUN errors for examples.

rbradford:
@markdryan I addressed this by using the go/doc package as a per go test.

PR #1454, opened 2017-09-27
sqlite3db: Use sqlite implicit transactions

rbradford:
From the sqlite documentation:

"No changes can be made to the database except within a transaction. Any
command that changes the database (basically, any SQL command other than
SELECT) will automatically start a transaction if one is not already in
effect. Automatically started transactions are committed when the last
query finishes."

As such we can simplify our database code by taking advantage of
automatic transactions. This will also make it easier when we start
working towards being "errcheck" clean.

This change also switches to using defer for releasing locks in places
where that was not yet happening. Any local variables named datastore
were also renamed to db to disambiguate with the general datatstore.


rbradford:
This takes errcheck in ciao-controller/internal/datastore from 98 -> 50. Still a lot of work left though.

rbradford:
@kaccardi fewer lines deleted than i'd have hoped!

Issue #1446, opened 2017-09-25
CNCI volumes are leaked when the CNCI is deleted


markdryan:
Essentially, CNCIs boot volume should be made ephemeral.

PR #1456, opened 2017-09-27
ciao-controller: Make CNCI volume ephemeral

rbradford:
When a CNCI is deleted the volume should not be leaked.

Fixes: #1446


rbradford:
Tested by starting instances, seeing rbd volume created for cnci. And then delete all instances and after the time delay observe that the cnci rbd volume has been cleaned up.

PR #1457, opened 2017-09-27
Remove transient database

rbradford:
Remove the transient database to resolve issues that present themselves when controller gets restarted.


rbradford:
@kaccardi rebased. The issue that I thought was caused by this, #1466, is also reproducible without this change so I don't think fixing that should be intertwined with landing this PR.

Issue #1459, opened 2017-09-27
ciao-deploy can't handle certs with fqdn


kaccardi:
ciao-deploy uses go's os.Hostname() to get the hostname to use in ciao-cli to contact controller. However, if the controller web cert was generated using fqdn, this won't work as os.Hostname() does not return fqdn.

PR #1465, opened 2017-09-28
ciao-deploy: Allow overriding name controller is reachable with

rbradford:
When the HTTPS certificate uses a FQDN it is necessary to let the user of ciao-deploy specify the name the controller's certificate specifies so that the invocations of ciao-cli will work.

Fixes: #1459


rbradford:
@kaccardi let me know if this works for you

PR #1471, opened 2017-09-28
Set DNS server for containers

markdryan:
ciao-launcher: Set DNS server for containers

This commit ensures that docker containers can access the DNS server
running on their CNCI.  We do this to ensure that containers will
be able to access other ciao instances by their names or UUIDs.


markdryan:
@mcastelino Could I ask you to take a look at this PR to see if I'm on the right track.

I'm trying to set the DNS servers correctly inside the container so that containers can do dns lookups of instances in the same tenant that have been launched with the --name parameter.

I set the DNS here

https://github.com/ciao-project/ciao/pull/1471/files#diff-439ae54b464d7ab403580b5a608ccc2cR214

And I get it from here

https://github.com/ciao-project/ciao/pull/1471/files#diff-bcd152c48dceec92adecd6e0a49dfcd3R287

If this looks okay to you I'll tidy things up and resubmit

mcastelino:
LGTM. This makes sense. There is a DNS resolver running in the CNCI on each bridge.

ciao-launcher/qemu.go
```diff
@@ -108,7 +108,7 @@ func (q *qemuV) ensureBackingImage() error {
 	return nil
 }
 
-func (q *qemuV) createImage(bridge string, userData, metaData []byte) error {
+func (q *qemuV) createImage(bridge, gatewayIP string, userData, metaData []byte) error {
```
mcastelino: @markdryan Why are we doing this for VMs. Today we push that automatically via DHCP. Also for a CNCI VM, we do not want this pushed in.

mcastelino: It should not cause issues, but I am worried about potential conficts between this and what we serve via DHCP

markdryan: We're not.  The value is ignored in this file.  It has to be passed however, otherwise qemuV will not implement the virtualizer interface.

mcastelino:
LGTM

ciao-launcher/qemu.go
markdryan: Maybe the virtualizer interface needs to be tidied up a bit, but this a current restriction.  All virtualizers get passed all information, whether they need it to create their specific instance types or not.

Issue #1467, opened 2017-09-28
Panic when controller idle


rbradford:
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x46a8f2]

goroutine 452 [running]:
sync.(*RWMutex).Lock(0x0)
	/home/rob/opt/go/src/sync/rwmutex.go:93 +0x22
main.transitionInstanceState(0xc420196000, 0x9d3dee, 0x8, 0x0, 0x0)
	/home/rob/go/src/github.com/ciao-project/ciao/ciao-controller/instance.go:183 +0x5f
main.(*CNCI).stop(0xc4201f5170, 0xc4201f5140, 0x1000)
	/home/rob/go/src/github.com/ciao-project/ciao/ciao-controller/cnci.go:77 +0x47
main.(*CNCIManager).RemoveSubnet(0xc4201c6d80, 0x1000, 0x0, 0x0)
	/home/rob/go/src/github.com/ciao-project/ciao/ciao-controller/cnci.go:299 +0x17c
main.(*CNCIManager).ScheduleRemoveSubnet.func1()
	/home/rob/go/src/github.com/ciao-project/ciao/ciao-controller/cnci.go:272 +0x97
created by time.goFunc
	/home/rob/opt/go/src/time/sleep.go:170 +0x44
```

Probably related to #1466 but different codepath.

PR #1473, opened 2017-09-28
ciao-controller: do not use a pointer for instance state lock

kaccardi:
Change the instance StateLock to be a struct rather than a pointer.
This way it is initialized at Instance creation.

Fixes: #1467


rbradford:
Tested and resolves problem for me!

Issue #1472, opened 2017-09-28
use ciao configuration file to get ceph_id for BAT tests


kaccardi:
If instead of hardcoding the ceph ID for the bat tests we read it from configuration.yaml, we could run bat tests in places where the ceph_id wasn't "ciao" and was whatever the user setup.

PR #1474, opened 2017-09-28
ceph_bat: Use system configuration file to determine cephID

kaccardi:
If we use the ciao cluster configuration file to retrieve the cephID,
then we don't have to hardcode it.

Fixes: #1472

### DIFF
--- a/_release/bat/storage_bat/ceph_test.go
+++ b/_release/bat/storage_bat/ceph_test.go
@@ -15,15 +15,26 @@
 package storagebat
 
 import (
+	"fmt"
+	"io/ioutil"
 	"os"
 	"testing"
 
 	"github.com/ciao-project/ciao/bat"
 	"github.com/ciao-project/ciao/ciao-storage"
+	"gopkg.in/yaml.v2"
 )
 
-var driver = storage.CephDriver{
-	ID: "ciao",
+var driver storage.CephDriver
+
+type storageConfig struct {
+	CephID string `yaml:"ceph_id"`
+}
+
+type configuration struct {
+	Configure struct {
+		Storage storageConfig `yaml:"storage"`
+	} `yaml:"configure"`
 }
 
 // Check creating a ceph backed block device works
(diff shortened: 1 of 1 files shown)


_release/bat/storage_bat/ceph_test.go
```diff
@@ -105,3 +116,23 @@ func TestCopyBlockDevice(t *testing.T) {
...
+	data, err := ioutil.ReadFile("/etc/ciao/configuration.yaml")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading config file: %v", err)
+		os.Exit(1)
+	}
+
+	var config configuration
+	err = yaml.Unmarshal(data, &config)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading config file: %v", err)
```
rbradford: nit: s/reading/parsing/

_release/bat/storage_bat/ceph_test.go
```diff
@@ -105,3 +116,23 @@ func TestCopyBlockDevice(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestMain(m *testing.M) {
+	data, err := ioutil.ReadFile("/etc/ciao/configuration.yaml")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading config file: %v", err)
+		os.Exit(1)
+	}
```
markdryan: What happens if we exit here?  Are all the tests in this package reported as failed?  If so it might be better not to exit but just not to initialise driver.ID or rather leave it set to "".  This way each test could check to see if driver.ID == "" and if it is called t.Skip().  This should give us a clean run when BAT is not being run on the scheduler node.

kaccardi:
@rbradford @markdryan please take another look when you get a chance.

_release/bat/storage_bat/ceph_test.go
```diff
@@ -105,3 +128,21 @@ func TestCopyBlockDevice(t *testing.T) {
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading config file: %v", err)
+	} else {
+		err = yaml.Unmarshal(data, &config)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error parsing config file: %v", err)
+		}
+	}
+
+	driver.ID = config.Configure.Storage.CephID
```
markdryan: I would only initialise driver.ID if yaml.Unmarshal succeeds.  If it fails, the contents of driver.ID are essentially undefined so we could end up running the tests.

Issue #1315, opened 2017-06-15
Remove live detatching of volumes from VM instances?


markdryan:
ciao currently supports live attaching and detaching of volumes.  This issue questions whether we should keep the live detaching of volumes feature.  The reasons to remove it being:

1. It only really works well on Fedora.  Support on Ubuntu 16.04 is flaky.  Live detaching sometimes fails and needs to be retried.  This flakyiness can lead to BAT test failures as this feature is tested by BAT.  Live detach doesn't work at all on Clearlinux.
2. The QMP command we use to implement this feature is experimental.
3. We don't have any real use case for this feature right now.
4. It could lead to data loss if the VM is writing to the volume when it is detached by controller.
5. The feature adds complexity to launcher and BAT which is currently hard to justify.

We could remove live detach but keep live attach and detach from stopped volumes.

markdryan:
@kaccardi What do you think about this?

kaccardi:
ok, let's get rid of it. If someone needs this we can reimplement it.

rbradford:
@markdryan wants to address this himself as there is some cleanup that could be in in this part of the code.

rbradford:
@markdryan do you think we should support detach on exited state (i'm thinking not as it's just going to create complexity..)

PR #1486, opened 2017-09-29
Disable live detach in controller

rbradford:
Reject requests to detach volumes from instances that are not exited as live detach is flaky due to relying on guest OS support. Update the BAT tests to stop the instances prior to trying to detach.

Fixes: #1315

### DIFF
--- a/ciao-controller/openstack_volume.go
+++ b/ciao-controller/openstack_volume.go
@@ -280,19 +280,21 @@ func (c *controller) DetachVolume(tenant string, volume string, attachment strin
 			continue
 		}
 
+		i.StateLock.RLock()
+		state := i.State
+		i.StateLock.RUnlock()
+
+		if state != payloads.Exited {
+			retval = errors.New("Can only detach from exited instances")
+			continue
+		}
+
 		// update volume state to detaching
-		info.State = types.Detaching
+		info.State = types.Available
 
 		err = c.ds.UpdateBlockDevice(info)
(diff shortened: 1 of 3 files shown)


rbradford:
@markdryan the unit tests pass now, perhaps you'd like to review? I have a follow-on PR that removes detach from ssntp event handling, etc.

ciao-controller/openstack_volume.go
```diff
@@ -288,11 +280,19 @@ func (c *controller) DetachVolume(tenant string, volume string, attachment strin
 			continue
 		}
 
-		// send command to attach volume.
-		err = c.client.detachVolume(a.BlockID, a.InstanceID, i.NodeID)
+		i.StateLock.RLock()
```
markdryan: It would be nice to have a BAT test for this.  Does ciao-cli return an error if you try to detach a volume from a running instance?

ciao-controller/openstack_volume.go
```diff
@@ -288,11 +280,19 @@ func (c *controller) DetachVolume(tenant string, volume string, attachment strin
...
+		if i.State != payloads.Exited {
+			retval = errors.New("Can only detach from exited instances")
+			continue
```
markdryan: Won't this deadlock?

rbradford: @markdryan great catch, fixed.

PR #1489, opened 2017-09-29
ciao-deploy: make ciao-launcher --hard-reset work

kaccardi:
This PR has 3 changes:
* output the combined stdout/stderr of ciao-launcher --hard-reset when it fails in addition to the exit code.
* output the combined stdout/stderr of rmdir /etc/pki/ciao - if this fails it prevents launcher from being reset.
* use an absolute path for ciao-launcher, as "sudo ciao-launcher --hard-reset" assumes /usr/local/bin/ciao-launcher is in your path (and it isn't always)


rbradford:
The path fix for sudo is great and i've pulled it into a separate PR ready for merging. For the output i'd rather generalise this so that all commands we invoke via SSH get's logged to a buffer than can be dumped to the filesystem on error (handling both join and unjoin). I've started working on that but now so i'd rather not merge this current PR.

kaccardi:
@rbradford  sounds good!

PR #1498, opened 2017-10-02
[RFC] New Ciao CLI WIP

tmarcu:
Rebased onto master and broke up the massive spew of changes into something more manageable to read. Separate PR's will be made for the subsequent to-be-added commands, as this one only has support for showing workload(s)  as a simple initial change.


tmarcu:
https://github.com/tmarcu/ciao/pull/3

markdryan:
@tmarcu Push the latest version of your code to this PR and I'll review it this evening/tomorrow.  Sorry about the delay.

tmarcu:
@markdryan no problems, I just didn't want this PR to start growing endlessly since you have to review it :)

tmarcu:
With the recent api changes and removals, this needs major rework.

rbradford:
@tmarcu yes unfortunately the de-openstackification is going to clash with the work you're doing. Perhaps I could suggest you focus just on functionality covered by the "ciao API": workloads, (now) images, tenants, and external IPs/pools (see ciao-controller/api/api.go) for a list. Those are going to change a lot less than the other APIs.

tmarcu:
I'll focus on that, and I'll try to keep up with the new commits as they get introduced so it doesn't hit so many conflicts hopefully.

ciao-sdk/helpers.go
```diff
@@ -0,0 +1,485 @@
...
+		}
+	}
+
+	if certBlock == nil {
+		return nil, errors.New("No certificate block block in cert file")
+	}
+
+	cert, err := x509.ParseCertificate(certBlock.Bytes)
+	if err != nil {
+		return nil, errors.New("Unable to parse x509 certificate data")
```
markdryan: Take a look at the errors package https://github.com/pkg/errors.  This allows you to wrap errors providing a complete context for the error message as it makes its way up the call stack.  We've sort of agreed to use this package in new code and this looks like a good candidate.

ciao-sdk/helpers.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	return cert.Subject.Organization, nil
+}
+
+func prepareWithClientCert() {
+	cert, err := tls.LoadX509KeyPair(*clientCertFile, *clientCertFile)
+	if err != nil {
+		fatalf("Unable to load client certiticate: %s", err)
```
markdryan: Don't call fatal, return an error.

ciao-sdk/helpers.go
```diff
@@ -0,0 +1,485 @@
...
+			fmt.Println("Tenants available:")
+			for i := range tenants {
+				fmt.Println(tenants[i])
+			}
+			fatalf("Multiple tenants available. Please specify one with -tenant-id")
+		}
+
+		*tenantID = tenants[0]
+	}
+
```
markdryan: Extra new line here.

ciao-sdk/show.go
```diff
@@ -0,0 +1,459 @@
...
+	"net/http"
+)
+
+type byCreated []compute.ServerDetails
+
+func (ss byCreated) Len() int           { return len(ss) }
+func (ss byCreated) Swap(i, j int)      { ss[i], ss[j] = ss[j], ss[i] }
+func (ss byCreated) Less(i, j int) bool { return ss[i].Created.Before(ss[j].Created) }
+
+func listEvent(cmd *cobra.Command, args []string) error {
```
markdryan: I know we've agreed to create generic commands like show and delete that can apply to a range of objects.  However, from a code organisation point of view I think it would make more sense to keep the code grouped by object type rather than by action.  So node.go would have all the node functionality, instance.go the instance functionality, just as before.  show.go would then just contain the Show function that calls into all the other modules.

markdryan: Also, I think the sdk package is currently mixing up two things.

1. It contains the code to formulate requests and parse responses for REST APIs
2. It contains cli code for building a cli.

I think these two things need to be separate.  If I was going to create my own tool to issue commands to the ciao cluster I probably wouldn't want any of the CLI stuff.  So I think this pacakge needs to be split into two.  Keep the REST stuff in here and move the cli stuff into a package under ciao/ciao

ciao-sdk/show.go
```diff
@@ -0,0 +1,459 @@
...
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	err = unmarshalHTTPResponse(resp, &servers)
+	if err != nil {
+		fatalf(err.Error())
+	}
+
+	for i, server := range servers.Servers {
```
markdryan: One thing @rbradford and I have discussed in the past is using tfortools to do all the formatting.

You could replace this whole loop with

tfortools.OutputToTemplate(os.Stdout, "servers", '{{table .}}', nil)

It would also make the code easier for all our list and show functons as there would only be one code path.  If the template string is not defined it would default to a template that displays things in the default format, otherwise you use the user supplied template.  But in all cases you output via a template.

markdryan: Actually, why is there no template support here?

ciao-sdk/show.go
```diff
@@ -0,0 +1,459 @@
...
+		sortedServers = append(sortedServers, v)
+	}
+	sort.Sort(byCreated(sortedServers))
+
+	if Template != "" {
+		return tfortools.OutputToTemplate(os.Stdout, "instance-list", Template,
+			&sortedServers, nil)
+	}
+
+	w := new(tabwriter.Writer)
```
markdryan: The rest of this function is one line with a template.

ciao/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
markdryan: This file shouldn't exist.  I realise it may have been added by cobra but we should remove it as we already have a license one level up.

ciao-sdk/show.go
markdryan: Actually, it already looks like you have a package for 2, ciao/cmd.

ciao-sdk/helpers.go
```diff
@@ -0,0 +1,485 @@
...
+func errorf(format string, args ...interface{}) {
+	glog.ErrorDepth(1, fmt.Sprintf("ciao ERROR: "+format, args...))
+}
+
+func fatalf(format string, args ...interface{}) {
+	glog.FatalDepth(1, fmt.Sprintf("ciao FATAL: "+format, args...))
+}
+
+var (
+	tenantID       = new(string)
```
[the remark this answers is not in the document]
tmarcu: It's deref'd all over the code already, I'll change it to be a regular variable.

PR #1507, opened 2017-10-04
ciao-deploy: log SSH output for join/unjoin

rbradford:
Log the SSH activity for join/unjoin to a debug log that is written to disk if anything fails.

### DIFF
--- a/ciao-deploy/deploy/nodes.go
+++ b/ciao-deploy/deploy/nodes.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"sync"
@@ -176,9 +177,13 @@ func SetupNodes(ctx context.Context, sshUser string, networkNode bool, hosts []s
 	for _, host := range hosts {
 		wg.Add(1)
 		go func(hostname string) {
+			var debugBuffer bytes.Buffer
+			ctx = WithDebugLogWriter(ctx, &debugBuffer)
 			err := setupNode(ctx, anchorCertPath, caCertPath, hostname, sshUser, networkNode)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error setting up node: %s: %v\n", hostname, err)
+				fmt.Fprintf(os.Stderr, "Output for %s:\n", hostname)
+				_, _ = io.Copy(os.Stderr, &debugBuffer)
 			}
 			wg.Done()
 		}(host)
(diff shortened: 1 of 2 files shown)


ciao-deploy/cmd/utils.go
```diff
@@ -38,3 +41,17 @@ func getSignalContext() (context.Context, context.CancelFunc) {
...
+	f, err := ioutil.TempFile("", "ciao-deploy-debug-log")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error creating debug log file: %v\n", err)
+	} else {
+		_, err = io.Copy(f, r)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error writing debug log\n")
+		} else {
+			fmt.Fprintf(os.Stderr, "Debug log at: %s\n", f.Name())
+		}
```
markdryan: We're missing an f.Close() here.

ciao-deploy/cmd/join.go
```diff
@@ -30,10 +31,15 @@ func join(args []string) int {
 	ctx, cancelFunc := getSignalContext()
 	defer cancelFunc()
 
+	var debugLog bytes.Buffer
+
+	ctx = deploy.WithDebugLogWriter(ctx, &debugLog)
```
markdryan: bytes.Buffer is not thread-safe.  The fact that you're passing it to a context suggests that the buffer might be used by different go routines concurrently.  Is this the case?  If so, I think you're going to need some locking.

ciao-deploy/deploy/nodes.go
```diff
@@ -173,17 +173,22 @@ func SetupNodes(ctx context.Context, sshUser string, networkNode bool, hosts []s
...
 
 	var wg sync.WaitGroup
+	var failedNodes int
 	for _, host := range hosts {
 		wg.Add(1)
 		go func(hostname string) {
 			err := setupNode(ctx, anchorCertPath, caCertPath, hostname, sshUser, networkNode)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error setting up node: %s: %v\n", hostname, err)
+				failedNodes++
```
markdryan: This is racy.  You could use an atomic int I or ditch the wait group and use a chan of ints.  The go routines write a 1 or 0 to the channel indicating whether they succeeded or failed and you replace w.Wait with a for loop that ranges over the channel until it's read len(hosts) numbers.

ciao-deploy/deploy/nodes.go
```diff
@@ -251,16 +256,21 @@ func teardownNode(ctx context.Context, hostname string, sshUser string) error {
...
 func TeardownNodes(ctx context.Context, sshUser string, hosts []string) error {
 	var wg sync.WaitGroup
+	var failedNodes int
 	for _, host := range hosts {
 		wg.Add(1)
 		go func(hostname string) {
 			err := teardownNode(ctx, hostname, sshUser)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error tearing down node: %s: %v\n", hostname, err)
+				failedNodes++
```
markdryan: Racy as well.

Issue #1497, opened 2017-10-02
subnet bits range check is wrong


rbradford:
The RFC1918 block we use inside ciao for tenant communication is limited to 20 bits. We should make sure that subnet range is limited to between 20 bits and 30

PR #1509, opened 2017-10-04
Restrict subnet-bits value to a range between 12-30

kaccardi:
Due to the private IP address block that is used for ciao tenant
networks, the subnet bits value must be a value between 12 and 30.

Fixes: #1497 

### DIFF
--- a/ciao-controller/tenants.go
+++ b/ciao-controller/tenants.go
@@ -81,12 +81,12 @@ func (c *controller) CreateTenant(tenantID string, config types.TenantConfig) (t
 		return types.TenantSummary{}, err
 	}
 
-	// SubnetBits must be between 4 and 30
+	// SubnetBits must be between 12 and 30
 	if config.SubnetBits == 0 {
 		config.SubnetBits = 24
 	} else {
-		if config.SubnetBits < 4 || config.SubnetBits > 30 {
-			return types.TenantSummary{}, errors.New("subnet bits must be between 4 and 30")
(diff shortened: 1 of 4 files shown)


rbradford:
@kaccardi looks like you need to tweak the BAT test

kaccardi:
maybe a stupid question, but after I doubled checked the allowed subnet mask bits for this 172.16 private network, it seems like you could go down to 12. Can you double check your assertion that it should be between 20-30?

rbradford:
@kaccardi depends how you define it. If subnet bits is the size of the mask then yes, that 12 bits. If subnet bits is the number of machines then it's 2**20.

You were correct that I got the numbers confused. Either way 4 was the wrong lower bound.

I agree with @mcastelino on cleaning up the nomenclature to describe the mask size as that is what people are used to from CIDR notation.

kaccardi:
@rbradford @mcastelino I went with "cidr-prefix-size". Hope that is clear enough.

ciao-controller/tenants.go
```diff
@@ -85,8 +85,8 @@ func (c *controller) CreateTenant(tenantID string, config types.TenantConfig) (t
 	if config.SubnetBits == 0 {
```
markdryan: The comment on the link above should be changed as well, e.g.,

// SubnetBits must be between 4 and 30

kaccardi:
@markdryan I updated the comment and rebased.

PR #1510, opened 2017-10-05
Networking: Add support for handling macvtap interfaces

mcastelino:
Add support for macvtap interfaces. This also brings in support
for generic multiqueue support in virt containers.

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -332,16 +332,70 @@ func (cdev CharDevice) QemuParams(config *Config) []string {
 	return qemuParams
 }
 
-// NetDeviceType is a qemu networing device type.
+// NetDeviceType is a qemu networking device type.
 type NetDeviceType string
 
 const (
 	// TAP is a TAP networking device type.
 	TAP NetDeviceType = "tap"
 
-	// MACVTAP is a MAC virtual TAP networking device type.
+	// MACVTAP is a macvtap networking device type.
 	MACVTAP = "macvtap"
+
+	// IPVTAP is a ipvtap virtual networking device type.
+	IPVTAP = "ipvtap"
+
+	// VETHTAP is a veth-tap virtual networking device type.
+	VETHTAP = "vethtap"
+
+	// VFIO is a direct assigned PCI device or PCI VF
+	VFIO = "VFIO"
+
+	// VHOSTUSER is a vhost-user port (socket)
+	VHOSTUSER = "vhostuser"
 )
+
+// QemuNetdevParam converts to the QEMU -netdev parameter notation
+func (n NetDeviceType) QemuNetdevParam() string {
+	switch n {
+	case TAP:
+		return "tap"
+	case MACVTAP:
+		return "tap"
+	case IPVTAP:
+		return "tap"
+	case VETHTAP:
+		return "tap" // -netdev type=tap -device virtio-net-pci
+	case VFIO:
+		return "" // -device vfio-pci (no netdev)
+	case VHOSTUSER:
+		return "vhost-user" // -netdev type=vhost-user (no device)
+	default:
+		return ""
+
+	}
+}
+
+// QemuDeviceParam converts to the QEMU -device parameter notation
+func (n NetDeviceType) QemuDeviceParam() string {
+	switch n {
+	case TAP:
+		return "virtio-net-pci"
+	case MACVTAP:
+		return "virtio-net-pci"
+	case IPVTAP:
+		return "virtio-net-pci"
+	case VETHTAP:
+		return "virtio-net-pci" // -netdev type=tap -device virtio-net-pci
+	case VFIO:
+		return "vfio-pci" // -device vfio-pci (no netdev)
+	case VHOSTUSER:
+		return "" // -netdev type=vhost-user (no device)
+	default:
+		return ""
+
+	}
+}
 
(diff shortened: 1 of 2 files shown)


mcastelino:
@markdryan please can you review this

qemu/qemu.go
```diff
@@ -426,42 +527,70 @@ func (netdev NetDevice) QemuParams(config *Config) []string {
...
 	}
 
-	if netdev.Script != "" {
-		netdevParams = append(netdevParams, fmt.Sprintf(",script=%s", netdev.Script))
+	// Macvtap can only be connected via fds
+	if (netdev.Type == MACVTAP) && (len(netdev.FDs) == 0) {
+		// Does not make sense to have more Queues that CPUs
+		// Specify the maximum number of queues possible for the
+		// current CPU configuration
+		netdev.FDs, _ = createMacvtapFds(netdev.ID, int(config.SMP.CPUs))
```
markdryan: Are these file descriptors are closed anywhere?  In launcher we close them after the QEMU process has been launched or has failed to launch.  I can see that the FDs aren't leaked but I can't see where they are closed.  We can't expect the user to close them as he doesn't know about them.

Also, any FDs that he assigns to netdev.FDs will be overwritten by this line and will be leaked.

markdryan:
There are also some unit test failures.

qemu/qemu.go
mcastelino: Yes. I plan to remove this logic from QEMU. I initially put it here to handle the case where the caller did not do setup properly and we have no means to return error from this API call (which is a bit unfortunate).  But now I feel it is not worth doing. Let the qemu launch fail and we can leave it to the user to recover.

mcastelino: @markdryan @sameo it may be worthwhile to provide a way to return error from a lot of these APIs. Yes, this will require changes to the consumers of these packages, but will have benefits in the future.

mcastelino:
@markdryan this fd leak should be addressed now. Also fixed the test cases.
@amshinde can you review this PR and make sure it will also work for VFIO support.

qemu/qemu_test.go
```diff
@@ -145,7 +144,32 @@ func TestAppendDeviceNetwork(t *testing.T) {
 	testAppend(netdev, deviceNetworkString, t)
 }
 
-var deviceNetworkPCIString = "-device driver=virtio-net-pci,disable-modern=true,netdev=tap0,mac=01:02:de:ad:be:ef,bus=/pci-bus/pcie.0,addr=ff -netdev tap,id=tap0,ifname=ceth0,downscript=no,script=no,fds=3:4,vhost=on"
+var deviceNetworkStringMq = "-netdev tap,id=tap0,vhost=on,fds=3:4 -device driver=virtio-net-pci,netdev=tap0,mac=01:02:de:ad:be:ef,disable-modern=true,mq=on,vectors=6"
+
+func TestAppendDeviceNetworkMq(t *testing.T) {
+	foo, _ := ioutil.TempFile(os.TempDir(), "qemu-ciao-test")
```
markdryan: If I understand correctly, the QEMU package doesn't close any file descriptors.  This is the responsibility of the caller.  Is this correct?  If so we should close the FDs in this test ( and others ).  Currently, they're leaked, i.e., we're missing a foo.Close() and a bar.Close()

mcastelino: @markdryan good catch. This was existing code which I missed. Let me fix this.

mcastelino:
@markdryan hopefully at the issues are fixed now.

qemu/qemu.go
```diff
@@ -399,68 +453,104 @@ func (netdev NetDevice) Valid() bool {
...
 
-	if netdev.Driver == VirtioNetPCI {
-		deviceParams = append(deviceParams, "driver=")
-	}
-	deviceParams = append(deviceParams, fmt.Sprintf("%s", netdev.Driver))
-	if netdev.DisableModern {
-		deviceParams = append(deviceParams, ",disable-modern=true")
-	}
+	deviceParams = append(deviceParams, "driver=")
+	deviceParams = append(deviceParams, netdev.Type.QemuDeviceParam())
```
jodh-intel: Won't this generate an invalid cli in the case of `VHOSTUSER` (where `QemuDeviceParam()` returns `""`)?

qemu/qemu.go
```diff
@@ -399,68 +453,104 @@ func (netdev NetDevice) Valid() bool {
...
-	if netdev.Script != "" {
-		netdevParams = append(netdevParams, fmt.Sprintf(",script=%s", netdev.Script))
+	return deviceParams
+}
+
+// QemuNetdevParams returns the -netdev parameters for this network device
+func (netdev NetDevice) QemuNetdevParams(config *Config) []string {
+	var netdevParams []string
+
+	netdevParams = append(netdevParams, netdev.Type.QemuNetdevParam())
```
jodh-intel: ibid.

qemu/qemu.go
mcastelino: @jodh-intel great catch.. I  have yet to add/test that path. But this a generic issue. I could always check if we have params to append to as this code will get even more complex when more use cases come in.

mcastelino: @markdryan @jodh-intel fixed.

PR #1511, opened 2017-10-05
ciao-deploy: When an SSH command fails report stdout/stderr

rbradford:
When an SSH command generates an error (e.g. non-zero exit code) embed
the output containing the command's stdout and stderr into the error
returned. This will then be printed out as part of the detailed error
message.


rbradford:
@markdryan I feel this is the best outcome after the experiments from the last few days.

PR #1512, opened 2017-10-06
singlevm: Move singlevm to use ciao-deploy setup

rbradford:
This makes our development environment match our physical cluster
environment.


rbradford:
@markdryan \o/

rbradford:
@markdryan I did test it fresh (like semaphore) and also from an old setup.

PR #1515, opened 2017-10-06
storage_bat: Test that detaching from active instance fails

rbradford:
Since it is no longer possible to detach from an active instance
ciao-cli should generate an error.  Ensure that this error occurs when
trying to detach from an active instance by creating a new BAT test.


rbradford:
@markdryan this is the BAT tested you asked me to write as part of #1486

PR #1516, opened 2017-10-06
Remove compute_fqdn

rbradford:
ciao-deploy doesn't currently set this field in the configuration but for the HTTP entry points that use absolute URLs we need to something better than os.Hostname() for when controller is being reached via a FQDN.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -209,14 +209,12 @@ func main() {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


rbradford:
@kaccardi can you test this on the DCS cluster where I know you use a certificate with a FQDN.

Is everything working fine on that cluster right now because we don't use the links in the Ciao API?

Issue #1451, opened 2017-09-26
OpenStack APIs for compute, volume, image should be migrated to ciao APIs


kaccardi:
We are no longer OpenStack compatible at all. We can abandon these old OpenStack APIs and migrate them to APIs that are designed for ciao instead.

PR #1519, opened 2017-10-09
opentack/compute: structure cleanup and code removal

rbradford:
Switch "workload list" over to using a newly added ciao API. This allows the removal of the openstack flavor APIs from the compute endpoint. This PR also cleans up some unused fields in the structures used. It also removes an unused entry point from the block API for getting absolute cluster limits.

See: #1451


rbradford:
@markdryan here is my first cleanup PR working towards #1451

PR #1524, opened 2017-10-11
Migrate image APIs to ciao and cleanup

rbradford:
Migrate the image APIs from openstack based APIs with many unused fields and structures to the ciao http endpoint. Currently this still used the ciao-image boltdb based datastore and infrastructure but in a future PR I will merge that code into controller and migrate to the sqlite database.

See: #1451


ciao-cli/image.go
```diff
@@ -251,8 +264,14 @@ func (cmd *imageListCommand) run(args []string) error {
 		}
 	}
 
-	url := buildImageURL("images")
-	resp, err := sendHTTPRequest("GET", url, nil, nil)
+	var url string
```
markdryan: This could could be factorized out into a separate function.  It's used in 5 different places.

PR #1525, opened 2017-10-12
uuid: Move the uuid package into its own top level directory

markdryan:
There was no real reason to have this package embedded inside the ssntp
package and having it in there confuses dep prune making the package
harder to vendor.


ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -30,9 +30,9 @@ import (
 
 	"github.com/ciao-project/ciao/ciao-controller/types"
 	"github.com/ciao-project/ciao/ciao-controller/utils"
-	"github.com/ciao-project/ciao/ciao-storage"
+	storage "github.com/ciao-project/ciao/ciao-storage"
```
rbradford: ^^^ ?

markdryan: Goimports made the change but the change is correct.  You can't use the package id ciao-storage in Go code as this is not a valid Go identifier.  You need to use storage.  This change makes the effective package name explicit.

Of course this change, while correct, is not really related to moving the uuid package.  It's just a by product of running goimports on the file.  I could split it out into a separate commit.

markdryan: I'll revert the "storage" change for this and the sqlite3db_test.go file.

PR #1527, opened 2017-10-16
Migrate to Ciao volume API

rbradford:
Migrate from the OpenStack block API to a ciao volume API. In this PR the HTTP endpoints are moved to the ciao style using Content-Type for versioning, and the number of endpoints is slightly reduced to remove duplication. The datastructure for use in reporting volume information is standardised on as types.BlockData (but then renamed to types.Volume) for clarity with the endpoint.

See #1451


rbradford:
Hey @markdryan. Almost there...just "compute" left to sort out.

ciao-controller/api/api.go
```diff
@@ -276,6 +302,18 @@ func listResources(c *Context, w http.ResponseWriter, r *http.Request) (Response
 
 	links = append(links, link)
 
+	// for the "volumes" resource
+	link = types.APILink{
```
markdryan: Should this be inside the if statement?

PR #1530, opened 2017-10-17
Remove openstack compute

rbradford:
Move the functionality from the openstack compute API to the ciao API renaming some uses of server (openstack terminology) to instance (ciao terminology)

See #1451


ciao-controller/api/api.go
```diff
@@ -91,6 +96,105 @@ type RequestedVolume struct {
 	ImageRef    string `json:"imageRef,omitempty"`
 }
 
+// BlockDeviceMappingV2 represents an optional block_device_mapping_v2
+// object within a /v2.1/{tenant}/servers request POST to "Create Server"
+// array of block_device_mapping_v2 objects.
+// NOTE: the OpenStack api-ref currently indicates in text that this is an
```
markdryan: Is this NOTE still relevant?

ciao-controller/api/api.go
```diff
@@ -91,6 +96,105 @@ type RequestedVolume struct {
 	ImageRef    string `json:"imageRef,omitempty"`
 }
 
+// BlockDeviceMappingV2 represents an optional block_device_mapping_v2
+// object within a /v2.1/{tenant}/servers request POST to "Create Server"
```
markdryan: Is the /v2.1/ prefix correct?  Is this still needed in our APIs?

ciao-controller/api/api.go
```diff
@@ -91,6 +96,105 @@ type RequestedVolume struct {
...
+// array of block_device_mapping_v2 objects.
+// NOTE: the OpenStack api-ref currently indicates in text that this is an
+// object not an array, but given the implementation/usage it is clearly in
+// fact an array.  Also volume size and uuid are not documented in the API
+// reference, but logically must be included.
+type BlockDeviceMappingV2 struct {
```
markdryan: Does the name need to end in V2?

ciao-controller/api/api.go
```diff
@@ -91,6 +96,105 @@ type RequestedVolume struct {
...
+	Status           string             `json:"status"`
+	TenantID         string             `json:"tenant_id"`
+	SSHIP            string             `json:"ssh_ip"`
+	SSHPort          int                `json:"ssh_port"`
+}
+
+// Servers represents the unmarshalled version of the contents of a
+// /v2.1/{tenant}/servers/detail response.  It contains information about a
+// set of instances within a ciao cluster.
+// http://developer.openstack.org/api-ref/compute/?expanded=list-servers-detailed-detail
```
markdryan: More openstack references here that are presumably made invalid by this PR.

ciao-controller/api/api.go
```diff
@@ -314,6 +418,18 @@ func listResources(c *Context, w http.ResponseWriter, r *http.Request) (Response
 		links = append(links, link)
 	}
 
+	// for the "instances" resource
+	link = types.APILink{
```
markdryan: Should this definition be inside the if statement?

ciao-controller/api/api.go
```diff
@@ -1065,6 +1181,127 @@ func volumeAction(bc *Context, w http.ResponseWriter, r *http.Request) (Response
...
+	values := r.URL.Query()
+
+	var workload string
+
+	// if this function is called via an admin context, we might
+	// have {workload} on the URL. If it's called from a user context,
+	// we might have workload as a query value.
+	workload, ok := vars["workload"]
+	if !ok {
+		if values["workload"] != nil {
```
markdryan: Not sure this check is sufficient to prevent a panic.  It should be 

if len(values["workload"]) > 0

[]string{} is a non nil slice with no elements.

rbradford:
@markdryan feedback should be addressed now

Issue #1468, opened 2017-09-28
Deleting tenant doesn't update quota service


rbradford:
Although tenant deletion removes rows from quotas database table it does not update the data in the quota service.

If a user then authenticates against the clusteer with a certificate for the deleted tenant the newly created tenant would have the same ID as the deleted tenant and so the old data in the quota service would apply even though the quotas would be empty according to the database.

Before a controller restart the tenant would see their old quotas, after a restart they would have no quotas present (as the quota service only reads from the database on startup.)

PR #1531, opened 2017-10-17
 ciao-controller: when deleting tenant also delete from quota service

rbradford:
Although the quota data was being deleted from the database it was not being deleted from the quota service. If the tenant was recreated then the quota service would have inconsistent data compared to the database.

Fixes: #1468

### DIFF
--- a/ciao-controller/internal/quotas/quotas.go
+++ b/ciao-controller/internal/quotas/quotas.go
@@ -295,38 +295,33 @@ func (qs *Quotas) Init() {
 				return
 			}
 
-			switch data.(type) {
+			switch op := data.(type) {
 
 			case *consumeOp:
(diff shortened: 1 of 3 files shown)


ciao-controller/internal/quotas/quotas.go
```diff
@@ -313,6 +322,11 @@ func (qs *Quotas) Init() {
 				dumpData := data.(*dumpOp)
 				dumpData.ch <- dump(tenantDetails, dumpData)
 				close(dumpData.ch)
+
```
markdryan: Note there's a little trick you can do with a type switch to avoid the explicit cast,

```
switch op := data.(type) {
   case *deleteTenantOp:
      // The type of op in this case statement is *deleteTenantOp
}
```

Issue #1224, opened 2017-03-03
Users can access images from other tenants


rbradford:
As the image service is not tightly integrated with controller and it's database it is possible for the contents of a user's image to leak to another user. This could leak the details of encryption keys, secrets and certificates, etc.

The only prerequisite is knowing the target's image UUID. When creating a workload there is no check that the provided UUID is an image owned by the user that is creating the workload. So therefore any user can create a workload accessing any other user's image.

It is possible that because user volumes are stored in the same way as images it might be possible for a tenant to access another tenant's volumes by using it in a workload.

obedmr:
@rbradford it will be addressed with #1123

PR #1532, opened 2017-10-18
ciao-controller: as part of workload validation check image/volume

rbradford:
Check the the volume or image specified in the workload creation request
is valid (including accessible to the current tenant.)

Fixes: #1224


rbradford:
@markdryan can finally fix this longstanding bug!

Issue #1491, opened 2017-09-29
ciao-controller doesn't calculate number of hosts per subnet correctly


kaccardi:
To duplicate: configure tenant with subnet bits of 17. Try to launch more than 253 instances.

PR #1534, opened 2017-10-19
ciao-controller: rework tenant IP allocation

kaccardi:
The existing tenant IP allocation algorithm doesn't work unless you
have a /24 network. Rework the algorithm so that it will work with
any valid network. Also add the ability to allocate more than one
IP addresses at a time.

Fixes: #1491 

### DIFF
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"net"
 	"time"
 
 	"github.com/ciao-project/ciao/ciao-controller/types"
@@ -243,6 +244,16 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance, 
 		return nil, err
 	}
 
+	var IPPool []net.IP
+
+	// if this is for a CNCI, we don't want to allocate any IPs.
+	if w.Subnet == "" {
+		IPPool, err = c.ds.AllocateTenantIPPool(w.TenantID, w.Instances)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	var newInstances []*types.Instance
 
 	for i := 0; i < w.Instances && e == nil; i++ {
@@ -255,7 +266,13 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance, 
 			}
 		}
 
-		instance, err := newInstance(c, w.TenantID, &wl, w.Volumes, name, w.Subnet)
+		var newIP net.IP
+
+		if w.Subnet == "" {
+			newIP = IPPool[i]
+		}
+
+		instance, err := newInstance(c, w.TenantID, &wl, w.Volumes, name, w.Subnet, newIP)
 		if err != nil {
 			e = errors.Wrap(err, "Error creating instance")
 			continue
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -48,7 +48,7 @@ func (c *controller) restartInstance(instanceID string) error {
 	}
 
 	if !i.CNCI {
-		err = t.CNCIctrl.WaitForActiveSubnetString(i.Subnet)
+		err = t.CNCIctrl.WaitForActive(i.Subnet)
 		if err != nil {
 			return errors.Wrap(err, "Error waiting for active subnet")
 		}
--- a/ciao-controller/command.go
+++ b/ciao-controller/command.go
@@ -19,6 +19,8 @@ package main
 import (
 	"fmt"
 	"net"
+	"runtime"
+	"sync"
 	"time"
 
 	"github.com/ciao-project/ciao/ciao-controller/types"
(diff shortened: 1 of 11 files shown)


kaccardi:
This algorithm is not as fast as the original for a single allocation, however, once we modify startWorkload to use a pool for bulk starts, we will be faster when doing bulk allocations.

ciao-controller/internal/datastore/datastore_test.go
```diff
@@ -277,6 +285,36 @@ func BenchmarkAllocateTenantIP(b *testing.B) {
 	}
 }
 
+func BenchmarkAllocate1000TenantIP(b *testing.B) {
```
markdryan: Are you concerned about performance of the IP allocator?  In some ways this isn't a good test as you are starting with a clean tenant with no IPs allocated and so is the best case for the allocator.

markdryan: If you are concerned about performance it may be possible to speed things up a little.  What operations do you perform on the  ds.tenants[tenantID].network[subnet] maps?  Does this map just exist to keep track of allocations?  Perhaps you could use a sorted slice to keep track of which ips are and are not allocated, with the allocated ips coming first, and the free ips second.  Allocation then becomes O(N) where N is the number of subnets.  De-allocation may be slower but in the worst case its a 1KB memmove.

markdryan: Or you could keep the existing algorithm but replace the make(map[uint32]bool) with a bitmask backed by big.Int.  This would use much less memory and your lookups, sets and deletes would be faster as there would be no allocations and you should have better cache performance.

markdryan: Or each subnet could have two maps or big.Ints.  One that records allocated ips and one which records free ips.  To allocate ips from a subnet you just need to iterate through the free list and move them to the allocated list.  If nothing is free you skip the subnet.  No need to iterate through it.

ciao-controller/internal/datastore/datastore.go
```diff
@@ -582,180 +587,193 @@ func (ds *Datastore) GetAllTenants() ([]*types.Tenant, error) {
...
-		if rest == 255 {
-			// this should never happen
-			glog.Warning("ran out of host numbers")
-
-			ds.tenantsLock.Unlock()
-
-			return nil, errors.New("rand out of host numbers")
+	for {
+		if start >= end {
+			addrs = nil
```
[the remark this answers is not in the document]
kaccardi: yes, thanks for catching that.

ciao-controller/internal/datastore/datastore_test.go
kaccardi: I am concerned with performance in that I don't want to regress (which this algorithm does due to the translations from addresses to bytes that didn't happen as much in the old algorithm). I can play with some of these ideas and see if they help significantly.

kaccardi: @markdryan I have come to the conclusion that working on optimizing this algorithm isn't the best use of time right now, as the larger issue during launch are the multiple sql accesses that still exist outside of the network allocation. Removing the sql table for tenant_network and parallelizing the launch seem to be a good improvement for now, and we can optimize further in the future.

kaccardi:
@markdryan I fixed the leakage issue you spotted. I discovered the problem I was seeing with deletes crashing controller was due to #1556 . So I think this PR is ready.

ciao-controller/command.go
```diff
@@ -266,65 +306,29 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance,
...
-			if name != "" {
-				if w.Instances > 1 {
-					name = fmt.Sprintf("%s-%d", name, i)
-				}
-			}
-
-			instance, err := newInstance(c, w.TenantID, &wl, w.Volumes, name, w.Subnet, newIP)
+			instance, err := c.createInstance(w, wl, name, newIP)
 			if err != nil {
 				e = errors.Wrap(err, "Error creating instance")
```
markdryan: I think this assignment is racy as e is declared outside the closure.  If multiple instances fail at the same time you'll have a race

ciao-controller/command.go
```diff
@@ -266,65 +306,29 @@ func (c *controller) startWorkload(w types.WorkloadRequest) ([]*types.Instance,
...
-
-			name := w.Name
-			if name != "" {
-				if w.Instances > 1 {
-					name = fmt.Sprintf("%s-%d", name, i)
-				}
-			}
-
-			instance, err := newInstance(c, w.TenantID, &wl, w.Volumes, name, w.Subnet, newIP)
+			instance, err := c.createInstance(w, wl, name, newIP)
```
markdryan: w and wl are effectively constant here aren't they in which case this is fine and there's no race.

ciao-controller/command.go
markdryan: You could replace the waitgroup with a chan error.  Each go routine writes to the channel when there's an error or when they're finished.  Instead of wg.Wait you have

```
for i := 0; i < w.Instances; i++ {
   err := <-errChan
  if e == nil {
    e = err
  }
}
```

This returns the first error.

markdryan: Also the first loop is wrong

 	for i := 0; i < w.Instances && e == nil; i++ {


There's no point checking e == nil here.  It's racy

ciao-controller/command.go
kaccardi: yes, they don't get modified.

kaccardi:
@markdryan I removed the waitgroups as you suggested, and this seems to be race condition free now. Thanks!

PR #1536, opened 2017-10-20
qemu: implement bridge struct

devimc:


qemu/qmp.go
```diff
@@ -629,8 +629,9 @@ func (q *QMP) ExecuteBlockdevAdd(ctx context.Context, device, blockdevID string)
 // using the device_add command.  blockdevID should match the blockdevID passed
 // to a previous call to ExecuteBlockdevAdd.  devID is the id of the device to
 // add.  Both strings must be valid QMP identifiers.  driver is the name of the
-// driver,e.g., virtio-blk-pci, and bus is the name of the bus.  bus is optional.
-func (q *QMP) ExecuteDeviceAdd(ctx context.Context, blockdevID, devID, driver, bus string) error {
+// driver,e.g., virtio-blk-pci, bus is the name of the bus and addr is the address
+// where the device is plugged. bus and addr are optional.
+func (q *QMP) ExecuteDeviceAdd(ctx context.Context, blockdevID, devID, driver, bus string, addr string) error {
```
markdryan: You could just do

driver, bus, addr string

@mcastelino I wonder at what point we're going to need to provide a structure for the parameters to this function.  Are we likely to be adding more fields in the future?

devimc: bootindex is another field that can be specified, but for now we don't need it

qemu/qmp.go
```diff
@@ -639,6 +640,9 @@ func (q *QMP) ExecuteDeviceAdd(ctx context.Context, blockdevID, devID, driver, b
 	if bus != "" {
 		args["bus"] = bus
 	}
+	if addr != "" {
```
markdryan: I'm not sure about this.  The name of the device address property is bus specific.  "addr" as used here is PCI specific.  As we can specify different buses it doesn't feel right to hardcode a PCI specific key name in here.  I guess the best thing to do would be to only set the address if the bus is specified and to use the bus type to figure out what key to use.  @mcastelino any thoughts on this?

mcastelino: @devimc can we generate the address based on PCI properties instead of using a string. That may help.

markdryan: Here's the document that concerned me

https://github.com/qemu/qemu/blob/master/docs/qdev-device-use.txt#L9

devimc: unfortunately in Q35 we have to specify the address (number between 0x0 - 0x1f) hence each device must have a unique address, probably the best way to address this conflict could be adding a new method to hot plug just PCI devices ```ExecutePCIDeviceAdd```

devimc: ```bus``` must be specified too, because we want to hot plug devices on pci(e) bridges

markdryan: Fair enough.

devimc:
@mcastelino  @markdryan  changes applied

Issue #1535, opened 2017-10-19
subnet bits should not be modified if tenant has instances


kaccardi:
Updates to subnet bits should not be allowed if there are already tenant IPs allocated.

kaccardi:
also - all cncis should be removed.

PR #1538, opened 2017-10-20
ciao-controller: do not allow subnet bits to be modified if active instances

kaccardi:
A tenant network may not contained mixed networks with variable subnet bits.
Do not allow update to tenant if there are any active instances at all,
including cncis.

Fixes: #1535


markdryan:
@kaccardi Any reason not to merge this one?

Issue #1539, opened 2017-10-23
ciao-deploy disables memory/disk checking by default


kaccardi:
ciao-deploy setup will create a configuration file with disk and memory limits disabled. This is probably not what you would want outside of a developer environment. The default should be geared towards production clusters anyway, and in this case memory/disk limit checking should be enabled so that compute nodes don't run out of memory or disk space.

markdryan:
Let's only enable limits for memory for now.  The concept of disk limits is currently broken, see #1541.

PR #1540, opened 2017-10-23
ciao-deploy: enable memory limits by default

kaccardi:
Enable memory limit checking by launcher by default,
and only disable if --disable-limits is passed to ciao-deploy setup.

Fixes: #1539 

### DIFF
--- a/ciao-deploy/deploy/setup.go
+++ b/ciao-deploy/deploy/setup.go
@@ -45,6 +45,7 @@ type ClusterConfiguration struct {
 	AuthCACertPath    string
 	AuthAdminCertPath string
 	ServerHostname    string
+	DisableLimits     bool
 }
 
 var ciaoConfigDir = "/etc/ciao"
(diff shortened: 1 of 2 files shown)


rbradford:
@kaccardi see @markdryan's comment on #1539 "Let's only enable limits for memory for now. The concept of disk limits is currently broken, see #1541."

rbradford:
@kaccardi  Thanks for the update. The changes in ciao-deploy lgtm. However do you know if we need to disable in singlevm.

Could you also update the commit message (i see you've updated the PR description already.)

kaccardi:
@rbradford I confirmed you do not need to disable memory in single vm, so I removed that.

ciao-deploy/deploy/setup.go
```diff
@@ -78,8 +79,13 @@ func createConfigurationFile(ctx context.Context, clusterConf *ClusterConfigurat
 
 	config.Configure.Launcher.ComputeNetwork = []string{clusterConf.ComputeNet}
 	config.Configure.Launcher.ManagementNetwork = []string{clusterConf.MgmtNet}
+
+	// Disk limit checking is broken and should always be disabled.
+	// See issue #1541
 	config.Configure.Launcher.DiskLimit = false
-	config.Configure.Launcher.MemoryLimit = false
+	if clusterConf.DisableLimits {
```
markdryan: I can't see how this has any effect.  By default,

config.Configure.Launcher.MemoryLimit = false

When we supply --disable-limits it will still be false.

Should this be
```
if !clusterConf.DisableLimits {
    config.Configure.Launcher.MemoryLimit = true
}
```

markdryan: or simply

config.Configure.Launcher.MemoryLimit = !clusterConf.DisableLimits

kaccardi: I swear this was tested... but you are right, this doesn't do anything. I must have broken something when I was doing some quick adjustment. will fix, sorry.

PR #1546, opened 2017-10-25
Tweak error handling in ciao-controller

rbradford:
Make ciao-controller "errcheck clean" but also also improve the logging so that we get messages in the log whenever we service a HTTP request that generates an error.


rbradford:
@markdryan sadly semaphore is broken but I have run BAT locally.

Issue #1556, opened 2017-11-01
ciao-controller is aborted while deleting thousands of instances


kaccardi:
To duplicate, launch 2000 instances. Then delete them all with ciao-cli instance delete -all. check log with journalctl -u ciao-controller.service.

ciao-controller[26635]: runtime/cgo: pthread_create failed: Resource temporarily unavailable
ciao-controller[26635]: SIGABRT: abort

kaccardi:
This is due to the systemd TasksMax value being too low.

PR #1557, opened 2017-11-01
ciao-deploy: set TasksMax in systemd service file

kaccardi:
In newer kernels, systemd's TasksMax is set too low for ciao
when launching thousands of instances. Set TasksMax to infinity
for the service file to allow ciao to spawn larger numbers of
pthreads.

Fixes: #1556

### DIFF
--- a/ciao-deploy/deploy/setup.go
+++ b/ciao-deploy/deploy/setup.go
@@ -200,6 +200,7 @@ Type=simple
 ExecStart=/usr/local/bin/%s --cacert=%s --cert=%s --v 3
 Restart=no
 KillMode=process
+TasksMax=infinity
 
 [Install]
 WantedBy=multi-user.target


ciao-deploy/deploy/setup.go
```diff
@@ -200,6 +200,7 @@ Type=simple
 ExecStart=/usr/local/bin/%s --cacert=%s --cert=%s --v 3
 Restart=no
 KillMode=process
+TasksMax=infinity
```
markdryan: I guess we only need this for launcher.   I have a patch that I'm developing for the running as non-root issue that allows us to customize these unit files.  We could merge this for now and I can make this launcher only in my patch.

rbradford: @markdryan I wonder if this could also happen in ciao-controller. It starts a new goroutine for each instance and although they are not long lived (for sending messages by ssntp) there is a good chance the number of alive and blocked ones will be quite high if a large number of instances are created simultaneously

markdryan: I'd thought TaskMax was related to creation of new processes via forks and not the creation of new threads.  However, thinking about it, controller does launch processes when creating images and snapshots, so it needs this too particularly if it does this in parallels.  Let's merge it.

kaccardi: FYI - this does happen on controller due to forks for snapshot/volume deletion (try launching 2000 instances, then typing ciao-cli instance delete -all)

PR #1561, opened 2017-11-06
ciao-deploy: Don't upload images as admin for BAT

rbradford:
We don't create the workloads as admin so we shouldn't create the
images as admin.

### DIFF
--- a/ciao-controller/server.go
+++ b/ciao-controller/server.go
@@ -32,7 +32,8 @@ import (
 )
 
 type clientCertAuthHandler struct {
-	Next http.Handler
+	Controller *controller
+	Next       http.Handler
 }
 
 func (h *clientCertAuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


ciao-controller/server.go
```diff
@@ -69,6 +70,13 @@ func (h *clientCertAuthHandler) ServeHTTP(w http.ResponseWriter, r *http.Request
 	}
 
 	r = r.WithContext(service.SetTenantID(r.Context(), tenantFromVars))
+	if tenantFromVars != "" {
+		err := h.Controller.confirmTenant(tenantFromVars)
+		if err != nil {
+			http.Error(w, "Error confirming tenant", http.StatusInternalServerError)
```
markdryan: Will this end up returning a different error code?  Should we return errorResponse(err) instead.  Note it's not entirely clear to me that errorResponse works properly with wrapped errors.

markdryan: I guess there's an inherent race here.  Just because the tenant is confirmed at this point of the transaction doesn't mean that the tenant will still exist when we actually create the volume.   Someone else could be calling DeleteTenant at the same time.  As far as I can tell, this race already exists in the code, so this is a question rather than a comment.  Will moving the tenant check here make it harder to fix the race, should it need fixing?  Maybe it makes it easier.

PR #1567, opened 2017-11-06
ciao-deploy: use a generic base clear image for building cnci.

kaccardi:
Rather than use a custom image, just use the latest clear release
as a base image and install the packages we need.

### DIFF
--- a/ciao-deploy/deploy/cnci.go
+++ b/ciao-deploy/deploy/cnci.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path"
@@ -28,7 +30,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-var cnciURL = "https://download.clearlinux.org/demos/ciao/clear-8260-ciao-networking.img.xz"
 var cnciImageID = "4e16e743-265a-4bf2-9fd1-57ada0b28904"
 
 func mountImage(ctx context.Context, fp string, mntDir string) (string, error) {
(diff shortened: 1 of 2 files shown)


kaccardi:
@rbradford my best guess on why this is failing on semaphore is the size of the image. We can try using a different clear base image - the cloud image is smaller, but would need the addition of "network-basic".
and "kvm_host" - https://download.clearlinux.org/current/config/image/cloud-config.json

kaccardi:
and that image is qcow2

rbradford:
@kaccardi  I think you need to install even more. I tried this on ccloudvm and i noticed that e.g. no dnsmasq was inside: here is the list of bundles for the original ciao-networking image: https://download.clearlinux.org/current/config/image/ciao-networking-config.json

rbradford:
@kaccardi  also does the chroot swupd process work inside a proxied network?

kaccardi:
@rbradford yes it should work on proxied network, but that's why we have to do that copy of /etc/resolv.conf before running swupd.

kaccardi:
@rbradford I think this change will address your proxy issues. I also added code to clean up any old clear images that might be left if latest changes.

ciao-deploy/deploy/utils.go
```diff
@@ -47,6 +47,31 @@ func DefaultImageCacheDir() string {
 	return path.Join(u.HomeDir, ".cache", "ciao", "images")
 }
 
+// CleanImage will remove all images from the cache directory that
+// match a specific pattern.
+func CleanImage(pattern string, keep []string, imageCacheDir string) error {
```
rbradford: I think this would be better named CleanupImages

ciao-deploy/deploy/utils.go
```diff
@@ -47,6 +47,31 @@ func DefaultImageCacheDir() string {
 	return path.Join(u.HomeDir, ".cache", "ciao", "images")
 }
 
+// CleanImage will remove all images from the cache directory that
+// match a specific pattern.
```
rbradford: And can you mention what keep does?

ciao-deploy/deploy/cnci.go
```diff
@@ -183,6 +241,27 @@ func prepareImage(ctx context.Context, baseImage string, agentCertPath string, c
 	return preparedImagePath, nil
 }
 
+func getCNCIURL() (string, error) {
```
rbradford: This should take a context.Context

ciao-deploy/deploy/cnci.go
```diff
@@ -81,6 +82,27 @@ func unMountImage(ctx context.Context, devPath string, mntDir string) error {
 	return errOut
 }
 
+func getProxy(upper, lower string) (string, error) {
```
rbradford: I don't think it's worth taking these strings as parameters. Just hardcode them in this function.

PR #1573, opened 2017-11-10
Simplify workload requirements specification

rbradford:
Simplify the way that the requirements to run a workload are specified in the payload of an SSNTP command. Replace the slice of resource requests with a simplified structure capturing the requirements to run the workload which can be further extended as necessary and later exposed directly to users in the workload definition.

This is enabling work to allow users to select a node to host their workload by node_id and hostname. The implementation of that requires the exposure of the new requirements as part of the types.Workload definition and will follow later.


ciao-launcher/tests/examples/start_legacy_volume.yaml
```diff
@@ -1,10 +1,8 @@
 ---
 start:
-  requested_resources:
-     - type: vcpus
-       value: 2
-     - type: mem_mb
-       value: 370
+  requirements:
+    vcpus: 2
+    mem_mb: 256
```
markdryan: Was there a reason for changing this value?

ciao-launcher/tests/examples/start_nn.yaml
```diff
@@ -3,13 +3,10 @@
...
-  requested_resources:
-     - type: vcpus
-       value: 2
-     - type: mem_mb
-       value: 256
-     - type: network_node
-       value: 1
+  requirements:
+    vcpus: 2
+    mem_mb: 370
```
markdryan: Was there a reason for changing this value?

ciao-scheduler/scheduler.go
```diff
@@ -471,104 +471,33 @@ func (sched *ssntpSchedulerServer) StatusNotify(uuid string, status ssntp.Status
...
 
-func networkDemandsSatisfied(node *nodeStat, workload *workResources) bool {
-	if !node.isNetNode {
-		return true
-	}
-
-	var matchedNetworksCount int
-	var requestedNetworksCount int
-
-	for _, requestedNetwork := range workload.physNets {
```
markdryan: Was this stuff simply not used?

ciao-launcher/tests/examples/start_legacy_volume.yaml
rbradford: Just a copy and paste flail.

ciao-launcher/tests/examples/start_nn.yaml
rbradford: Again just a copy and paste error.

ciao-scheduler/scheduler.go
rbradford: With the exception of the !node.isNetNode() which is now done in the caller, yep this code was not used.

PR #1575, opened 2017-11-13
[RFC] CLI Rework & New SDK

tmarcu:
This PR introduces a new (reworked) CLI in /ciao, as well as a library it uses to perform given actions (ciao-sdk). The list & show functionality is fully available with supporting sub-options for all the existing ciao objects, and abstracts the new client library functionality a bit further to make it easier to use.


tmarcu:
I'm interested in what you think of this implementation, and if you like how the templating is done in ciao-sdk/tenants.go. There are a couple spots such as quotas where the general case template will not print nicely - should these be handled or constructed specially or given a different default template?

client/client.go
```diff
@@ -51,6 +54,28 @@ type queryValue struct {
...
+	if glog.V(1) {
+		glog.InfoDepth(1, fmt.Sprintf("ciao-cli INFO: "+format, args...))
+	}
+}
+
+func Errorf(format string, args ...interface{}) {
+	glog.ErrorDepth(1, fmt.Sprintf("ciao-cli ERROR: "+format, args...))
+}
+
+func Fatalf(format string, args ...interface{}) {
```
tmarcu: I think these changes can actually be dropped, ignore for now since the SDK is not printing anything anymore and only returning data.

ciao-sdk/instance.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"github.com/ciao-project/ciao/ciao-controller/api"
+	"github.com/ciao-project/ciao/ciao-controller/types"
+	"github.com/ciao-project/ciao/client"
+	"github.com/pkg/errors"
+)
+
+func GetInstance(c *client.Client, flags CommandOpts) (api.Server, error) {
+	if len(flags.Args) == 0 {
+		errors.New("Missing required -cn parameter")
```
rbradford: Looks bogus.

tmarcu:
@rbradford Yeah I talked to Kristen about that as well, and I wasn't convinced that the ciao-sdk package needs to exist anymore. What package would you move it into? Just merge into the client one?

ciao-sdk/instance.go
```diff
@@ -0,0 +1,57 @@
...
+
+	servers, err := c.ListInstancesByWorkload(flags.Tenant, flags.Workload)
+	if err != nil {
+		return []api.ServerDetails{}, errors.Wrap(err, "Error listing instances")
+	}
+
+	Servers := []api.ServerDetails{}
+	for _, v := range servers.Servers {
+		Servers = append(Servers, v)
+	}
```
markdryan: Why not just return servers.Servers?

ciao-sdk/node.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nodes, err
+}
+
+func listCNCINodes(c *client.Client, flags CommandOpts) (types.CiaoCNCIs, error) {
+	cncis, err := c.ListCNCIs()
+	if err != nil {
+		return cncis, errors.Wrap(err, "Error listing CNCIs")
+	}
+	/*
```
markdryan: We should have any commented out code.

ciao-sdk/tenant.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/intel/tfortools"
+	"github.com/pkg/errors"
+)
+
+func GetTenants(c *client.Client, flags CommandOpts) (bytes.Buffer, error) {
+	var t *template.Template
+	var result bytes.Buffer
+
+	if c.Template != "" {
+		var err error
```
markdryan: You could factorize some of the code here.  The only difference I see between the two blocks is the template program.

ciao-sdk/trace.go
```diff
@@ -0,0 +1,31 @@
...
+		return traces, errors.Wrap(err, "Error listing trace labels")
+	}
+
+	return traces, err
+}
+
+func GetTraceData(c *client.Client, flags CommandOpts) (types.CiaoTraceData, error) {
+	var traceData types.CiaoTraceData
+
+	if flags.Args[0] == "" {
```
markdryan: Are we sure there'll always be at leas one argument if we get here?

tmarcu: Yes it's guaranteed since it will only be called when Show() is given some argument(s). If we're trying to future proof it for use in other code paths I can add the check in here as well, but afaik that's not the plan.

ciao/cmd/detach.go
```diff
@@ -0,0 +1,68 @@
...
+		fmt.Printf("Requested unmap of: %s\n", IP)
+	},
+}
+
+var detachVolCmd = &cobra.Command{
+	Use:  "volume [UUID]",
+	Long: `Detach a volume from an instance`,
+	Args: cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		volume := args[0] /* Guaranteed from MinimumNArgs(1) */
```
rbradford: Ditto.

ciao/cmd/restart.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"os"
+
+	"github.com/ciao-project/ciao/ciao/tool"
+
+	"github.com/spf13/cobra"
+)
+
+var restartCmd = &cobra.Command{
+	Use:   "restart [UUID]",
```
rbradford: The rest of the commands are <verb> <type>, are you worried about the inconsistency? What happens if ciao gets something else we can restart in the future, beyond instances, nodes? unicorns? flying cars?

ciao/cmd/root.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/ciao-project/ciao/ciao/tool"
+	"github.com/ciao-project/ciao/client"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var C client.Client
```
rbradford: This only needs to be capitalised if it is used by another package...is that the case?

ciao/cmd/root.go
```diff
@@ -0,0 +1,132 @@
...
+
+var CommandFlags = new(tool.CommandOpts)
```
rbradford: What other package needs this?

ciao/cmd/root.go
```diff
@@ -0,0 +1,132 @@
...
+
+var cfgFile string
+
+var (
+	tenantID       = new(string)
```
rbradford: I've never seen people do it this way, maybe just follow the normal way to declare?

client/client.go
rbradford: Yes please don't add these.

client/client.go
```diff
@@ -51,6 +54,28 @@ type queryValue struct {
...
+	glog.FatalDepth(1, fmt.Sprintf("ciao-cli FATAL: "+format, args...))
+}
+
+func (client *Client) PrettyPrint(buff *bytes.Buffer, tname string, obj interface{}) {
+	if client.Template != "" {
+		tfortools.OutputToTemplate(buff, tname, client.Template, obj, nil)
+	} else {
+		tfortools.OutputToTemplate(buff, tname, "{{table .}}", obj, nil)
+	}
+}
```
rbradford: Move PrettyPrint to your package.

ciao/cmd/root.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+	getCiaoEnvVariables()
+	C.Init()
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
```
rbradford: I'd rather we didn't use the viper stuff for now and stick with just environment variables. We can look at adding it later.

ciao/cmd/show.go
```diff
@@ -0,0 +1,202 @@
...
+)
+
+var showCmd = &cobra.Command{
+	Use:   "show",
+	Short: "Show information about various ciao objects",
+	Long:  `Show outputs a list and/or details for available commands`,
+}
+
+var eventShowCmd = &cobra.Command{
+	Use:  "event [TENANT ID]",
```
rbradford: Let's come up with a consistent way to document mandatory vs optional positional arguments. man pages seem to use 
[foo] for optional and foo for non optional.

ciao/cmd/show.go
```diff
@@ -0,0 +1,202 @@
...
+		CommandFlags.Args = args
+		result, err := tool.Show(&C, object, *CommandFlags)
+		if err == nil {
+			fmt.Println(result.String())
+		}
+	},
+}
+
+var externalipShowCmd = &cobra.Command{
+	Use:  "externalip",
```
rbradford: s/externalip/external-ip/ to be consistent with elsewhere?

ciao/cmd/show.go
```diff
@@ -0,0 +1,202 @@
...
+	for _, cmd := range showcmds {
+		showCmd.AddCommand(cmd)
+	}
+
+	RootCmd.AddCommand(showCmd)
+
+	showCmd.PersistentFlags().StringVarP(&C.Template, "template", "t", "", "Template used to format output")
+
+	eventShowCmd.Flags().BoolVar(&CommandFlags.All, "all", false, "List events for all tenants in a cluster")
+	eventShowCmd.Flags().StringVar(&CommandFlags.Tenant, "tenant", "", "Tenant to list events from")
```
rbradford: This is positional according to the documentation above?

ciao/cmd/show.go
```diff
@@ -0,0 +1,202 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/ciao-project/ciao/ciao/tool"
+	"github.com/spf13/cobra"
+)
+
+var showCmd = &cobra.Command{
```
rbradford: I don't like the overloading of show to mean show one and show multiple. Instead I think that should be a list command.

ciao/cmd/show.go
```diff
@@ -0,0 +1,202 @@
...
+	Long:  `Show outputs a list and/or details for available commands`,
+}
+
+var eventShowCmd = &cobra.Command{
+	Use:  "event [TENANT ID]",
+	Long: `When called with no args, it will print all events.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		object := strings.Fields(cmd.Use)[0]
+		CommandFlags.Args = args
+		result, err := tool.Show(&C, object, *CommandFlags)
```
rbradford: Now that the tool package is just for the command line i'm happy with that to print out.

ciao/cmd/root.go
rbradford: I this the flags should be in this package and also split up for the sub sub commands (like you did for add.)

ciao/cmd/stop.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/ciao-project/ciao/ciao/tool"
+
+	"github.com/spf13/cobra"
+)
+
+var stopCmd = &cobra.Command{
```
rbradford: Ditto what I said about restart.

ciao/cmd/add.go
```diff
@@ -0,0 +1,902 @@
...
+	"github.com/intel/tfortools"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+)
+
+var config string
+
+var addCmd = &cobra.Command{
+	Use:   "add",
```
rbradford: One thing that we talked about is that "add" has always bugged us. It should probably be create. @kaccardi does that align with your recollection?

ciao/tool/externalip.go
```diff
@@ -0,0 +1,16 @@
+package tool
+
+import (
+	"github.com/ciao-project/ciao/ciao-controller/types"
+	"github.com/ciao-project/ciao/client"
+	"github.com/pkg/errors"
+)
+
+func ListExternalIP(c *client.Client, flags CommandOpts) ([]types.MappedIP, error) {
```
rbradford: This function doesn't seem to add much value, call client directly in cmd?

ciao/tool/event.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/ciao-project/ciao/client"
+	"github.com/pkg/errors"
+)
+
+func ListEvents(c *client.Client, flags CommandOpts) ([]types.CiaoEvent, error) {
+	if flags.Tenant == "" {
+		flags.Tenant = c.TenantID
+	}
+
+	if flags.All == false && flags.Tenant == "" {
```
rbradford: I think flag checking should be in cmd.

ciao/tool/image.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/ciao-project/ciao/ciao-controller/types"
+	"github.com/ciao-project/ciao/client"
+	"github.com/pkg/errors"
+)
+
+func GetImage(c *client.Client, flags CommandOpts) (types.Image, error) {
+	imageID := flags.Args[0]
+
+	if imageID == "" {
```
rbradford: flag checking should be in cmd.

ciao/tool/instance.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"github.com/ciao-project/ciao/ciao-controller/api"
+	"github.com/ciao-project/ciao/ciao-controller/types"
+	"github.com/ciao-project/ciao/client"
+	"github.com/pkg/errors"
+)
+
+func StartStopInstance(c *client.Client, instance string, stop bool) error {
+	if c.TenantID == "" {
```
rbradford: Ditto flag checking should be in cmd.

ciao/tool/instance.go
```diff
@@ -0,0 +1,80 @@
...
+		return errors.New("Missing required -tenant-id parameter")
+	}
+
+	if instance == "" {
+		return errors.New("Missing required -instance parameter")
```
rbradford: Instance ID should be positional and checked in cmd.

ciao/tool/instance.go
```diff
@@ -0,0 +1,80 @@
...
+	return server, nil
+}
+
+func GetNodeInstances(c *client.Client, flags CommandOpts) ([]types.CiaoServerStats, error) {
+	if flags.Tenant == "" {
+		flags.Tenant = c.TenantID
+	}
+
+	if flags.ComputeName == "" {
+		errors.New("Missing required -cn parameter")
```
rbradford: Flag checking should be in cmd.

ciao/tool/node.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return nodes, err
+}
+
+func listCNCINodes(c *client.Client, flags CommandOpts) (types.CiaoCNCIs, error) {
+	cncis, err := c.ListCNCIs()
+	if err != nil {
+		return cncis, errors.Wrap(err, "Error listing CNCIs")
+	}
+	/*
```
rbradford: Deadcode.

ciao/tool/show.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/pkg/errors"
+)
+
+/* Intended calling convention by user would be:
+ * instances := tool.Show(ClientHandle, "instance", nil)
+ * for example, to return a string of all instances using
+ * the default tfortools template {{table .}}. This can
+ * then be printed or parsed as needed. */
+func Show(c *client.Client, objName string, data CommandOpts) (bytes.Buffer, error) {
```
rbradford: I think this should be split into List/Show and each type should be their own flags. With that I think it makes sense not to try and generalise this code and instead just do it for each object type separately.

ciao/cmd/add.go
```diff
@@ -0,0 +1,902 @@
...
+func argMatch(patterns []string, arg string) bool {
+	for _, pattern := range patterns {
+		matched, _ := regexp.MatchString(pattern, arg)
+		if matched {
+			return true
+		}
+	}
+	return false
+}
+func isInstanceVolumeBoolArg(arg string) bool {
```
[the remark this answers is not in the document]
tmarcu: I did copy them for these supporting functions heh, I can change/update the code though. A lot was pulled from the current CLI implementation and just layered under the new CLI layout.

ciao/cmd/detach.go
```diff
@@ -0,0 +1,68 @@
...
+	Long:  `Detach the objects below from their given connections`,
+	Args:  cobra.MinimumNArgs(2),
+}
+
+var detachIpCmd = &cobra.Command{
+	Use:  "external-ip [IP]",
+	Long: `Detach an external IP`,
+	Args: cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		IP := args[0] /* Guaranteed from MinimumNArgs(1) */
```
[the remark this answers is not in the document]
tmarcu: I received a comment asking about this type of thing previously, so just covering my bases ;P

ciao/cmd/show.go
tmarcu: +1

ciao/cmd/root.go
tmarcu: I don't think normal is the right word here... this can be changed, though.

ciao/cmd/restart.go
tmarcu: They can be trivially added to the restartCmd when they need to be supported :)

ciao/cmd/delete.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+var workloadDelCmd = &cobra.Command{
+	Use:  "workload [UUID]",
+	Long: `Delete a workload.`,
+	Args: cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		workload := args[0]
+		if workload == "" {
+			fmt.Println("Missing required workload UUID paramter")
```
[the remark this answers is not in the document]
tmarcu: Agreed +1

ciao/cmd/detach.go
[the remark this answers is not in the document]
markdryan: The question was from me as I was too lazy to check it myself.  Whenever I see things like args[0] in code without a check alarm bells start ringing.  Having said that I'm happy with your earlier response that args[0] is safe and I agree the comment isn't needed here.

ciao/cmd/add.go
```diff
@@ -0,0 +1,902 @@
...
+	Args: cobra.MinimumNArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		poolFlags.name = args[0]
+		if poolFlags.name == "" {
+			fmt.Println("Missing required pool NAME parameter")
+			return
+		}
+
+		/* Handle the create option so we can add new pools */
+		if poolFlags.new {
```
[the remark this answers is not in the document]
tmarcu: External-ip and pool are mashed into external-ip in the client library now, so I assumed by that they should go together..which I don't particularly like cause they seem separate to me.

ciao/cmd/add.go
```diff
@@ -0,0 +1,902 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		if imgFlags.name == "" {
+			fmt.Println("Missing required --name parameter")
+			return
+		}
+
+		if imgFlags.file == "" {
+			fmt.Println("Missing required --file parameter")
+			return
+		}
```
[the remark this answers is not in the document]
tmarcu: Yes, I will update the mandatory args and remove them from the -- flags section.

ciao/cmd/add.go
tmarcu: Up until now I thought everyone was in agreement over which commands are to be merged and which ones will end up being used to perform actions going forward. This seems off track to split things up again, and a large change that should have a written spec if it's going to be done.

ciao/cmd/add.go
```diff
@@ -0,0 +1,902 @@
...
+	Long: `Create a new tenant with the supplied flags`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if !C.IsPrivileged() {
+			fmt.Fprintf(os.Stderr, "Creating tenants is only available for privileged users")
+		}
+
+		if tenantFlags.tenantID == "" {
+			fmt.Fprintf(os.Stderr, "Missing required tenantID")
+			return
+		}
```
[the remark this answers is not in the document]
kaccardi: I think this could be changed if you changed the controller api - you'd have to output the new tenant ID so that an admin can then use that to generate a cert.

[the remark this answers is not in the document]
kaccardi: I would recommend we leave the requirement the way it is for now, and file an issue with a request to change this to not require a tenant ID. That way we can keep controller out of the picture with this PR.

rbradford:
One more thing that occurs to me is that unfortunately you're going to rework all the code paths that generate an error to return a non-zero exit code. It's a shame that the cobra Run field isn't a function that returns an int (which would then be passed to os.Exit()). We need to return non-zero exit codes on error for BAT testing (and general goodness.)

Check out how ciao-deploy does it as it has exactly the same problem, it's better to have a wrapper function as calling os.Exit() means deferred functions do not get called and we will need to have teh ability to run deferred functions as client will soon be gaining a context.Context for all calls.

PR #1578, opened 2017-11-13
ciao-controller: Use payload.WorkloadRequirements in types.Workload

rbradford:
Replace the concept of workload defaults with workload requirements in the user experience and in the database store. This opens up the opportunity for extending workload requirements beyond the current CPU and memory needs to include further workload routing such as node id and hostname and capabilities such as CPU features.


rbradford:
Superseded by #1580

PR #1580, opened 2017-11-15
Workload routing by node ID or hostname requirements

rbradford:
This PR adds support for routing workloads by hostname or workload requirements. It is based off #1578 that exposes the new workload requirement struct to the user.

### DIFF
--- a/_release/bat/workload_bat/workload_bat_test.go
+++ b/_release/bat/workload_bat/workload_bat_test.go
@@ -16,6 +16,7 @@ package workloadbat
 
 import (
 	"context"
+	"errors"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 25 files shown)


_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -265,3 +266,309 @@ func TestCreateWorkloadWithSizedVolume(t *testing.T) {
...
+	}
+
+	workloadID, err := bat.CreateWorkload(ctx, tenant, opt, vmCloudInit)
+	if err != nil {
+		t.Error(err)
+	}
+
+	w, err := bat.GetWorkloadByID(ctx, tenant, workloadID)
+	if err != nil {
+		t.Fatal(err)
```
markdryan: Should this be t.Error?  Otherwise the workload might not get deleted.

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -265,3 +266,309 @@ func TestCreateWorkloadWithSizedVolume(t *testing.T) {
...
+}
+
+// Check that scheduling by requirement works if the workload
+// cannot be scheduled
+//
+// Create a workload with a hostname requirement that cannot be met
+//
+// The workload should be created but an instance should not be successfully
+// created for that workload.
+func TestCreateUnschedulableHostnameWorkload(t *testing.T) {
```
markdryan: This function and the one above seem to be identical apart from one line.

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -265,3 +266,309 @@ func TestCreateWorkloadWithSizedVolume(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	w, err := bat.GetWorkloadByID(ctx, tenant, workloadID)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	instances, err := bat.LaunchInstances(ctx, tenant, w.ID, 1)
```
markdryan: We could check the node-id here to make sure it's actually worked.

_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -265,3 +266,309 @@ func TestCreateWorkloadWithSizedVolume(t *testing.T) {
...
+	}
+
+	workloadID, err := bat.CreateWorkload(ctx, tenant, opt, vmCloudInit)
+	if err != nil {
+		t.Error(err)
+	}
+
+	w, err := bat.GetWorkloadByID(ctx, tenant, workloadID)
+	if err != nil {
+		t.Fatal(err)
```
markdryan: Probably should be t.Error

rbradford:
@markdryan I've addressed your feedback and refactored the common code into a function (and also check the node ID is correct where possible.)

PR #1583, opened 2017-11-17
Reduce capabilities of ciao services and their children

markdryan:
This is a large PR designed to improve the security of ciao ( and fix a number of other bugs in the process ).  Essentially, it modifies ciao-deploy and the ciao services to ensure that these services run with lower privileges than before.   ciao-scheduler and ciao-controller now run as non-root (under a new user called ciao).  ciao-launcher still runs as root but with reduced capabilities.  The majority of the processes spawned by ciao-launcher now also run as the ciao user with no special privileges.  This means that the qemu instances that support VM workloads no longer run as root.

This PR also modifies the way in which ciao's dependencies are installed.  It creates a new set of services ciao-[service-name]-prepare that runs the services as root passing an -oprepare flag.  The services install any needed dependencies and exit.  They are restarted by the existing systemd unit files which now run them with reduced privileges.

The PR also fixes a number of other issues

1.  It is now possible for VM instances to take advantage of virtio-net multiqueue support potentially improving their network performance.  We're not currently taking advantage of this yet, but enabling it is now a 1 line change.
2. The introduction of the ciao-[service]-prepare unit files fixes a race condition noticeable on clean installs on fedora and clear.  Notable ciao-deploy returns before the cluster is fully set up and before launching is ready to receive workloads.

Fixes: https://github.com/ciao-project/ciao/issues/669

### DIFF
--- a/ciao-deploy/deploy/nodes.go
+++ b/ciao-deploy/deploy/nodes.go
@@ -152,7 +152,8 @@ func createRemoteCiaoDirectory(ctx context.Context, hostname, sshUser, path stri
 }
 
 func setupNode(ctx context.Context, anchorCertPath string, caCertPath string, hostname string, sshUser string, networkNode bool) (errOut error) {
-	err := SSHRunCommand(ctx, sshUser, hostname, fmt.Sprintf("sudo useradd -r %s -G docker,kvm", ciaoUser))
+	err := SSHRunCommand(ctx, sshUser, hostname,
+		fmt.Sprintf("sudo useradd -r %s -G docker,kvm -d %s -s /bin/false", ciaoUser, ciaoDataDir))
 	if err != nil {
 		return errors.Wrapf(err, "Error creating %s user", ciaoUser)
 	}
--- a/ciao-deploy/deploy/nodes.go
+++ b/ciao-deploy/deploy/nodes.go
@@ -259,6 +259,9 @@ func setupNode(ctx context.Context, anchorCertPath string, caCertPath string, ho
 		Caps: []string{"CAP_NET_ADMIN", "CAP_NET_RAW", "CAP_DAC_OVERRIDE",
 			"CAP_SETGID", "CAP_SETUID", "CAP_SYS_PTRACE", "CAP_SYS_MODULE"},
 		Roles: roles,
+		Deps: []string{
+			"docker.service",
+		},
 	})
 	if err != nil {
 		return errors.Wrap(err, "Error installing tool on node")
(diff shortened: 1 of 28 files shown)


markdryan:
@mcastelino I'd appreciate your feedback on these two commits.

https://github.com/ciao-project/ciao/pull/1583/commits/89da128e66b7fce53851aae6cf6c796a9b5f9b04

https://github.com/ciao-project/ciao/pull/1583/commits/ee64776e0a1eca450c4c0eb9a3da0b3b06df2d9e

markdryan:
I'm going to split this one up a little.  Let's not merge for now.

markdryan:
Okay, all the non-security related commits that were originally part of this PR have been submitted as separate PRs and merged,

ciao-launcher/main.go
```diff
@@ -93,6 +100,8 @@ func init() {
 	flag.BoolVar(&hardReset, "hard-reset", false, "Kill and delete all instances, reset networking and exit")
 	flag.BoolVar(&simulate, "simulation", false, "Launcher simulation")
 	flag.StringVar(&cephID, "ceph_id", "", "ceph client id")
+	flag.BoolVar(&prepare, "osprepare", false, "Install dependencies")
+	flag.StringVar(&roles, "roles", "agent", "Roles for which dependencies are to be installed")
```
rbradford: Suggestion: As --roles is only used for --osprepare you could make the latter a string variable and remove the duplicate

ciao-deploy/deploy/nodes.go
```diff
@@ -160,7 +259,18 @@ func setupNode(ctx context.Context, anchorCertPath string, caCertPath string, ho
 		}
 	}()
 
-	err = InstallToolRemote(ctx, sshUser, hostname, "ciao-launcher", remoteCertPath, caCertPath)
+	err = InstallToolRemote(ctx, sshUser, hostname, unitFileConf{
+		Tool:       "ciao-launcher",
+		User:       ciaoUser,
+		CertPath:   remoteCertPath,
+		CACertPath: caCertPath,
+		Caps: []string{"CAP_NET_ADMIN", "CAP_NET_RAW", "CAP_DAC_OVERRIDE",
```
rbradford: Do we still need CAP_DAC_OVERRIDE? That's a pretty permissive capability...

markdryan: Afraid so.  I couldn't figure out how to get things to work without it.

ciao-launcher/main.go
markdryan: Good idea.  I can't see any reason not to do this.   I'll make the change.

markdryan: We discussed this offline and decided not to make the change in the end as the difference in behaviour between launcher and the other services might be a little confusing.

PR #1593, opened 2017-11-20
ciao-deploy: do not use latest for cnci version, use last known good

kaccardi:
It turns out that the "latest" file that is located on clearlinux.org
only refers to the latest generated image, but that image may not
necessarily end up in download.clearlinux.org/image if it does not
pass QA for some reason. Therefore, any cloud image located in
clearlinux.org/image is the last known good image and should be
used. Modify getCNCIURL() to download the images listing and look
for the cloud image filename to use from there.

### DIFF
--- a/ciao-deploy/deploy/cnci.go
+++ b/ciao-deploy/deploy/cnci.go
@@ -17,10 +17,14 @@ package deploy
 import (
 	"context"
 	"fmt"
+	"golang.org/x/net/html"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path"
+	"regexp"
 	"strings"
 
 	"github.com/ciao-project/ciao/bat"
@@ -28,7 +32,6 @@ import (
 	"github.com/pkg/errors"
 )
 
-var cnciURL = "https://download.clearlinux.org/demos/ciao/clear-8260-ciao-networking.img.xz"
 var cnciImageID = "4e16e743-265a-4bf2-9fd1-57ada0b28904"
 
 func mountImage(ctx context.Context, fp string, mntDir string) (string, error) {
@@ -81,6 +84,27 @@ func unMountImage(ctx context.Context, devPath string, mntDir string) error {
 	return errOut
 }
 
+func getProxy(env string) (string, error) {
+	proxy := os.Getenv(strings.ToLower(env))
+	if proxy == "" {
+		proxy = os.Getenv(strings.ToUpper(env))
+	}
+
+	if proxy == "" {
+		return "", nil
+	}
+
+	if proxy[len(proxy)-1] == '/' {
+		proxy = proxy[:len(proxy)-1]
+	}
+
+	proxyURL, err := url.Parse(proxy)
+	if err != nil {
+		return "", fmt.Errorf("Failed to parse %s : %v", proxy, err)
+	}
+	return proxyURL.String(), nil
+}
+
 func copyFiles(ctx context.Context, mntDir string, agentCertPath string, caCertPath string) error {
 	p := path.Join(mntDir, "/var/lib/ciao")
 	err := SudoMakeDirectory(ctx, p)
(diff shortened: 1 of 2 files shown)


ciao-deploy/deploy/cnci.go
```diff
@@ -183,6 +245,62 @@ func prepareImage(ctx context.Context, baseImage string, agentCertPath string, c
 	return preparedImagePath, nil
 }
 
+func getCNCIURL(ctx context.Context) (string, error) {
```
markdryan: The code looks fine but I'm a feel a bit uneasy about parsing HTML to find the correct file name.  Might it be easier to retrieve the latest build from the latest file as before, and use this number to retrieve the cloud image.  If the image does not exist, subtract 10 from the latest build number and try again.  We could try a maximum of 10 times before erroring out.  I realise this isn't great either as we're building some knowledge of the clear linux versioning into our tool and we're still vulnerable to layout changes in the clear linux web site.  Just wanted to suggest it though.

rbradford: I agree with @markdryan I think we should try and engage with the Clear Linux team to try and get a less fragile solution to this.

kaccardi: @markdryan I don't understand what problem your solution is trying to solve. Is it that you don't like the idea of assuming that the file will exist? or that we know the format of the file? I'd like some specifics about what is considered fragile about /images before I engage with the clear team.

markdryan: I suppose what's making me feel a little bit uneasy is that we're not calling some well defined API to determine the URL of the latest clearlinux cloud image.  We're parsing a HTML page.  The alternative I suggest isn't really much better.  So I was wondering whether we're missing something here.  Do the clearlinux team have a recommended way of programmatically retrieving the latest cloud image?

markdryan: My other concern is that we don't really understand why umount works now.  I'm worried that there's an underlying race condition which hasn't really gone away.  It would be good to try to figure out what is really causing the failure, when it happens.

kaccardi: I don't have definitive proof, but from talking with swupd developer, there are post install scripts that are forked off and the program doesn't wait for them to exit. I do have proof the unmount fails because of device busy. I suspect these post install scripts haven't quite completed. As we've observed, adding any amount of delay seems to make the problem vastly more difficult to reproduce.

kaccardi: @markdryan I modified the PR to include switches to swupd to disable any post install scripts. This causes our umount problems to go away. I think this both confirms the root cause and resolves it :).

kaccardi: @markdryan @rbradford right now there is no other way to get the last known good cloud image than the one I have provided you with. I feel this is better than leaving our image not updated.

tmarcu: @markdryan with the state of things as of today, there is not a simple way to deduce the image version without just doing a regex match on the image dir. The latest available "update" https://download.clearlinux.org/update/ may be newer than the latest "good" image, because while base images could fail creation, updates can succeed and are the key part that we need to create. This is not limited to just the cloud image, but could happen for any image type.
Right now they are in sync, but it won't always be the case, as we don't have to release images daily like we release updates - they are sort-of a bonus/secondary artifact. 
The URL however is well defined and not subject to change from what I know, so we can always query and parse that page for the latest <imagetype>-*.img needed. It will always contain images that have passed verification and can update to the latest released update as well.

ciao-deploy/deploy/utils.go
```diff
@@ -47,6 +47,32 @@ func DefaultImageCacheDir() string {
...
+	for _, k := range keep {
+		keepMap[k] = true
+	}
+
+	for _, m := range matches {
+		if keepMap[m] {
+			continue
+		}
+
+		os.Remove(m)
```
markdryan: We should discard the error here.  Right now ciao-deploy is errcheck clean but this will introduce a warning from this tool.

ciao-deploy/deploy/cnci.go
markdryan: @kaccardi,@tmarcu okay for me if there is no better way.  I just have one very minor comment and then I'm happy to merge.

ciao-deploy/deploy/utils.go
kaccardi: thanks - fixed.

PR #1594, opened 2017-11-21
Remove instance volumes

rbradford:
Predating the ability for users to build their own workloads ciao supports the ability to associate some volumes with a specific instance at launch time. This feature adds a lot of code to the CLI and controller and is not tested via BAT.

In the interest of not having multiple methods to do the same thing and reduce our code complexity i'm proposing to remove this functionality.

It's also important to note that is code tried to implement the openstack API however there is a functional mismatch between ciao and openstack resulting in greater complexity.


rbradford:
@kaccardi @markdryan  what do you think?

PR #1595, opened 2017-11-22
Use image names in workload definition

rbradford:
This PR enables the use of image names (and enforces some rules on them) in workload definitions as part of the storage sources.

The name is resolved at workload creation time to an ID and that is what is preserved in the database. The PR also makes some cleanups related to the code adjacent to this change.

### DIFF
--- a/ciao-cli/image.go
+++ b/ciao-cli/image.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"regexp"
 	"text/template"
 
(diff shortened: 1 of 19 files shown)


ciao-cli/image.go
```diff
@@ -75,6 +76,11 @@ func (cmd *imageAddCommand) run(args []string) error {
 		return errors.New("Missing required -name parameter")
 	}
 
+	r := regexp.MustCompile("^[a-z0-9-.]{1,64}?$")
```
markdryan: I don't think the ? is needed.

PR #1596, opened 2017-11-23
Add support for privileged containers

rbradford:
Add support for privileged containers. This necessitated adding a mechanism to give tenants permissions so that is also implemented in this PR.

### DIFF
--- a/ciao-controller/internal/datastore/sqlite3db_test.go
+++ b/ciao-controller/internal/datastore/sqlite3db_test.go
@@ -1207,9 +1207,9 @@ func TestSQLiteDBUpdateTenant(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// change the name and subnet bits
 	tenant.Name = "name2"
 	tenant.SubnetBits = 20
+	tenant.Permissions.PrivilegedContainers = true
 
 	err = db.updateTenant(&tenant.Tenant)
 	if err != nil {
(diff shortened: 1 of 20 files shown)


payloads/start.go
```diff
@@ -201,6 +201,10 @@ type WorkloadRequirements struct {
 
 	// NetworkNode specifies that this workload must be scheduled on a network node
 	NetworkNode bool `yaml:"network_node,omitempty"`
+
+	// Privileged indicates that this container workload should be run with increased
+	// permissionsl
```
rbradford: nit: typo. Looks like I mashed "l" rather than "."

ciao-controller/internal/datastore/sqlite3db_test.go
```diff
@@ -1221,13 +1221,61 @@ func TestSQLiteDBUpdateTenant(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(ts) != 1 {
+		t.Fatalf("Unexpected number of tenants")
+	}
+
+	if !ts[0].Permissions.PrivilegedContainers {
+		t.Fatal("Expected tenant permission set correctly")
```
markdryan: Does it matter if we don't delete the tenant or disconnect from the db on failure?

ciao-launcher/docker_test.go
```diff
@@ -662,6 +662,78 @@ func TestDockerCreateImageWithResources(t *testing.T) {
...
+			VnicIP:     testutil.AgentIP,
+			Privileged: true,
+		}}
+
+	if err := d.createImage("bridge", "172.16.0.1", nil, nil); err != nil {
+		t.Fatalf("Unable to create image : %v", err)
+	}
+
+	if tc.hostConfig.Privileged != true {
+		t.Fatal("Expected privileged host config")
```
markdryan: Not sure that it really matters, but these should probably be t.Error so deleteImage gets called.  deleteImage doesn't really do anything but still.

ciao-launcher/docker_test.go
```diff
@@ -662,6 +662,78 @@ func TestDockerCreateImageWithResources(t *testing.T) {
...
+			break
+		}
+	}
+
+	if !found {
+		t.Fatal("Require security option not found")
+	}
+
+	count := 0
+	for _, b := range tc.hostConfig.Binds {
```
markdryan: I'm not sure this tests the correct thing.  The test would pass if tc.hostConfig.Bind = []string{"/dev:/dev","/dev:/dev"}.

ciao-controller/internal/datastore/sqlite3db_test.go
rbradford: No not really. I spotted this too. Each test is run in it's own temporary in-memory DB, everything is thrown out on test exit.

ciao-launcher/docker_test.go
rbradford: Fixed.

ciao-launcher/docker_test.go
rbradford: Fixed.

markdryan:
We could also do with a few words about this on the web site, maybe in Working with Workloads:

https://ciao-project.github.io/workloads.html

PR #1597, opened 2017-11-24
ciao-controller: Remove default CNCI password

rbradford:
If the user didn't specify a CNCI user password we would default to the well
known "ciao". It is not required, nor expected, for an admin to need to
be able to login (and thus specify a password.)

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -2503,9 +2503,17 @@ func (ds *Datastore) GenerateCNCIWorkload(vcpus int, memMB int, diskMB int, key 
 #cloud-config
 users:
   - name: cloud-admin
-    gecos: CIAO Cloud Admin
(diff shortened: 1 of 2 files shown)


rbradford:
@markdryan Shall we fix this little security problem?

markdryan:
Maybe we should also set lock_password to true and remove the password from here?

https://github.com/ciao-project/ciao/blob/master/ciao-controller/internal/datastore/datastore.go#L2507

rbradford:
@markdryan like this?

markdryan:
I'm not sure what the default is, i.e., whether lock_password defaults to true if it's not specified.  I'm not sure it does.  As this code is only used for the CNCI should we just remove the password stuff altogether?  Or are we leaving the code in place for debug purposes?

rbradford:
@markdryan better plan. See #1598 for full removal.

PR #1598, opened 2017-11-24
Remove CNCI admin password

rbradford:
Remove admin password for CNCI. The admin can provide an SSH key and get passwordless sudo anyway. This fixes a security issue where not specifying a password would set it to "ciao" and allow access.

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -2497,15 +2497,14 @@ func (ds *Datastore) UnMapExternalIP(address string) error {
 
 // GenerateCNCIWorkload is used to create a workload definition for the CNCI.
 // This function should be called prior to any workload launch.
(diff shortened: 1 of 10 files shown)


ciao-controller/internal/datastore/datastore.go
```diff
@@ -2497,15 +2497,13 @@ func (ds *Datastore) UnMapExternalIP(address string) error {
...
 // This function should be called prior to any workload launch.
-func (ds *Datastore) GenerateCNCIWorkload(vcpus int, memMB int, diskMB int, key string, password string) {
+func (ds *Datastore) GenerateCNCIWorkload(vcpus int, memMB int, diskMB int, key string) {
 	// generate the CNCI workload.
 	config := `---
 #cloud-config
 users:
   - name: cloud-admin
     gecos: CIAO Cloud Admin
-    lock-passwd: false
```
markdryan: I think we should set this explicitly to true otherwise ucd wont lock the password

rbradford: Okay. Fixed!

PR #1601, opened 2017-11-27
workload_bat: Add container workload tests

rbradford:
Add a test to create a workload with a docker container using the
default registry and another one with the registry explicitly
specified.


_release/bat/workload_bat/workload_bat_test.go
```diff
@@ -506,7 +506,7 @@ func TestPriviligedWorkload(t *testing.T) {
 		Privileged: true,
 	}
 
-	testSchedulableContainerWorkload(ctx, t, requirements, false)
+	testSchedulableContainerWorkload(ctx, t, "docker:latest", requirements, false)
```
markdryan: Should this be debian:latest?

rbradford: Oops! Fixed.

PR #1603, opened 2017-11-28
ciao-controller: remove duplicate volume creation paths

rbradford:
The paths to create volumes for an instance and to create volumes
explicitly duplicated lot of code which this commit consolidates.

### DIFF
--- a/ciao-controller/api/api.go
+++ b/ciao-controller/api/api.go
@@ -95,6 +95,7 @@ type RequestedVolume struct {
 	Description string `json:"description,omitempty"`
 	Name        string `json:"name,omitempty"`
 	ImageRef    string `json:"imageRef,omitempty"`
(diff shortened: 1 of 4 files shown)


rbradford:
@markdryan You're right - it is necessary to handle the quotas only on the !internal case.

ciao-controller/api/api.go
```diff
@@ -95,6 +95,7 @@ type RequestedVolume struct {
 	Description string `json:"description,omitempty"`
 	Name        string `json:"name,omitempty"`
 	ImageRef    string `json:"imageRef,omitempty"`
+	Internal    bool   `json:"-"`
```
markdryan: Does this allow clients to create internal volumes via the API?  Is this something that they can currently do?

PR #1604, opened 2017-11-28
ciao-controller: support same image name for multiple tenants

rbradford:
The original code for implementing image name resolution inadvertently
enforced a restriction of image name uniqueness across the whole cluster
rather than per-tenant. Adopt the same algorithm that instance name
resolution uses so that the uniqueness is only per tenant.

### DIFF
--- a/ciao-controller/internal/datastore/datastore.go
+++ b/ciao-controller/internal/datastore/datastore.go
@@ -192,7 +192,6 @@ type Datastore struct {
 
 	imageLock      *sync.RWMutex
 	images         map[string]types.Image
-	imagesByName   map[string]string
 	publicImages   []string
 	internalImages []string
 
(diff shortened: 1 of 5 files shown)


ciao-controller/internal/datastore/datastore.go
```diff
@@ -2655,21 +2649,39 @@ func (ds *Datastore) GetImage(ID string) (types.Image, error) {
...
-		name = id
+	if tenantID != "" && tenantID != "admin" {
+		t, ok := ds.tenants[tenantID]
+		if !ok {
+			return "", fmt.Errorf("Tenant not found: %s", tenantID)
+		}
+
+		for _, id := range t.images {
+			i := ds.images[id]
+			if i.Name == name || i.ID == name {
```
markdryan: I'm wondering about this check.  If you had two different images, one whose id was equal to the name of the second, you would not be able to resolve the second image by name.  Corner case I know, but I thought I'd mention it.

markdryan: Here's the example, I'm thinking about.

Image 1 is created

Id ID1
Name: ID2

Image 2 is created

Id ID2
Name:  hello

This will be allowed by AddImage as image1.ID != image2.ID and image1.Name != image2.Name.  However, it will not be possible to resolve Image2 by its ID.

However, as discussed, since it's no longer possible to specify image IDs when creating an image it's almost impossible that this could happen in practice.

Would it make sense to impose an extra Constraint in AddImage that said that a name cannot be set to an existing id and vice-versa?

rbradford: I think this should be addressed now (and fix problem with updating image and name clash there.)

PR #1608, opened 2017-11-28
send cnci refresh command

kaccardi:
This PR implements the remainder of the support for per subnet CNCIs. Each tenant will now have a set of CNCIs, one per subnet. When a new CNCI is added or removed, as indicated by a ConcentratorInstanceAdded event, the controller will send each CNCI a list of the current CNCIs for that tenant. Each CNCI will have an IP allocated to it which will become a GRE tunnel endpoint. Each CNCI will create a point to multipoint GRE tunnel to direct traffic to the appropriate CNCI as necessary.

### DIFF
--- a/ciao-controller/main.go
+++ b/ciao-controller/main.go
@@ -22,6 +22,7 @@ import (
 	"crypto/x509"
 	"flag"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -56,6 +57,23 @@ type controller struct {
 	httpServers         []*http.Server
 }
 
+type cnciNetFlag string
+
+func (c *cnciNetFlag) String() string {
+	return string(*c)
+}
+
+func (c *cnciNetFlag) Set(val string) error {
+	IP := net.ParseIP(val)
+	if IP == nil {
+		return fmt.Errorf("Unable to parse CNCI network address")
+	}
+
+	*c = cnciNetFlag(IP.String())
+
+	return nil
+}
+
 var cert = flag.String("cert", "", "Client certificate")
 var caCert = flag.String("cacert", "", "CA certificate")
 var serverURL = flag.String("url", "", "Server URL")
@@ -72,6 +90,10 @@ var clientCertCAPath = "/etc/pki/ciao/auth-CA.pem"
 var cephID = flag.String("ceph_id", "", "ceph client id")
 
 var adminSSHKey = ""
+
+// this default allows us to have up to 32K hosts within the upper part
+// of the 192.168.0.0/16 private address space.
+var cnciNet cnciNetFlag = "192.168.128.0"
 
 func init() {
 	flag.Parse()
@@ -193,6 +215,14 @@ func main() {
 
 	if clusterConfig.Configure.Controller.ClientAuthCACertPath != "" {
 		clientCertCAPath = clusterConfig.Configure.Controller.ClientAuthCACertPath
+	}
+
+	if clusterConfig.Configure.Controller.CNCINet != "" {
+		err = cnciNet.Set(clusterConfig.Configure.Controller.CNCINet)
+		if err != nil {
+			glog.Fatalf("Invalid CNCI Net cluster configuration: %v", err)
+			return
+		}
 	}
 
 	ctl.ds.GenerateCNCIWorkload(cnciVCPUs, cnciMem, cnciDisk, adminSSHKey)
--- a/networking/libsnnet/cnci.go
+++ b/networking/libsnnet/cnci.go
@@ -377,7 +377,7 @@ func createCnciBridge(bridge *Bridge, brInfo *bridgeInfo, tenant string, subnet 
 	return err
 }
 
-func createCnciTunnel(gre *GreTunEP) (err error) {
+func createCnciTunnel(gre *GreTapEP) (err error) {
 	if err = gre.create(); err != nil {
 		return err
 	}
@@ -405,7 +405,7 @@ func checkInputParams(subnet net.IPNet, subnetKey int, cnIP net.IP) error {
 //If the function returns error the bridgeName can be ignored
 //If the function does not return error and has a valid bridge name
 //then the subnet has been found and no further processing is needed
-func (cnci *Cnci) addSubnetToTopology(bridge *Bridge, gre *GreTunEP, brInfo **bridgeInfo) (brExists bool,
+func (cnci *Cnci) addSubnetToTopology(bridge *Bridge, gre *GreTapEP, brInfo **bridgeInfo) (brExists bool,
 	greExists bool, bLink *linkInfo, gLink *linkInfo, err error) {
 	err = nil
 
@@ -477,7 +477,7 @@ func (cnci *Cnci) AddRemoteSubnet(subnet net.IPNet, subnetKey int, cnIP net.IP) 
 		return "", err
 	}
 
-	gre, err := newGreTunEP(genGreAlias(subnet, cnIP), cnci.ComputeAddr[0].IPNet.IP, cnIP, uint32(subnetKey))
+	gre, err := newGreTapEP(genGreAlias(subnet, cnIP), cnci.ComputeAddr[0].IPNet.IP, cnIP, uint32(subnetKey))
 	if err != nil {
 		return "", err
 	}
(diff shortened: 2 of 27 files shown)


markdryan:
@kaccardi Will this fix #1618 ?

kaccardi:
@markdryan yes

networking/libsnnet/network.go
```diff
@@ -34,7 +34,7 @@ func netError(dev interface{}, format string, args ...interface{}) error {
 		return fmt.Errorf("vnic error: "+format, args...)
 	case CnciVnic, *CnciVnic:
 		return fmt.Errorf("cncivnic error: "+format, args...)
-	case GreTunEP, *GreTunEP:
+	case GreTapEP, *GreTapEP:
```
mcastelino: Thanks for fixing this. This was an error in the original implementation.

networking/libsnnet/network.go
```diff
@@ -189,3 +189,14 @@ type GreTunEP struct {
 	CNCIId   string // UUID of the CNCI
 	CNId     string // UUID of the CN
 }
+
+// GreTunEP ciao GRE Tap device representation
```
mcastelino: Just a nit about the comment. 
This is actually a GRE point to multi-point tunnel. Not a simple GRE Tunnel.

networking/libsnnet/gretun.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if err := netlink.LinkSetAlias(g.Link, alias); err != nil {
+		return netError(g, "set alias link set alias %v %v", alias, err)
+	}
+
+	return nil
+}
+
+// Attach the GRE tunnel to a device/bridge/switch
+func (g *GreTunEP) attach(dev interface{}) error {
```
mcastelino: Do we need these. Also I am not sure if attaching to a bridge will even work for a tun interface.

networking/libsnnet/cnci.go
```diff
@@ -462,6 +471,174 @@ func (cnci *Cnci) addSubnetToTopology(bridge *Bridge, gre *GreTunEP, brInfo **br
...
+
+			var added bool
+			for _, a := range addrs {
+				if addr.Equal(a) {
+					added = true
+				} else {
+					// we should remove this addr.
+					// we can do this because there should
+					// be only one.
+					err = netlink.AddrDel(tun.Link, &a)
```
mcastelino: I like this. This will help support reconfiguration

networking/libsnnet/cnci.go
```diff
@@ -462,6 +471,174 @@ func (cnci *Cnci) addSubnetToTopology(bridge *Bridge, gre *GreTunEP, brInfo **br
...
+			dst := net.IPNet{
+				IP:   net.ParseIP(n.TunnelIP),
+				Mask: net.CIDRMask(32, 32),
+			}
+
+			route := netlink.Route{
+				LinkIndex: tun.Link.Index,
+				Dst:       &dst,
+			}
+			err = netlink.RouteAdd(&route)
```
mcastelino: We check for neigh conflicts but not route conflict? So you are using neigh as an indicator for existence of the CNCI to CNCI relationship?

networking/libsnnet/cnci.go
```diff
@@ -462,6 +471,174 @@ func (cnci *Cnci) addSubnetToTopology(bridge *Bridge, gre *GreTunEP, brInfo **br
...
+			LLIPAddr:  net.ParseIP(n.PhysicalIP),
+			LinkIndex: tun.Link.Index,
+			State:     netlink.NUD_PERMANENT,
+		}
+		updated = append(updated, neigh)
+
+		var exists bool
+		// see if this already exists
+		for _, neighbor := range neighs {
+			if strings.Compare(neighbor.String(), neigh.String()) == 0 {
```
mcastelino: Is this is the right comparison. We are comparing full equality.

However what we should be checking is if the IP is reachable via a different LLIPAddr.

So I think we should be checking if IP matches. We should not have the same IP reachable via two different LLAddrs.

So if IP matchs and we have a full match the config has not changed. And we skip ahead.

However if IP matches but LLIPAddr does not match, we should delete the old entry and add a new one.

networking/libsnnet/cnci.go
mcastelino: This may actually be ok as the tunnel IP should never change even if the CNCI moves around.

kaccardi: @mcastelino yes - since for each neigh we add routes it was sufficient to check neigh IMO.

networking/libsnnet/cnci.go
kaccardi: ok - I'll fix this.

networking/libsnnet/gretun.go
kaccardi: probably not - i'll clean that up.

networking/libsnnet/network.go
kaccardi: thanks, I will make a more accurate comment.

ciao-controller/main.go
```diff
@@ -73,6 +91,8 @@ var cephID = flag.String("ceph_id", "", "ceph client id")
 
 var adminSSHKey = ""
 
+var cnciNet cnciNetFlag = "192.168.0.0"
```
rbradford: Should this be specified in CIDR notation? Also is the body of the commit message correct? It says it should be a /18 should it be /16?

testutil/payloads.go
```diff
@@ -230,6 +232,23 @@ const RestoreYaml = `restore:
   workload_agent_uuid: ` + AgentUUID + `
 `
 
+// CNCITunnelID is a gre tunnel ID derived from the tenant UUID
+var CNCITunnelID = crc32.ChecksumIEEE([]byte(TenantUUID))
```
rbradford: What happens if there is a clash?

ciao-controller/main.go
kaccardi: @rbradford no - we don't need a cidr notation as we require that the network have enough hosts to cover the maximum needed number of cncis - so it just doesn't really need cidr notation. /18 is the max number of cncis we'll ever need.

testutil/payloads.go
kaccardi: The probability of a collision is small for the likely number of tenants. On top of that, I think you'd have to have the same tunnelID for the same tunnel source and destination for it to be an issue. We should check with manohar to see if this is something we really need to be concerned about.

ciao-controller/main.go
rbradford: As this IP address range is exposed to the user, i.e. from instances these IPs are reachable I think we should choose a range that is much less widely used than 192.168.0.0/18 as this would prevent instances from accessing resources in the external network if that range is in use.

kaccardi: it's just a default - it's configurable by the user via the config file.

testutil/payloads.go
kaccardi: @mcastelino can you check my reasoning here?

ciao-controller/main.go
kaccardi: I will have to make sure that it's documented that they can't chose 172.16-31 and that this is the default.

testutil/payloads.go
mcastelino: @rbradford great question. Luckily for us, here the tunnel traffic is only examined once it lands inside the CNCI. So even if we end up with a collusion it will not be an issue as it will not conflict within the CNCI. Nothing on the wire looks at the GRE tunnel info, so that helps us.

ciao-controller/main.go
kaccardi: @rbradford and btw just to correct your comment - these IP addresses are not reachable by the instances. This network is for cnci to cnci communication only - and the physical IPs of the CNCI are what would actually go out on the wire. The issue would be if you had the same network for the physical IPs as the cnci tunnel IPs.

rbradford: Before your change any packets that an instance sent to 192.168.0.0/18 would have fallen through the default route and NATing on the CNCI and hit the external network. However after your change those packets will be directed towards the CNCI.

These IP addresses are reachable from instances:
Here is the output from mtr:
```
 Host                                                         Loss%   Snt   Last   Avg  Best  Wrst StDev
 1. 172.16.0.33                                                0.0%     3    0.6   0.5   0.4   0.6   0.0
 2. 192.168.0.4                                                0.0%     2    0.8   0.8   0.8   0.8   0.0
```

My problem is that your default range encompasses the most common RFC1918 subnet 192.168.1.1/24 and its going to be a problem for users.

Please choose a default network which is much less common. Also you should modify ciao-deploy to reflect this new configuration option.

kaccardi: ok, I switched it to 192.168.128.0 and added the option to configure to ciao-deploy.

PR #1611, opened 2017-11-30
Fix kubicle

markdryan:
This PR fixes kubicle which had been broken by two separate issues:

1. Kubicle was not working in SingleVM with the recommended ccloudVM size of 6GB as the CNCI was consuming too much memory (2GBs instead of 128MB).  This issue has been fixed by adding a new option, --cnci, to ciao-deploy allowing users some control of the resource consumption of CNCIs.
2. The installation instructions for flannel have changed.  The old instructions no longer worked.  This meant that kubicle failed as the master node did not complete its setup correctly.

This commit also bumps up the default version of kubernetes installed by kubicle to 1.7.11

### DIFF
--- a/ciao-deploy/cmd/setup.go
+++ b/ciao-deploy/cmd/setup.go
@@ -25,14 +25,36 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
+type cnciFlag string
+
+func (f *cnciFlag) String() string {
+	return string(*f)
+}
+
+func (f *cnciFlag) Type() string {
+	return "string"
+}
+
+func (f *cnciFlag) Set(val string) error {
+	if val != "tiny" && val != "medium" && val != "large" {
+		return fmt.Errorf("tiny, medium or large expected")
+	}
+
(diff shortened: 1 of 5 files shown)


ciao-deploy/cmd/setup.go
```diff
@@ -124,4 +146,5 @@ func init() {
 	setupCmd.Flags().BoolVar(&force, "force", false, "Overwrite existing files which might break the cluster")
 	setupCmd.Flags().BoolVar(&localLauncher, "local-launcher", false, "Enable a local launcher on this node (for testing)")
 	setupCmd.Flags().BoolVar(&clusterConf.DisableLimits, "disable-limits", false, "Disable memory limit checking for cluster nodes")
+	setupCmd.Flags().Var(&cnciSize, "cnci", "Specifies the resources (mem, cpu) available to CNCIs.  Can be 'tiny', 'medium', 'large'")
```
rbradford: I would prefer this to be --cnci-size?

PR #1616, opened 2017-12-01
networking: Enable ssh for the last instance in a subnet

markdryan:
There was a bug in the CNCI agent which was incorrectly computing the
number of ip addresses in a subnet.  The result being that it was
impossible to ssh into the last instance in a subnet.

Fixes: https://github.com/ciao-project/ciao/issues/1615


mcastelino:
LGTM.

PR #1623, opened 2017-12-06
qemu: add function to hotplug CPUs

devimc:
This patch adds a new function to hot add CPUs in a QEMU instance

### DIFF
--- a/qemu/qmp.go
+++ b/qemu/qmp.go
@@ -107,7 +107,8 @@ type QMPEvent struct {
 }
 
 type qmpResult struct {
-	err error
+	response interface{}
+	err      error
 }
 
 type qmpCommand struct {
@@ -203,14 +204,14 @@ func (q *QMP) processQMPEvent(cmdQueue *list.List, name interface{}, data interf
 	}
 }
 
-func (q *QMP) finaliseCommand(cmdEl *list.Element, cmdQueue *list.List, succeeded bool) {
+func (q *QMP) finaliseCommandWithResponse(cmdEl *list.Element, cmdQueue *list.List, succeeded bool, response interface{}) {
 	cmd := cmdEl.Value.(*qmpCommand)
 	cmdQueue.Remove(cmdEl)
 	select {
 	case <-cmd.ctx.Done():
 	default:
 		if succeeded {
-			cmd.res <- qmpResult{}
+			cmd.res <- qmpResult{response: response}
 		} else {
 			cmd.res <- qmpResult{err: fmt.Errorf("QMP command failed")}
 		}
@@ -218,6 +219,10 @@ func (q *QMP) finaliseCommand(cmdEl *list.Element, cmdQueue *list.List, succeede
 	if cmdQueue.Len() > 0 {
 		q.writeNextQMPCommand(cmdQueue)
 	}
+}
+
+func (q *QMP) finaliseCommand(cmdEl *list.Element, cmdQueue *list.List, succeeded bool) {
+	q.finaliseCommandWithResponse(cmdEl, cmdQueue, succeeded, nil)
 }
 
 func (q *QMP) processQMPInput(line []byte, cmdQueue *list.List) {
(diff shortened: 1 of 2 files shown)


devimc:
@markdryan changes applied

qemu/qmp.go
```diff
@@ -722,3 +752,40 @@ func (q *QMP) ExecutePCIVFIODeviceAdd(ctx context.Context, devID, bdf, addr, bus
...
+}
+
+// ExecuteQueryHotpluggableCPUs returns a slice with the list of hotpluggable CPUs
+func (q *QMP) ExecuteQueryHotpluggableCPUs(ctx context.Context) ([]HotpluggableCPU, error) {
+	response, err := q.executeCommandWithResponse(ctx, "query-hotpluggable-cpus", nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	// convert response to json
```
markdryan: This is a bit weird, but I can't actually think of a better way of doing this.

However, we should wrap any returned errors.  Right now, if one of the json Marshal or UnMarshal methods failed we'd return a json error to the user, which might be a little confusing, so something like

```
if err != nil {
    return nil, fmt.Errorf("Unable to extract CPU information: %v", err)
}
```

qemu/qmp.go
```diff
@@ -722,3 +752,40 @@ func (q *QMP) ExecutePCIVFIODeviceAdd(ctx context.Context, devID, bdf, addr, bus
...
+	data, err := json.Marshal(response)
+	if err != nil {
+		return nil, err
+	}
+
+	var cpus []HotpluggableCPU
+	// conver json to []HotpluggableCPU
```
markdryan: Should be convert

qemu/qmp_test.go
```diff
@@ -810,3 +810,42 @@ func TestQMPPCIDeviceAdd(t *testing.T) {
...
+	q.Shutdown()
+	<-disconnectedCh
+}
+
+// Checks that hotpluggable CPUs are listed correctly
+func TestQMPExecuteQueryHotpluggableCPUs(t *testing.T) {
+	connectedCh := make(chan *QMPVersion)
+	disconnectedCh := make(chan struct{})
+	buf := newQMPTestCommandBuffer(t)
+	buf.AddCommand("query-hotpluggable-cpus", nil, "return", nil)
```
markdryan: This isn't actually testing the function as it doesn't pass in any response data.  We should pass valid information for the last parameter of AddCommand and check it is correctly converted into a slice of HotpluggableCPU CPUs.

devimc: this is an interesting issue, because ```AddCommand``` receives a ```map``` and what I need is a ```slice```, I'll add a new test result structure (```qmpTestResultSlice```)  and a new member of this type to ```qmpTestCommandBuffer```, then I'll implement a new version of ```AddCommand```, otherwise I'd have to modify ```qmpTestResult``` to support slices and patch all the ```qmp_test.go``` file

anyway lemme check what I can do

qemu/qmp.go
devimc: fixing, thanks

qemu/qmp_test.go
markdryan: I think AddCommand should be modified to take an interface{} instead of a map.  This would make more sense.  Note this parameter has never been used before which is why it was defined incorrectly.  As none of the other tests are currently using it, changing the data parameter to interface{} won't break anything. You'd also need to change qmpTestResult.data

devimc:
@markdryan changes applied, thanks

qemu/qmp.go
```diff
@@ -751,3 +767,25 @@ func (q *QMP) ExecuteCPUDeviceAdd(ctx context.Context, driver, cpuID, socketID,
...
+	// convert response to json
+	data, err := json.Marshal(response)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to extract CPU information: %v", err)
+	}
+
+	var cpus []HotpluggableCPU
+	// convert json to []HotpluggableCPU
+	if err = json.Unmarshal(data, &cpus); err != nil {
+		return nil, fmt.Errorf("Unable to conver json to hotpluggable CPU: %v", err)
```
markdryan: "Unable to cover"

devimc: fixed, thanks

PR #1625, opened 2017-12-07
Remove node status API and command

rbradford:
Remove legacy API for node summary and "ciao-cli node status" command that uses it along with BAT testing functionality relating to it.


rbradford:
@markdryan would be great if you could review & merge this as soon as you can so I can rebase new CLI branch and take in this.

PR #1628, opened 2017-12-11
qemu: Add maxcpus attribute to -smp

devimc:
maxcpus is used to specify how many cpus a VM can have.
This attribute must be specified to enable the hotplugging CPUs capability,
otherwise the maximum number of CPU will be defined by the number of CPU
in -smp.

### DIFF
--- a/qemu/qemu.go
+++ b/qemu/qemu.go
@@ -980,6 +980,10 @@ type SMP struct {
 
 	// Sockets is the number of sockets made available to qemu.
 	Sockets uint32
+
+	// MaxCPUs is the maximum number of VCPUs that a VM can have.
+	// This value MUST BE equal to or greater than CPUs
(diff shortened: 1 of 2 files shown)


devimc:
@mcastelino from performance and memory footprint perspective I don't see any issue, but I still have some questions.

why only 4 sockets?

is going to be 1:1 the relation between CPUs and sockets (1 CPU 1 socket) ?
or N:4 (N CPUs 4 sockets) ?

the maximum number of CPUs per VM recommended by KVM is 240, have you tried that case with both scenarios (1:1  and  N:4) ?

why 4 sockets and not 2?

either way maxcpus must be specified, the only thing that change is the relation between CPUs, cores, threads and sockets.

devimc:
moving this PR to https://github.com/intel/govmm/pull/3

PR #1633, opened 2018-01-02
ciao: Add templating usage to command

rbradford:
Put template usage information in the usage output of commands in ciao that use tfortools for output.

### DIFF
--- a/ciao/cmd/root.go
+++ b/ciao/cmd/root.go
@@ -28,6 +28,7 @@ import (
 var c client.Client
 
 var template string
+var rootUsageFunc (func(cmd *cobra.Command) error)
 
 func render(cmd *cobra.Command, data interface{}) error {
 	if template == "" && cmd.Annotations != nil {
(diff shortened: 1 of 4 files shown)


ciao/cmd/root.go
```diff
@@ -42,6 +43,25 @@ func render(cmd *cobra.Command, data interface{}) error {
...
+		return err
+	}
+
+	if cmd.Annotations == nil {
+		return nil
+	}
+
+	templateUsage := cmd.Annotations["template_usage"]
+	if templateUsage != "" {
+		fmt.Printf("When using the template flag the following structure is provided:\n\n")
```
markdryan: If might be nice to prepend a '\n' here.  This is what I currently see.

```
$ ciao list cncis
Error: Listing CNCIs is limited to privileged users
When using the template flag the following structure is provided:

[]struct {
```
and I'm thinking

```
$ ciao list cncis
Error: Listing CNCIs is limited to privileged users

When using the template flag the following structure is provided:

[]struct {
```

would be a bit nicer.

markdryan: done

